Issue #3, opened 2025-07-17
Cannot assign to non-existent property "inlineReplySupported"


Amadoabad:
I get this error after pulling the last commit 
`
 ERROR: Failed to load configuration
 ERROR:   caused by @shell.qml[174:19]: Type TopBar unavailable
 ERROR:   caused by @Widgets/TopBar/TopBar.qml[156:17]: Type LauncherButton unavailable
 ERROR:   caused by @Widgets/TopBar/LauncherButton.qml[-1:-1]: Type AppSearchService unavailable
 ERROR:   caused by @Services/AppSearchService.qml[-1:-1]: Type AudioService unavailable
 ERROR:   caused by @Services/AudioService.qml[-1:-1]: Type BatteryService unavailable
 ERROR:   caused by @Services/BatteryService.qml[-1:-1]: Type BluetoothService unavailable
 ERROR:   caused by @Services/BluetoothService.qml[-1:-1]: Type BrightnessService unavailable
 ERROR:   caused by @Services/BrightnessService.qml[-1:-1]: Type CalendarService unavailable
 ERROR:   caused by @Services/CalendarService.qml[-1:-1]: Type FocusedWindowService unavailable
 ERROR:   caused by @Services/FocusedWindowService.qml[-1:-1]: Type LauncherService unavailable
 ERROR:   caused by @Services/LauncherService.qml[-1:-1]: Type MprisController unavailable
 ERROR:   caused by @Services/MprisController.qml[-1:-1]: Type NetworkService unavailable
 ERROR:   caused by @Services/NetworkService.qml[-1:-1]: Type NiriWorkspaceService unavailable
 ERROR:   caused by @Services/NiriWorkspaceService.qml[-1:-1]: Type NotificationService unavailable
 ERROR:   caused by @Services/NotificationService.qml[29:9]: Cannot assign to non-existent property "inlineReplySupported"
`

purian23:
Hi there, this project is not yet released and is still in active development, we expect a lot of potential breaking changes until v1.0 is released. [Inline-replies](https://github.com/quickshell-mirror/quickshell/commit/c40074dd5684df0efa1eda7aa0820d7be2a3e43e) were just merged to the master branch of quickshell a couple of days back.  
Thanks for the request.

bbedward:
But this is because of a newer quickshell change, if you re-build from git this will work.

PR #4, opened 2025-07-17
fix: Notification Service - non-existent property related to issue #3

Amadoabad:
this fixes the issue #3.


purian23:
Closing from remarks left on the previous reply. 
https://github.com/bbedward/DankMaterialShell/issues/3#issuecomment-3092340784

PR #5, opened 2025-07-19
Added number indexing option for the workspace switcher

Amadoabad:


purian23:
Thanks, this looks good!

PR #37, opened 2025-08-11
Remove dependency and import of Niri HM module

bluelinden:
DankMaterialShell's flake failed to build on my system because it imported the Niri Home-Manager module a second time, after my config. 

This PR refactors the flake to no longer require the Niri flake to be imported directly. The Niri flake is now only needed if the user wants to use an option that requires it, such as enableKeybinds or enableSpawn. This also reduces the download requirements of the flake.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,4 +1,3 @@
-
 {
   description = "Dank material shell.";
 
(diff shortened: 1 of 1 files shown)


bbedward:
Could you also make a flake for dgop? https://github.com/AvengeMedia/dgop

People will see it as just another monitor, but it’s not because of how smart it is by being stateless. It’s removed sampling delays from every desktop widget that wants a cpu percentage or net rate or disk rate.

since it’s now required for all the shells monitoring widgets it kinda makes sense.

single binary there so should be easy.

bluelinden:
hey! it may be a little while as i just accidentally formatted my laptop’s hard drive and it’ll take a few days for me to restore from backup.- blue lindenOn Aug 10, 2025, at 23:18, BB ***@***.***> wrote:﻿
Merged #37 into master.

PR #53, opened 2025-08-15
Made a Pomodoro Timer

jiboli:
On the barebones for now, works and have the core functionality.

Adds a pomodoro timer to the topBar. The times show when it is running, hides itself when idle.
Popout have 2 tabs, one for control and another for settings.

Much ugly, but functional

Hope that i managed to fix all merge conflicts.


Widgets/DankCircularProgress.qml
```diff
@@ -0,0 +1,60 @@
+import QtQuick 2.15
```
bbedward: Remove the version here

shell.qml
```diff
@@ -216,53 +217,170 @@ ShellRoot {
...
+
+        ProcessListModal {
+            id: processListModal
+        }
+    }
+
+    PomodoroPopout {
+        id: pomodoroPopout
+    }
+
```
bbedward, line 174: Pretty much no changes to shell.qml should be here besides the addition of Pomodoro stuff

Services/PomodoroService.qml
```diff
@@ -0,0 +1,126 @@
+pragma Singleton
+
+import QtQuick
+import qs.Common
+
+Item {
```
bbedward: Add `pragma ComponentBehavior: Bound` and change to `Singleton {`

Common/PomodoroSettings.qml
```diff
@@ -0,0 +1,91 @@
+pragma Singleton
+
```
bbedward, line 2: This should be part of existing settings data or session data

.gitignore
```diff
@@ -97,3 +97,6 @@ go.work.sum
 # Editor/IDE
 # .idea/
 # .vscode/
+Common/qmldir
+Services/qmldir
+Widgets/qmldir
```
bbedward, line 102: gitignore additions not needed

bbedward (changes requested):

shell.qml
jiboli: sory, made some mistakes merging my branch

PR #76, opened 2025-08-20
Fix Wallpaper preview

xdenotte:
Fixed wallpaper preview in settings and filebrowsermodal, by changing imagepath to source


bbedward:
I'm having copilot review just because i wanted to see what happened lol, didn't know it was an option

PR #96, opened 2025-08-24
Fix Wifi state icon but not a signal

xdenotte:
This fixes the wifi icon with a cross for me, but there is a problem that the wifi signal always returns to network_wifi_1_bar, no matter how much I tried, I could not do anything.


xdenotte:
I'd rather create a issue

PR #103, opened 2025-08-24
added a network speed monitor widget

asaadmohammed74:


purian23:
Nice work @asaadmohammed74, merged!

PR #116, opened 2025-08-25
fix: ensure matugen script exits 0 on success

xdenotte:
on nixos this resulted in constant "Failed to generate system themes" errors after each wallpaper change


purian23:
Thanks, merged!

PR #117, opened 2025-08-26
Update time format to HH:MM

Vantesh:
Currently, times are displayed without leading zeros, e.g. 3:00.

This PR updates the formatting so that single-digit hours are padded, resulting in 03:00.


bbedward:
I don't think that's desirable for 12 hour time, where you would probably not want that - 24 hour time of course makes sense

Vantesh:
> I don't think that's desirable for 12 hour time, where you would probably not want that - 24 hour time of course makes sense

Sorry about that, its fixed now.

PR #119, opened 2025-08-26
Add option to only show apps in the current workspace in the running apps widget

gonengazit:
basically on each monitor show the running apps on its active workspace - instead of showing all apps on all monitors (added as a toggleable setting)
niri only currently (because that's what I use) - but should be simple enough to add support for others - just need to implement the filterCurrentWorkspace function


bbedward:
I like it, good job. I'll add hyprland to it and make some tweaks after meging.

PR #121, opened 2025-08-26
Add UWSM session check to prefer uwsm stop over compositor exit

Vantesh:
Implements a check for UWSM and uses `uwsm stop` instead of compositor exit, as recommended by the [Arch Wiki](https://wiki.archlinux.org/title/Hyprland#Universal_Wayland_Session_Manager
)

### DIFF
--- a/Services/CompositorService.qml
+++ b/Services/CompositorService.qml
@@ -13,6 +13,7 @@ Singleton {
     // Compositor detection
     property bool isHyprland: false
     property bool isNiri: false
+    property bool uwsmActive: false
     property string compositor: "unknown"
 
     readonly property string hyprlandSignature: Quickshell.env("HYPRLAND_INSTANCE_SIGNATURE")
@@ -80,6 +81,7 @@ Singleton {
 
     Component.onCompleted: {
         detectCompositor()
+        uwsmCheck.running = true
     }
 
     function filterCurrentWorkspace(toplevels, screen){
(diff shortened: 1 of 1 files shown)


Services/CompositorService.qml
```diff
@@ -187,10 +187,33 @@ Singleton {
     }
 
     function logout() {
-        if (isNiri) {
-            NiriService.quit()
-            return
+    // Trigger the check process
+    uwsmCheck.running = true
```
bbedward: this would make sense to  do in `Component.onCompleted`

kinda like

```js
property bool uwsmActive: false;

Component.onCompleted {
  ...
  uwsmCheck.running = true
  ...
}
```

Services/CompositorService.qml
```diff
@@ -187,10 +187,33 @@ Singleton {
...
+    }
+
+    Process {
+        id: uwsmCheck
+        // `uwsm check is-active` returns 0 if in uwsm-managed session, 1 otherwise
+        command: ["uwsm", "check", "is-active"]
+        running: false
+        onExited: function(exitCode) {
+            if (exitCode === 0) {
```
bbedward: Similar here, then

`uwsmActive = exitCode === 0`

Services/CompositorService.qml
```diff
@@ -187,10 +187,33 @@ Singleton {
...
+            if (isNiri) {
+                NiriService.quit()
+                return
+            }
+            Hyprland.dispatch("exit")
         }
-        Hyprland.dispatch("exit")
+    }
+
+    Process {
```
bbedward: Then get rid of this

Services/CompositorService.qml
```diff
@@ -187,10 +187,33 @@ Singleton {
     }
 
     function logout() {
-        if (isNiri) {
```
bbedward, line 190: (last) then you could just do

```js
if (uwsmActive)
  Quickshell.execDetached(["uwsm", "stop"]);
  return;

if (isNiri)
  NiriService.quit();
  return;

// Hyprland
```

bbedward:
Tbh I think it also makes more sense in session service but that’s on me for putting it here to begin with, I’ll copy paste it later

Issue #24, opened 2025-08-07
[Feature] Gnome-like VPN management in controls panel


docryte:
## Feature Description

Well, as I mentioned in title, I want to be able at least to turn on/off VPN in control panel. It would also be great to switch between VPNs.

## Use Case

I have personal and internal business VPNs and sometimes have to switch between them

## Proposed Solution

I'm pretty new to QuickShell-stuff so I don't know. If this issue wouldn't be closed for a long time, I'll try to dive in and make it by myself, but there are no guarantees

## Alternatives/Existing Solutions

Place VPN element in separate optional element in shell instead of placing it in control panel

bbedward:
I think it’d make more sense for it to be its own module/widget and service. That can be added/removed from the top bar

sezaru:
Do you know if there is support for listing/connection/disconnecting to vpns with the QML API? I did a quick search for some documentation on how to do that but couldn't find anything.

bbedward:
There’s not, it’d have to be with CLI stuff. 

When this one is finished and merged it should allow it through network manager https://github.com/quickshell-mirror/quickshell/pull/96

For now it’d have to be cli commands (or relying on nm-applet or some systray thing)

ipg0:
What kind of VPN?

AFAIK you can add OpenVPN connection configs to NetworkManager, but probably not the newer/less popular protocols like Xray. Apart from the NM API I'm not sure how you would detect or control an arbitrary VPN clients, most of which don't even have a proper CLI.

bbedward:
The nm VPN integration was merged, but leaving it open since it'd still be nice to be able to import VPN profiles, maybe open a file chooser or something.

sezaru:
Thank you a lot for that! I just tested it and it works great!

Here are just some suggestions if you don't mind:

Maybe add the vpn widget as an option in the Control Center widget, something similar to what Gnome does.
Probably it is a good idea to add to the readme that you need `glib` (for gdbus) and `nmcli` installed for the VPN widget to work.

devnullvoid:
I initially had it as a pill in the control center widget, like in gnome, but it looked out of place due to the odd number of items.

nmcli and glib are also required for the NetworkService. I assumed everyone would have those already installed, but probably good to note in the readme anyway.

sezaru:
Now that the Control Center widget can be personalized by the user with the widgets they desire to have there, maybe we can have the VPN widget as an option there too?

devnullvoid:
[...]

PR #139, opened 2025-08-30
Feature: VPN Connection Widget for TopBar

devnullvoid:
Addresses request in #24 -- simple NetworkManager-based VPN connection manager for the TopBar


bbedward:
Does it work with v2ray ? I know network manager is its own thing just curious if it has that support out of the box

devnullvoid:
Not unless v2ray can have an entry in NetworkManager. 

Essentially, it uses nmcli to filter for connection profiles of type "vpn" or "wireguard" and allows you to toggle them on/off.

NetworkManager supported VPN connection types: https://networkmanager.dev/docs/vpn/

PR #149, opened 2025-09-02
Advanced Workspace Switcher Widget + Lockscreen Virtual Keyboard

KyleKrein:
# Advanced Workspace Switcher Widget
This new widget is turned off by default. It is just a better version of an embedded one for workspace switching. It shows all the opened apps on every workspace. For inactive workspaces it shows app icons and number on icons, if there're more than one windows of a program on the workspace. For active workspace it shows icons for each window in the correct order, highlights active window and if user clicks the icon of non active window on the current workspace, this window will be focused.

# Lockscreen Virtual Keyboard
Adds a button that launches virtual keyboard as a QML widget on the lockscreen. 
Use case: I use Niri on Framework 12 and I can't unlock my laptop when it's in tablet mode, because [it's not possible to show virtual keyboard above lockscreen](https://github.com/YaLTeR/niri/issues/2201). And this virtual keyboard is actually a part of the lockscreen itself, therefore it can be used by me and other Niri users. For Hyprland it's not necessary

P.s. This is my first time using QML, so the pull request can have problems, but I think that this contribution can be valuable for others (especially the virtual keyboard).


Modules/TopBar/AdvancedWorkspaceSwitcher.qml
```diff
@@ -0,0 +1,336 @@
...
+                var sorted = workspaces.slice().sort((a, b) => a.id - b.id)
+                root.workspaceList = SettingsData.showWorkspacePadding ? root.padWorkspaces(sorted) : sorted
+            }
+        }
+
+        target: Hyprland.workspaces
+        enabled: CompositorService.isHyprland
+    }
+
+    Connections {
```
bbedward: You can scrap this whole connections block. And every connections block tbh, I know it's not you just leftover artifacts from the workspace switcher non-advanced version, I just deleted them from that one.

Modules/TopBar/AdvancedWorkspaceSwitcher.qml
```diff
@@ -0,0 +1,336 @@
...
+            anchors.centerIn: parent
+            text: modelData.count
+            font.pixelSize: 9
+            color: "white"
+        }
+    }
+}
+                    }
+
+                    // fallback: если иконок нет — показываем текст (номер/название)
```
bbedward: Prefer english comments

Modules/TopBar/AdvancedWorkspaceSwitcher.qml
```diff
@@ -0,0 +1,336 @@
...
+        var wsCandidates = [ws.id, ws.idx].filter(x => typeof x !== "undefined");
+    
+        var wins = [];
+        if (CompositorService.isNiri) {
+            wins = NiriService.windows || [];
+        } else if (CompositorService.isHyprland) {
+            wins = Hyprland.clients ? Hyprland.clients.values : [];
+        }
+    
+        var byApp = {}; // ключ = app_id/класс, значение = агрегат
```
bbedward: english comment

Modules/TopBar/AdvancedWorkspaceSwitcher.qml
```diff
@@ -0,0 +1,336 @@
...
+                    anchors.centerIn: parent
+                    spacing: 4
+
+                    Repeater {
+                        model: root.getWorkspaceIcons(wsData)
+                        delegate: Item {
+    width: wsBox.height * 0.9
+    height: wsBox.height * 0.9
+
+    Image {
```
bbedward: Might be better to use QS' IconImage here, think it ensures icons have consistent padding which will probably make them look more consistent.

bbedward (changes requested):
Looks awesome, just a couple comments

Modules/TopBar/AdvancedWorkspaceSwitcher.qml
KyleKrein: Done

Modules/TopBar/AdvancedWorkspaceSwitcher.qml
KyleKrein: Done

Modules/TopBar/AdvancedWorkspaceSwitcher.qml
KyleKrein: Done

Modules/TopBar/AdvancedWorkspaceSwitcher.qml
KyleKrein: Done

KyleKrein:
Resolved all comments

bbedward:
I'm going to make some tweaks after merging and probably combine the functionality into the main workspace widget and just make it a toggle with other workspace settings (since I assume, nobody would want two workspace widgets so a global override is fine)

image

purian23:
Nice, thanks @KyleKrein!

PR #163, opened 2025-09-05
Add configurable per-monitor workspaces and system tray options

kylerm42:
## Background

Love the shell, but I have multiple monitors and am using [split-monitor-workspaces](https://github.com/Duckonaut/split-monitor-workspaces) on Hyprland, so I don't want *all* the workspaces to show up on every monitor. 

## Workspaces per monitor
- Add `workspacesPerMonitor` setting (defaults to true for backward compatibility)
- Update `WorkspaceSwitcher` to respect the setting for both Hyprland and Niri
- Add UI toggle in Settings > Widgets > Workspace Settings
- Allows users without split-monitor setups to show all workspaces on every monitor
- Maintains current behavior for existing users while enabling traditional desktop workflow

## System tray monitor option
- Add the system tray to `variantComponents` so it can be shown only on specific monitors (useful on vertical monitors with limited horizontal space)

## Other
- Also add a couple of apps to `moddedAppId` since the icons were missing

### DIFF
--- a/Common/Paths.qml
+++ b/Common/Paths.qml
@@ -50,6 +50,10 @@ Singleton {
     function moddedAppId(appId: string): string {
         if (appId === "Spotify")
             return "spotify-launcher"
+        if (appId === "beepertexts")
+            return "beeper"
+        if (appId === "home assistant desktop")
+            return "homeassistant-desktop"
         return appId
     }
 }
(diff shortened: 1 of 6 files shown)


purian23:
Nice, thanks @kylerm42, we appreciate the support!

PR #167, opened 2025-09-06
Add config options to features

sezaru:
Thanks for accepting my PRs so far, this is the last one.

This PR makes the flake more flexible by allowing the user to enable/disable features (in this case dependencies that will automatically disable widgets that depends on them).

I made all of them enabled by default to not make people that already use the flake had some surprise because they updated the flake and some of the features were disabled.

This also adds some dependencies that were missing in the original flake. I think I added all of them.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -154,7 +154,7 @@
               pkgs.ddcutil
               pkgs.libsForQt5.qt5ct
               pkgs.kdePackages.qt6ct
-              dms-cli.packages.${system}.default
+              dms-cli.packages.${pkgs.system}.default
             ]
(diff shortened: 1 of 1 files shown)


sezaru:
Give me a sec, there was a conflict and it broke the flake, fixing it in a minute

purian23:
No worries, we rely on experienced Nix users to help update and maintain as we do not run Nix on our end. We're always glad for the assist though and to be able to support the platform.

sezaru:
Ok, I just tested it here and seems to work fine.

I not sure what dms-cli is used for, so for now I just added to the default list of dependencies.

Issue #182, opened 2025-09-09
Fix the homemanager flake on hyprland


shalevc1098:
## Compositor

- [ ] niri
- [X] Hyprland
- [ ] Other (specify)

## Description
i tried using the homemanager flake on hyprland but i got the error below:
```sh
error: attribute 'niri' missing
       at /nix/store/3a61pjcl44m6ppmfg0xbrn9vncl5p91i-source/flake.nix:128:38:
          127|                     (lib.mkIf cfg.enableKeybinds {
          128|                         binds = with config.lib.niri.actions; let
             |                                      ^
          129|                             dms-ipc = spawn "dms" "ipc";
```

for some reason it tries accessing non existent niri under config.lib but i'm using hyprland so this is expected, but it seems like it doesn't have a proper block for non-niri compositors.

## Expected Behavior
it should build without problems

## Steps to Reproduce
use the homemanager flake on hyprland, enable and it should show this error

bbedward:
@Vortriz  would you mind helping out with this one?

Vortriz:
sure, i can give it a shot.

i will add `enableNiriIntegration` and `enableHyprlandIntegration` options. although this will introduce a sort of breaking change, since so far we have been assuming only niri integration and hence it was enabled by default. now it will have to be manually enabled.

sezaru:
What about having both options but having enableNiriIntegration as true by default, that way I guess it would not break anyone config not?

Either way, I think it is fine if there are breaking changes with the flake, as it doesn't really break the user system in a way that is irreversible (since they can just revert back to a previous generation if it breaks for them anyway).

Maybe there is a way to force the user to set one of these two options? By force I mean that the build will fail if they are not set, that way it will prevent the user from building a new generation and forcing them to update their config.

The whole point of nixos is that we can easily revert back to a functional version of the system if something breaks, so personally I would not be too concerned that the flake has breaking changes (but having some documentation about it would be nice)

sysedwinistrator:
I think the main problem here is that the niri home-manager module is not part of upstream home-manager, but is provided by a  separate repo: https://github.com/sodiboo/niri-flake.

If you're not importing the home-manager modules from niri-flake (which is very likely if you aren't using niri), just _referencing_ its config options (even via `lib.mkIf cfg.enableNiriIntegration` or similar) will cause an evaluation error, which is the issue @shalevc1098 is having.

[...]

PR #185, opened 2025-09-10
configure the default settings using flake

asaadmohammed74:


bbedward:
I want to hold off for now because I need some general collaboration and documentation around this flake. Some stuff gets merged which breaks stuff, then someone else contributes something which breaks something else, then it doesn't work with Hyprland.

I'm not a nix user so I trust others with it, but would like just one unified PR and documentation on how to install it for systemd, for non-systemd, for hyprland, for niri, with configs, without configs. 

@Vortriz  has been a solid flake contributor - but I'd just like to get one solid PR for the flake to solve all of the issues and document it. (I removed nixOS support from the dankinstall because of the flake inconsistencies, and would like to get it back #186 #182 )

asaadmohammed74:
totally understandable, since this is implemented as a separate module I will include it in my own configs and leave it here for anyone that want to do the same.

bbedward:
It's good I just want to get feedback from more nix users to iron out the flake

purian23:
@sezaru @bluelinden for visibility as well so we may be able to achieve the best solution for Nix users.

sezaru:
I'm not a seasoned NixOS user as I started using it not long ago, so take my opinion with a grain of salt.

Since none of the maintainers of this project use NixOS (you guys don't know what you are missing out :grin: ) I think we need to either have some way of validating the flake when new changes arrive or make it as simple as possible and avoid new changes unless really needed.

For example, we could have PRs like these (or the ones I pushed some days ago) require some documentation about the changes and we can ask known users of DMS that uses NixOS to test the PR out (I, for example, can help on that, I just can't guarantee that I would do that in a timely manner).

After we get some Ok's, the PR can be merged.

Also, another thing to keep in mind is maintainability of the flake. For example, this PR adds support for the user to define the `default-settings.json` directly with nix syntax, right now I'm doing something similar by having the `default-settings.json` file in my nix config repo and nix just copy it to the correct `~/.config/DankMaterialShell` directory. Personally I really like changes like that since it makes the config more declarative IMO and would gladly use it (also, I would ask @asaadmohammed74 to expand it to support `~/.local/state/DankMaterialShell/default-session.json` too), but, at the same time, it means that any new changes that you do to these files (like a new option, renaming an old option, removing one, etc), needs to be updated in this flake too.

Keep in mind that, at least in my opinion, NixOS users normally know how to fix these things themselves (as, in the end of the day, all nix configs are simply code written in the Nix Programming Language), so, if we find something missing or broken we would probably push a PR when we have time.

Also, the nature of NixOS is that you can't really break the system, if, for example, the flake is broken for some reason, or there is a breaking change that would, I don't know, make DMS not load until the user updates some config, it doesn't mean that the user system is broken, the user will just reboot and choose an older generation that doesn't have the bug, then see what the issue is, build again with the fix and be done with it. In other words, even if the flake is broken or have breaking changes, it doesn't mean that you really risk breaking the user system in any way.

bluelinden:
> @sezaru @bluelinden for visibility as well so we may be able to achieve the best solution for Nix users.

i'm no longer using DMS, so i may not be of use anymore. sorry!

Vortriz:
@asaadmohammed74 Looks quite solid from a quick glance. I'll look at it more in depth sometime this week. Although one thing that i'll mention is that i am not sure about the idea of declaratively managing settings at this stage. The project is under rapid iteration and instead of providing options that may lag behind the master branch, we should defer that particular part to a later stage. The pace at which DMS is progressing is quite insane and I am all for it.

So, I'd suggest we keep the flake only for the package and the HM module.

As I said, I'll try to fix all current flake related issues sometime this week (although I don't use hyprland so I might need someone to test it out).

asaadmohammed74:
@Vortriz Thats reasonable, we can keep this as reference for now. When we reach a stable release we can revisit, with maybe the addition of auto generated docs, like niri-flake.

purian23:
Closed in favor of gaining the callab on https://github.com/AvengeMedia/DankMaterialShell/pull/242. Thanks for the PR!

PR #191, opened 2025-09-12
Add support for linux-wallpaperengine wallpapers

sgtaziz:
Coming from KDE, I really missed having my WallpaperEngine live wallpapers (https://github.com/catsout/wallpaper-engine-kde-plugin/)

This PR adds support for WallpaperEngine wallpapers using https://github.com/Almamu/linux-wallpaperengine.
I'm not sure if this feature is something this project wants to delve into or not, but I implemented it locally (on Hyprland, not sure about niri) and it works very well, so figured I'd submit the PR and gauge interest. 

Few things to note:
1. WallpaperEngine wallpapers are essentially videos being drawn using wlr-layer-shell. That means, Lockscreen wallpapers and Matugen theme generation won't work directly. I'm using --screenshot flag when running linux-wallpaperengine and saving it in the Cache dir, which is then used by Matugen and LockScreenContent. I added a delay as well since screenshots aren't immediately available.
2. Currently, I'm assuming WallpaperEngine will be installed in Steam's default directory. This isn't always the case, as users could have a secondary Steam library, or just the wallpaper engine scenes directly without installing WallpaperEngine from Steam. Maybe this can be configured to better handle these cases
3. I had to make some changes to the FileBrowser modal, since Wallpapers are folders not direct files. Each folder contains preview.(png/jpg/gif/webp) etc. I tried to handle these different extensions using onStatusChanged and catching the error, then looping through the next file extension, but maybe there is a better way to do this.
4. linux-wallpaperengine has some handy configuration flags, such as --silent to silence backgrounds that have audio, --scaling, --fps, and --set-property to set wallpaper-specific configuration options. I currently have --silent coded into it directly (who really wants audio with their background?), but all these options could be configurable in the Settings modal if needed as well. 

I'm pretty new to QML/Quickshell, so apologies if any of my implementations aren't ideal!

image
image

### DIFF
--- a/Modules/WallpaperBackground.qml
+++ b/Modules/WallpaperBackground.qml
@@ -43,12 +43,6 @@ LazyLoader {
                     monitor: modelData.name
                 }
 
-                Component.onCompleted: {
-                    if (isWE) {
-                        weProc.start(source.substring(3))
-                    }
-                }
-
                 Component.onDestruction: {
                     weProc.stop()
                 }


sgtaziz:
I don't mind having it integrated within the same wallpaper picker as you mentioned - actually would be a lot cleaner. Issue is, wallpaper engine wallpapers (or scenes) aren't files - they're folders. What you are selecting is that folder, not necessarily a file within it. Inside that folder is also the preview image that I am currently using to draw the previews in the file browser modal. For reference, here is the folder structure:
image
And inside one of the "scenes":
image
For this specific example, its an HTML-based wallpaper. Sometimes its videos (mp4), other times its a scene with a .pkg file and shaders. What we are passing to linux-wallpaperengine is the folder name of that scene, which is also its scene ID. I guess we could have it so you select the preview file, but that would mean you have to go into the scene directory to see the preview, go back and go to another scene directory, see the preview there, etc. Not very user-friendly. If you have an idea on how we could implement this better, I would be more than happy to oblige! 

As for the the hardcoded folder, theoretically you are right. By default, linux-wallpaperengine looks in these following paths according to the project's README:
```
~/.steam/steam/steamapps/common
~/.local/share/Steam/steamapps/common
~/.var/app/com.valvesoftware.Steam/.local/share/Steam/steamapps/common
~/snap/steam/common/.local/share/Steam/steamapps/common
```
Path 1 and 2 are the same (.steam/steam is a symlink of .local/share/Steam). However, if Steam is installed using snap, the hardcoded path would not work as it is expected path 4. The other option would be to have an `assets` folder next to the linux-wallpaperengine binary, which is not usable if it was installed from AUR as it was for my case. WallpaperEngine currently can ONLY be installed through Steam, therefore it will always be in a steam library. Technically it can be in a SteamLibrary in a different drive, but linux-wallpaperengine itself doesn't currently support this.

bbedward:
I'm going to merge and make some tweaks after the fact

bbedward:
image

I bueried it behind this option on wallpaper chooser

sgtaziz:
> image
> 
> I bueried it behind this option on wallpaper chooser

That looks great, thank you! Really appreciate the work you and the others have been putting in into this shell.

sgtaziz:
@bbedward I noticed in the change you implemented, you also removed the --screenshot flag from WallpaperEngineProc. This caused Lockscreen Background/Matugen support to break, since it no longer has a static image to reference from. Was this intended?

bbedward:
I wasnt working at it whollistically enough I guess, i thought it was breaking multi-monitor setup but it turns out the --screenshot option is just generating a black image for me - idk maybe my build of the wallpaperengine. I re-added it

sgtaziz:
Hey @bbedward! I'm thinking of moving this into a plugin to make it a bit more modular. Is this something you and the team would be interested in?

bbedward:
I think it'd make more sense that way yea

PR #201, opened 2025-09-14
Add an ArchUpdater widget

sgtaziz:
Add an update widget for Arch-based systems. Requires AUR helper (paru and/or yay, will default to paru if available.)
image

Might be nice if we can somehow disable the widget from the list of widgets to be picked if not an Arch-based system.

### DIFF
--- a/Modules/ArchUpdaterPopout.qml
+++ b/Modules/ArchUpdaterPopout.qml
@@ -1,0 +1,353 @@
+import QtQuick
+import QtQuick.Effects
+import Quickshell
+import Quickshell.Io
+import Quickshell.Wayland
+import Quickshell.Widgets
+import qs.Common
+import qs.Services
+import qs.Widgets
+
+DankPopout {
+    id: archUpdaterPopout
+
+    property var parentWidget: null
+    property string triggerSection: "right"
+    property var triggerScreen: null
+
+    function setTriggerPosition(x, y, width, section, screen) {
+        triggerX = x;
+        triggerY = y;
+        triggerWidth = width;
+        triggerSection = section;
+        triggerScreen = screen;
+    }
+
+    popupWidth: 400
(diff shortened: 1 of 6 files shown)


Vantesh:
> It currently requires TERMINAL env variable to be set, otherwise fallback to xterm. I'm not sure if there is a way to open default term through XDG or similar?

You can try out this [xdg-terminal-exec](https://github.com/Vladimir-csp/xdg-terminal-exec), It allows opening the default terminal or any terminal added  in .config/xdg-terminals.list

bbedward:
I'm going to do some mods to this after merging to make it more generic per-distro, but I like the idea

Vantesh:
> I'm going to do some mods to this after merging to make it more generic per-distro, but I like the idea

Suggestion: Add an option to hide it when there are no updates.

bbedward:
My mods are kind of a fail because paru and yay don't let you run with sudo so there's no non-interactive path. With pacman or dnf or other distros I could bring up a polkit dialog :cry:  but not with paru, not sure what the solution to that is.

sgtaziz:
I'm not sure, but maybe we can avoid AUR updates all together if you really want it non-interactive? there is a "checkupdates" command on Arch, which shows all non-AUR updates. Then instead of an AUR helper ig we can do pacman directly with polkit. Though might be risky to do non-interactive updates, for example if a kernel or a kernel driver update failed and you didn't know, you might not be aware until the next time you restart

I think even if we get paru/yay working, idk how safe/secure it is to build AUR packages as root, so sticking to non-AUR only might be best in that case

bbedward:
I just thought it'd be cool to be able to contain the update output into the shell rather than opening an xterm, but the AUR helpers are tricky because they cant be run as root - the TERMINAL env var is fine for now - it works well enough I think.

PR #242, opened 2025-09-22
Optimize Nix Flake for Niri / Hyprland

sezaru:
Not sure if we want to merge this since it changes a lot of the flake structure, but this do make it possible to use the flake without niri from my tests.

If we decide to merge this, it would be great that others who had issue with the flake before test it out before too


bbedward:
I'm ok with a breaking flake change like this, but can wait for others to give it a go. Maybe would be nice to crudely document the flake options.

nix/niri.nix
```diff
@@ -0,0 +1,10 @@
+{ self, dgop, dms-cli, ... }: {config, pkgs, lib, ...}: let
+  cfg = config.programs.dankMaterialShell;
+in {
+  options.programs.dankMaterialShell = (import ./options.nix { inherit lib pkgs; }).niri;
+
+  config = lib.mkIf cfg.enable (lib.mkMerge [
+    (import ./configs/default.nix { inherit self cfg lib pkgs dgop dms-cli; })
+    (import ./configs/niri.nix { inherit config cfg lib; })
+  ]);
+}
```
sysedwinistrator: You could reduce the complexity here by using standard home-manager conventions, that is:
- Pass the flake's packages by injecting them via the `nixpkgs.overlays` option. You could do this directly in `flake.nix` by writing an inline home-manager module that imports the other modules
- For the Niri-only options, simply create a separate module file that defines `options.programs.dankMaterialShell.niri.*` and configures niri accordingly
  - The module file would only be imported when using the separate Niri homeModule
  - IMHO the separate Niri homeModule should only provide the niri-option and be imported by the user in addition to the default module
  - So basically, instead of having two "flavors" of the homeModule, we would just have an additional one for Niri

sezaru: Do you mind making a patch or fork with these changes? I'm in no way an expert with Nix, so I'm not sure I know exactly how to do these changes.

purian23:
Thanks everyone, we appreciate the collab on this one so we can hopefully implement it with success for most users.

sezaru:
> I'm ok with a breaking flake change like this, but can wait for others to give it a go. Maybe would be nice to crudely document the flake options.

I just added some documentation on how to use in the README.md.

luis-agm:
I did my [best](https://github.com/sezaru/DankMaterialShell/pull/1) to address the comments from @sysedwinistrator , which I think makes sense and makes it a bit simpler. Testing locally with my home-manager config with Hyprland seems to work fine but will do more testing later. Haven't tested with Niri since I don't use it, but I believe the main change there would be that you need to add both homeModules instead of just one.

sezaru:
Thanks! I will test and review it tonight!

sysedwinistrator:
> I did my [best](https://github.com/sezaru/DankMaterialShell/pull/1) to address the comments from @sysedwinistrator , which I think makes sense and makes it a bit simpler. Testing locally with my home-manager config with Hyprland seems to work fine but will do more testing later. Haven't tested with Niri since I don't use it, but I believe the main change there would be that you need to add both homeModules instead of just one.

I'm a bit busy these days, so thanks for stepping up and doing the work! I already took a look at your PR, there's one small error but other than it's just what I had in mind.

sezaru:
Hey @luis-agm I did a code review in your PR and I think I saw only one issue where the niri config is inside another niri level.

Meaning that to set niri configs, you would do something like

```nix
programs.dankMaterialShell.niri.niri.enable = true;
```

Also, if I understood it correct, this changes makes that you need to import both the default home module but the niri as-well if you want niri support correct? Honestly I don't recall seeing a package in nix that does that, all the ones I'm using or used in the past you only import one of the home modules.

But maybe it is just that I'm not aware of them.

Either way, I'm fine with that change, but the documentation in the README file will also need to be updated to reflect that.

luis-agm:
I've fixed the niri config levels, thanks for catching that! Also updated the docs.

About the 2 module concern, I admit I also find it unusual. But architecturally speaking, I find it better than the alternative of having `niri-flake` as a dependency of this whole flake. This 2 module approach provides flexibility for the users to either include an optional, configuration-focused integration module, or manage the keybindings and spawn themselves. The latter I think will be a pretty common scenario for people coming from other shells, that are very used to their current keybindings and don't necessarily want to learn new ones. 

In my case with Hyprland, I've set conditions in my `hyprland.nix` to register my keybindings when DMS is enabled. This also allowed me to stick to `hyprlock` while I figure out why DMS lockscreen won't unlock with the correct password, but I still get to enjoy and learn more about DMS while I do that. 😄

sezaru:
> About the 2 module concern, I admit I also find it unusual. But architecturally speaking, I find it better than the alternative of having `niri-flake` as a dependency of this whole flake. This 2 module approach provides flexibility for the users to either include an optional, configuration-focused integration module, or manage the keybindings and spawn themselves. The latter I think will be a pretty common scenario for people coming from other shells, that are very used to their current keybindings and don't necessarily want to learn new ones.

But that (not having niri-flake as a dependency for the whole flake) holds true for both the PRs no? The only difference I can see if that with my PR you only import one home module, and for the new one you need to import both of them (if you want niri support, ofc).

Regardless, it is a minor thing.

I will be testing your PR with niri today and if it works great, I will be merging it so we can prepare to merge this PR too :smile:

sezaru:
All right, @luis-agm changes are applied to this PR. I'm good merging it if others don't see any other issue with it.

luis-agm:
Not sure how you guys manage releases and stuff. How would you communicate this breaking change?

bbedward:
> Not sure how you guys manage releases and stuff. How would you communicate this breaking change?

I mean, we don't really have any way to do that - but it's alright they'll come to github/discord/matrix if they have an issue.

PR #268, opened 2025-09-29
feat: WorldClock Bar Widget

rochacbruno:
> [!note]
> I wanted to add this as a plugin, or local only module
> but I could not find how to do that
> if someone point me to the direction I would close the PR
> and have this as a custom module.

Screenshot from 2025-09-29 17-07-21

Settings

Screenshot from 2025-09-29 17-08-01


rochacbruno:
being replaced by the plugin

PR #278, opened 2025-09-30
Add emacs-/readline- and vim-style list navigation to app launcher

ambirdsall:
### the problem I'm trying to fix with this PR
I have a lot of muscle memory for emacs-/readline-style `ctrl-n`/`ctrl-p` navigation of lists, and also for vim-style `hjkl` navigation of, well, most things. Because of this, I frequently find myself running face-first into this slightly, but consistently, frustrating scenario:
- I open the app launcher
- I see the item I want in the second or third position
- I type `ctrl-n` to select it
- nothing happens

The real fly in the proverbial ointment is that, since I heavily use a hand-rolled script to focus or move windows to named workspaces with `fuzzel` (which _does_ support `ctrl-{n,p}` navigation), this muscle memory is constantly getting reinforced inside my niri/dms setup; my only options are to patch it in or be slightly annoyed forever.

### what it actually does
Adds a few new keybindings to the spotlight modal:
- `ctrl-n` goes forward in a list/grid agnostic way like `tab`
- `ctrl-p` does the inverse, like `backtab`
- `ctrl-{h,j,k,l}` behave equivalently to the arrow keys

I also made a tiny drive-by fix of a typo in `CONTRIBUTING.md`, which I thought wasn't worth it's own PR; I'm happy to split that out if you'd prefer to keep the two changes separate.

### open questions and choices
Feel free to merge as-is, but these are the potential changes that make sense to me up-front:
- I only added the new bindings to `Modals/Spotlight/SpotlightContent.qml`, but scanning the codebase for `select{Next,Previous}()` or `Key_{Up,Down}`, I see a few other spots where an equivalent change would, or at least might, make sense to include:
  - `Widgets/DankDropdown.qml`
  - `Modals/Clipboard/ClipboardKeyboardController.qml`
  - `Modals/PowerMenuModal.qml`
  - `Modules/AppDrawer/AppDrawerPopout.qml`
  - `Modules/Notifications/Center/NotificationKeyboardController.qml`
  - `Modules/Settings/WidgetSelectionPopup.qml`
  - maybe `Modals/Common/ConfirmModal.qml`
  - maybe `Modals/FileBrowser/FileBrowserModal.qml` but maybe not: this one feels like the biggest stretch
- I feel much more strongly that the emacs/readline style bindings should be included than the pseudo-vim ones; they're much more standardized and ubiquitous, while `ctrl-{h,j,k,l}` is a more niche adaption of a modal-native binding scheme. Happy to pare this down to only add `ctrl-{n,p}` if that's the preference of a maintainer

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 # Contributing
 
-Contributions are welcome and encourages.
+Contributions are welcome and encouraged.
 
 ## Formatting
 


bbedward:
I don't think there's any problem supporting vim+emacs as there's no conflicts here

PR #298, opened 2025-10-03
add nixos support for greeter

Pacman99:
didn't test hyprland but works with niri

### DIFF
--- a/nix/greeter.nix
+++ b/nix/greeter.nix
@@ -103,15 +103,20 @@ in {
             };
         };
         systemd.services.greetd.preStart = ''
-            cd /var/lib/dmsgreeter/
-            cp ${lib.concatStringsSep " " cfg.configFiles} .
+            cd /var/lib/dmsgreeter
+            ${lib.concatStringsSep "\n" (lib.map (f: ''
+                if [ -f "${f}" ]; then
+                    cp "${f}" .
+                fi
+            '') cfg.configFiles)}
+
             if [ -f session.json ]; then
-                cp "$(${lib.getExe pkgs.jq} -r '.wallpaperPath' session.json)" wallpaper.jpg
-                mv session.json session.orig.json
-                ${lib.getExe pkgs.jq} '.wallpaperPath = "/var/lib/dmsgreeter/wallpaper.jpg"' session.orig.json > session.json
-                chown ${user} *
+                if cp "$(${lib.getExe pkgs.jq} -r '.wallpaperPath' session.json)" wallpaper.jpg; then
+                    mv session.json session.orig.json
+                    ${lib.getExe pkgs.jq} '.wallpaperPath = "/var/lib/dmsgreeter/wallpaper.jpg"' session.orig.json > session.json
+                fi
             fi
-            cd -
+            chown ${user}: *
         '';
         programs.dankMaterialShell.greeter.configFiles = lib.mkIf (cfg.configHome != null) [
             "${cfg.configHome}/.config/DankMaterialShell/settings.json"


bbedward:
I ripped out your change for XDG_DATA_DIRS since that, is how it should be done and just pushed that to master - thanks.

For the flake changes, I want to just get confirmation from some other nix guys since I have no idea if this is a breaking change or if it effects anything.

@sezaru  @luis-agm  @sysedwinistrator

(I know, not enough about nix to approve nixos-related changes)

Pacman99:
> I ripped out your change for XDG_DATA_DIRS since that, is how it should be done and just pushed that to master - thanks.

I think you got my doc change also on master

bbedward:
> > I ripped out your change for XDG_DATA_DIRS since that, is how it should be done and just pushed that to master - thanks.
> 
> I think you got my doc change also on master

Yea, it's fine (accident). This looks fine to me though just want to get a check from one of the other nixos contributors.

Pacman99:
Did you have permission issues with syncing a users config files? Not sure if this is nixos or my config specific but my home directory has 700 permissions, so all the DMS config files are not visible. And even if I change permissions it gets reset eventually, but probably 700 is safer.

bbedward:
You're correct, on arch the dms gets installed to /etc/xdg... So greetd user is fine. On danklinux it configures greetd to run as the user that installs the greeter (which bypasses any permission issues accessing dms)

Pacman99:
Pushed a commit to fix permissions by hard linking the config files to the greeter data directory as root in the greetd ExecStartPre

nix/greeter.nix
```diff
@@ -0,0 +1,114 @@
...
+            hotkey-overlay {
+                skip-at-startup
+            }
+
+            environment {
+                DMS_RUN_GREETER "1"
+                DMS_GREET_CFG_DIR "/var/lib/dmsgreeter"
+            }
+
+            spawn-at-startup "sh" "-c" "${pkgs.quickshell}/bin/qs -p ${dmsPkgs.dankMaterialShell}/etc/xdg/quickshell/dms; niri msg action quit --skip-confirmation"
```
bokicoder: Hope there is a quickshell.package option

nix/greeter.nix
```diff
@@ -0,0 +1,114 @@
...
+    config,
+    pkgs,
+    dmsPkgs,
+    ...
+}: let
+    inherit (lib) types;
+    cfg = config.programs.dankMaterialShell.greeter;
+
+    sessionConfigs = {
+        niri = pkgs.writeText "niri.kdl" ''
```
bokicoder: niri needs to configure the cursor.
```
cursor {
    xcursor-theme "Bibata-Modern-Classic"
    xcursor-size 24
}

```

bokicoder: Is it better to disable animations?
```
animations { off; }
```

nix/greeter.nix
```diff
@@ -0,0 +1,114 @@
...
+            env = DMS_RUN_GREETER,1
+            env = DMS_GREET_CFG_DIR,/var/lib/dmsgreeter
+
+            exec = sh -c "${pkgs.quickshell}/bin/qs -p ${dmsPkgs.dankMaterialShell}/etc/xdg/quickshell/dms; hyprctl dispatch exit"
+        '';
+    };
+
+    sessionCommands = {
+        niri = ''
+            ${config.programs.niri.package}/bin/niri -c ${sessionConfigs.niri}
```
bokicoder:
```suggestion
            ${pkgs.dbus}/bin/dbus-run-session ${config.programs.niri.package}/bin/niri -c ${sessionConfigs.niri}
```
Reference https://github.com/NixOS/nixpkgs/blob/nixos-unstable/nixos/modules/programs/regreet.nix#L154

sezaru:
I will see if I can review and try it out today night or tomorrow :)

nix/greeter.nix
```diff
@@ -0,0 +1,114 @@
...
+        fonts.packages = with pkgs; [
+            fira-code
+            inter
+            material-symbols
+        ];
+        environment.systemPackages = with pkgs; [
+            fira-code
+            inter
+            material-symbols
+        ];
```
bokicoder: `fonts.packages` and `environment.systemPackages` have the same content, I think keeping one is enough. If I'm wrong, please ignore

nix/greeter.nix
```diff
@@ -0,0 +1,114 @@
...
+        systemd.tmpfiles.settings."10-dmsgreeter" = {
+            "/var/lib/dmsgreeter".d = {
+                user = "greeter";
+                group = "greeter";
+                mode = "0755";
+            };
+        };
```
bokicoder, line 104: FYI
```
let
  user = config.services.greetd.settings.default_session.user;
in
{
  systemd.tmpfiles.settings."10-dmsgreeter" = {
    "/var/lib/dmsgreeter".d = {
      inherit user;
      group =
        if config.users.users.${user}.group != "" then config.users.users.${user}.group else "greeter";
      mode = "0755";
    };
  };
  assertions = [
    {
      assertion = (config.users.users.${user} or { }) != { };
      message = "dmsgreeter: user ${user} does not exist. Please create it before referencing it.";
    }
  ];
}
```

nix/greeter.nix
sezaru: I wonder if we could read the original file (from Modules/Greetd/assets) and just replace the `spawn-at-startup` line instead of copying it fully here..

It would make it easier to keep it up-to-date, but, as a downside, if things change too much in the file, then it could generate a broken config.

nix/greeter.nix
sezaru: Agreed, this should be an option the same way we do for the home manager part

nix/greeter.nix
sezaru: I'm pretty sure you can also use `${lib.getExe config.programs.niri.package}` here but it is syntax sugar, so fell free to ignore

nix/greeter.nix
```diff
@@ -0,0 +1,114 @@
...
+            exec = sh -c "${pkgs.quickshell}/bin/qs -p ${dmsPkgs.dankMaterialShell}/etc/xdg/quickshell/dms; hyprctl dispatch exit"
+        '';
+    };
+
+    sessionCommands = {
+        niri = ''
+            ${config.programs.niri.package}/bin/niri -c ${sessionConfigs.niri}
+        '';
+        hyprland = ''
+            ${config.programs.hyprland.package}/bin/hyprland -c ${sessionConfigs.hyprland}
```
sezaru: I'm pretty sure you can also use `${lib.getExe config.programs.hyprland.package}` here but it is syntax sugar, so fell free to ignore

nix/greeter.nix
sezaru: I'm pretty sure you can just use `fonts.packages` here.

sezaru (changes requested):

sezaru:
I just tested it, seems to be working fine, but for whatever reason it doesn't seem to get my wallpaper correctly.

Also, I was wondering if there should be an option to customize the niri config used.

For example, in my case I'm using NixOS in my M2 macbook, and one of the things I do in my niri config is set the scale of the monitor (output in niri) to 1.2, otherwise everything is too big.

bbedward:
> I just tested it, seems to be working fine, but for whatever reason it doesn't seem to get my wallpaper correctly.
> 
> Also, I was wondering if there should be an option to customize the niri config used.
> 
> For example, in my case I'm using NixOS in my M2 macbook, and one of the things I do in my niri config is set the scale of the monitor (output in niri) to 1.2, otherwise everything is too big.

Greeter runs in a different context (as a not logged in user, usually - danklinux blurs that line a little). But the guidance is to symlink the configs if you want to keep the greeter in sync with a specific user's settings.

Not sure how that works on nix of course, but other distros `sudo ln -sf /home/superkid6969/.config/DankMaterialShell/settings.json /etc/greetd/.dms/settings.json`

And niri scale yea, should just be configurable in the greeter compositor config itself. I think that's consistent with other greeeters.

nix/greeter.nix
Pacman99: added a `compositor.extraConfig` for this, I don't want to add a default since it won't be configurable

nix/greeter.nix
Pacman99: done

nix/greeter.nix
Pacman99: done

nix/greeter.nix
Pacman99: Added this with `builtins.replaceStrings`, hopefully that will be robust enough but we'll need to keep it updated if anything changes in the assets folder

nix/greeter.nix
Pacman99: Used `lib.makeBinPath` to set PATH so that niri can also be used within the niri `spawn-at-startup` line in the dms-niri.kdl file

nix/greeter.nix
Pacman99: added, thanks!

Pacman99:
> I just tested it, seems to be working fine, but for whatever reason it doesn't seem to get my wallpaper correctly.
> 
> Also, I was wondering if there should be an option to customize the niri config used.
> 
> For example, in my case I'm using NixOS in my M2 macbook, and one of the things I do in my niri config is set the scale of the monitor (output in niri) to 1.2, otherwise everything is too big.

Try the latest commit I just added some logic in the preStart of greetd to get the wallpaper path with jq, copy it to the data directory, then change session.json to reference that file.
Make sure to run `rm /var/lib/dmsgreeter/*` before using the latest commit. I switched from hard links to just copying config files to make editing them easier so there might be some permission issues with overwriting the hard links.

I also changed the `compositor` option to `compositor.name` and added a `compositor.extraConfig`  that will get appended to the greeter compositor config. You can set cursor, monitor scale, and other stuff there. Since its just appending, anything in `dms-niri.kdl` can't be overridden, but it should be enough for most customization.

nix/greeter.nix
Pacman99: Is this necessary? I was just basing the script off of the greet-niri.sh in the greetd assets folder. Will adding this fix anything?

bokicoder: I looked into it, and this seems to be for solving the issue of startup delays in GTK applications. Feel free to close it

sezaru:
I dunno why, but after the latest changes greeter is failing to start here, I can't debug it right now, but I will try to figure out what is happening later.

Anyway, I think the `extraConfig` option should have a default value, otherwise you will get this error when trying to build without that option set:

```
The option `programs.dankMaterialShell.greeter.compositor.extraConfig' was accessed but has no value defined. Try setting the option.
```

Pacman99:
> I dunno why, but after the latest changes greeter is failing to start here, I can't debug it right now, but I will try to figure out what is happening later.
> 

Did you remove the files in `/var/lib/dmsgreeter`? I had an error if I didn't do that once I switched from using ln to cp. Otherwise could you send logs from `greetd.service`, its likely an error from the preStart script which is where most of the changes have been.

> Anyway, I think the `extraConfig` option should have a default value, otherwise you will get this error when trying to build without that option set:
> 
> ```
> The option `programs.dankMaterialShell.greeter.compositor.extraConfig' was accessed but has no value defined. Try setting the option.
> ```

Good catch I added a default.

sezaru:
Indeed, I forgot to remove the files inside `/var/lib/dmsgreeter`, now everything seems to work, including the wallpaper, great job!

sezaru:
Actually, there is something that is bothering me a lot with the current greet.

The generated `dms-greeter` that greetd runs will show all niri logs in the screen before niri starts.

I don't know if there is a "propert" greetd way of avoiding that, but, some workarounds for it is to either lower niri logs to only errors by adding this right before the niri call:

```bash
RUST_LOG=niri=error ${sessionCommands.${cfg.compositor.name}}
```

Or by ignoring the logs but keep a copy in `/tmp`:

```bash
 ${sessionCommands.${cfg.compositor.name}} > /tmp/niri.log 2>&1
```

Maybe we can add something like this as an option?

bbedward:
> Actually, there is something that is bothering me a lot with the current greet.
> 
> The generated `dms-greeter` that greetd runs will show all niri logs in the screen before niri starts.
> 
> I don't know if there is a "propert" greetd way of avoiding that, but, some workarounds for it is to either lower niri logs to only errors by adding this right before the niri call:
> 
> ```shell
> RUST_LOG=niri=error ${sessionCommands.${cfg.compositor.name}}
> ```
> 
> Or by ignoring the logs but keep a copy in `/tmp`:
> 
> ```shell
>  ${sessionCommands.${cfg.compositor.name}} > /tmp/niri.log 2>&1
> ```
> 
> Maybe we can add something like this as an option?

I want to solve it just not sure how yet #302 I'm not sure about just suppressing the logs since idk if that'll hide them from systemd too

nix/greeter.nix
```diff
@@ -0,0 +1,114 @@
...
+                mv session.json session.orig.json
+                ${lib.getExe pkgs.jq} '.wallpaperPath = "/var/lib/dmsgreeter/wallpaper.jpg"' session.orig.json > session.json
+                chown ${user} *
+            fi
+            cd -
+        '';
+        programs.dankMaterialShell.greeter.configFiles = lib.mkIf (cfg.configHome != null) [
+            "${cfg.configHome}/.config/DankMaterialShell/settings.json"
+            "${cfg.configHome}/.local/state/DankMaterialShell/session.json"
+            "${cfg.configHome}/.cache/quickshell/dankshell/dms-colors.json"
```
bokicoder, line 124: These three files missing any one of them will cause the greetd service to fail to start, for example, if I set the environment variable `DMS_DISABLE_MATUGEN=1`, it will not generate the `~/.cache/quickshell/dankshell/dms-colors.json` file.

bokicoder:
> The generated `dms-greeter` that greetd runs will show all niri logs in the screen before niri starts.

Based on my observation, there are only about 3 lines of output on my host, and they flash by without glaring colors (such as red), so I can accept it. Of course, I also understand your frustration

nix/greeter.nix
```diff
@@ -0,0 +1,114 @@
...
+            };
+        };
+        systemd.services.greetd.preStart = ''
+            cd /var/lib/dmsgreeter/
+            cp ${lib.concatStringsSep " " cfg.configFiles} .
+            if [ -f session.json ]; then
+                cp "$(${lib.getExe pkgs.jq} -r '.wallpaperPath' session.json)" wallpaper.jpg
+                mv session.json session.orig.json
+                ${lib.getExe pkgs.jq} '.wallpaperPath = "/var/lib/dmsgreeter/wallpaper.jpg"' session.orig.json > session.json
+                chown ${user} *
```
bokicoder:
```suggestion
                chown ${user}: *
```
Set primary group

nix/greeter.nix
```diff
@@ -0,0 +1,114 @@
...
+            "/var/lib/dmsgreeter".d = {
+                user = user;
+                group = if config.users.users.${user}.group != ""
+                    then config.users.users.${user}.group else "greeter";
+                mode = "0755";
+            };
+        };
+        systemd.services.greetd.preStart = ''
+            cd /var/lib/dmsgreeter/
+            cp ${lib.concatStringsSep " " cfg.configFiles} .
```
bokicoder:
```suggestion
            for file in ${lib.concatStringsSep " " cfg.configFiles}; do [ -f "$file" ] && cp "$file" .; done
```

nix/greeter.nix
bokicoder: Should this command be moved outside the if block from a logical standpoint?

bokicoder:
There is an issue with the current method of synchronizing user profiles, where user profiles are not synchronized when logging out of a Wayland session and returning to the greeter.

nix/greeter.nix
Pacman99: Right I just added this but with `concatStringsSep` instad of bash `for`

nix/greeter.nix
Pacman99: fixed

nix/greeter.nix
Pacman99: fixed

Pacman99:
I added `logs.save` and `logs.path` options to optionally redirect logs to `/tmp/dms-greeter.log` or the configured path. 

> There is an issue with the current method of synchronizing user profiles, where user profiles are not synchronized when logging out of a Wayland session and returning to the greeter.

Good point, but not really sure how to fix this. We only have root access during the greetd preStart and root access is needed to copy over the settings/wallpaper. I figure most people once logged in would use the lock screen instead of greeter so it might not really matter that much.

bokicoder:
> Good point, but not really sure how to fix this. We only have root access during the greetd preStart and root access is needed to copy over the settings/wallpaper. I figure most people once logged in would use the lock screen instead of greeter so it might not really matter that much.

I tried to solve it, but there really isn't a good solution

bokicoder:
Attempted the latest commit, LGTM!

PR #307, opened 2025-10-04
matugen-worker.sh: Change prefer-light to default

YaLTeR:
Mirroring the change in the other place.

Newline change not intended, blame GitHub web editor


YaLTeR:
Got changed on master

PR #309, opened 2025-10-04
Fix toggle function to return correct theme mode

AC17dollars:
fix toggle returning incorrect theme


AC17dollars:
ipc theme toggle returns previous theme instead of currently set theme 
image

PR #353, opened 2025-10-08
Add support for fedora to the SystemUpdateService

metlos:
First of all, thanks for one dank shell 😄 

This PR adds support for listing and applying system updates using `dnf` on Fedora.

This is my first experience with QML so please forgive any sins I'm committing in the code...


bbedward:
I have some thoughts around the system updater maybe making it a plugin and more customizable or something but this is good for now if it works.

PR #354, opened 2025-10-09
Add nix home-manager option for adding plugins

Pacman99:
Chose to have the type be an attribute set instead of a list so modules can disable or override plugins added by other modules.


bbedward:
@sezaru  @bokicoder  looks good to me, but want input from another nix user to see if this is how we wanna handle it.

bokicoder:
LGTM

Pacman99:
Added a note in readme for how to use the option

PR #356, opened 2025-10-09
Add default configuration option to home-manager

bokicoder:


Pacman99:
This should fix it
```nix
        xdg.configFile = lib.mkMerge [
            (lib.mapAttrs' (name: plugin: {
                name = "DankMaterialShell/plugins/${name}";
                value.source = plugin.src;
            }) (lib.filterAttrs (n: v: v.enable) cfg.plugins))
            {
                "DankMaterialShell/default-settings.json" = lib.mkIf (cfg.default.settings != { }) {
                    source = jsonFormat.generate "default-settings.json" cfg.default.settings;
                };
            }
        ];

```

PR #360, opened 2025-10-09
i18n: add Chinese Simplified (zh_Hans)

Moraxyc:
`zh.json` was exported from poeditor.com


bbedward:
I pushed these up, thanks for the translation! Usually, I prefer to pull them down myself from poeditor (or script it which, I'll do when we get more terms). Just to guarantee it's the source of truth.

In the case of chinese, I modified the loader to support the full locale so it picks up zh_CN.json

PR #361, opened 2025-10-09
i18n: add ja (ja_JP)

kamikaze211:
update the new added terms for ja translation

### DIFF
--- a/translations/ja.json
+++ b/translations/ja.json
@@ -30,16 +30,16 @@
         "A file with this name already exists. Do you want to overwrite it?": "この名前のファイルは既に存在します。上書きしてもよろしいですか?"
     },
     "About": {
-        "About": ""
+        "About": "詳細"
     },
     "Access clipboard history": {
-        "Access clipboard history": ""
+        "Access clipboard history": "クリップボードの履歴へのアクセス"
     },
     "Access to notifications and do not disturb": {
-        "Access to notifications and do not disturb": ""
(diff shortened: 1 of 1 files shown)


bbedward:
I pushed these up, thanks for the translation! Usually, I prefer to pull them down myself from poeditor (or script it which, I'll do when we get more terms). Just to guarantee it's the source of truth.

kamikaze211:
> I pushed these up, thanks for the translation! Usually, I prefer to pull them down myself from poeditor (or script it which, I'll do when we get more terms). Just to guarantee it's the source of truth.

> I pushed these up, thanks for the translation! Usually, I prefer to pull them down myself from poeditor (or script it which, I'll do when we get more terms). Just to guarantee it's the source of truth.

You are welcome!
So I just need to translate on the poedit website, and then you will pull them to the github?
It seems will let me work more easily, thank you!

bbedward:
> > I pushed these up, thanks for the translation! Usually, I prefer to pull them down myself from poeditor (or script it which, I'll do when we get more terms). Just to guarantee it's the source of truth.
> 
> > I pushed these up, thanks for the translation! Usually, I prefer to pull them down myself from poeditor (or script it which, I'll do when we get more terms). Just to guarantee it's the source of truth.
> 
> You are welcome! So I just need to translate on the poedit website, and then you will pull them to the github? It seems will let me work more easily, thank you!

Yep, that's right. When we get more languages I'll probably setup some workflow to just export them from poeditor api automatically on changes

PR #366, opened 2025-10-09
Re-introduce default settings option and fix dms-cli build

Pacman99:
fixes a conflict between my plugins PR and @bokicoder's default settings PR

also I'm having a dms-cli build error thats fixed by updating the flake input.


bokicoder:
Thanks.

PR #373, opened 2025-10-10
Modify the greeter module to use the `dms-greeter` wrapper

bokicoder:


RadicalTray:
Should the directory permission be 0770 like the Greetd README or 0750 like SDDM?
not that it matters idk

bbedward:
> Should the directory permission be 0770 like the Greetd README or 0750 like SDDM? not that it matters idk

750 won't allow group members to write to it I believe, greeter group needs write access for its memory file

bokicoder:
The focus is on the permissions of the greeter user. Is the greeter group used by anyone? And currently, 755 hasn't found any issues

bbedward:
Oh right, the user, 750 should be fine. Any user files put there like the wallpaper configs are read only by the greeter so it's fine if group doesn't have write.

RadicalTray:
dms should have the same permission as the greeter user since everything is probably run by the greeter user. The nix module also copies configs including the wallpaper from a user's home to the dmsgreeter directory.
I pointed that out because I noticed nix is currently using 5 for *other* permission but SDDM and the docs use 0 for *other* permission.

bbedward:
Is this good to go? I'm hesitant to approve nix stuff because, it always makes someone unhappy.

RadicalTray:
yeah, it works and nothing changes aside from an option name. the directory permission isn't really a real problem, i just saw sddm and the docs do it like that and wondered if it should be hidden.

PR #396, opened 2025-10-12
Close ControlCenter popout on Settings open.

Bodify:
Closes the ControlCenter popout when you open Settings. Improves UX by not having to manually click elsewhere on the screen to close the ControlCenter popout after being done doing what you wanted in Settings.


Bodify:
Uhm... I may have accidentally deleted this branch due to similar naming.

PR #399, opened 2025-10-12
Make popouts respect screen edge margins and align with windows

Bodify:
Did some tweaks to make popouts respect screen edge margins. This is heavily carried by AI so please review it. I'm not entirely sure how it would look on hyprland. I tested all layouts for DankBar and it works fine.

One consideration is if it should strictly follow the DankBar's Exclusive Zone Offset like it does in this PR or if it should follow the Edge Spacing instead. To me this made the most sense but not by much. Not super confident in this PR but please let me know how it looks to you guys. It makes things look much better on my end at least :3

Some befores and afters:
TooltipOld
TooltipNew
LauncherOld

LauncherNew

### DIFF
--- a/Common/Theme.qml
+++ b/Common/Theme.qml
@@ -17,7 +17,7 @@ Singleton {
     readonly property bool envDisableMatugen: Quickshell.env("DMS_DISABLE_MATUGEN") === "1" || Quickshell.env("DMS_DISABLE_MATUGEN") === "true"
 
     // ! TODO - Synchronize with niri/hyprland gaps?
-    readonly property real popupDistance: 2
+    readonly property real popupDistance: SettingsData.dankBarSpacing
 
     property string currentTheme: "blue"
     property string currentThemeCategory: "generic"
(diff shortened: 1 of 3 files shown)


bbedward:
I do like it, though for niri gaps when we sync the layout kdl we enforce a minimum of 4, so I wonder if we'd want it to match the niri/hyprland window gaps at all times or not. But it feels like a preference

folke:
These changes make a lot of sense if you have spacing around the dankbar, but like in my case, it's not ideal.
I prefer the bar to be full width and without spacing. Now I get this:

image

bbedward:
I think the most sensible thing, would be to synchronize it with the compositor gaps - but that only would work now if using include dms/layout.kdl (on niri) 

I'll think about it and adjust it before the next release

bbedward:
I made a setting override here

image

The default auto - will apply the Math.min(4... logic like the niri gap layout generation, to keep gaps present on no spacing. It basically keeps it in sync with "dms/layout.kdl" if that's included in niri's config always - though we arent yet generating equivalents for hyprland and also don't have compositor-specific settings, particularly.

And being able to override it, s hould allow you to manually do whatever you want.

folke:
Awesome, thanks!

PR #414, opened 2025-10-13
Enhancement: custom system updater command

max72bra:
It is now possible to configure a user-defined system update command along with any additional parameters passed to the terminal (such as the class or title of the opening window). The terminal is always chosen via the TERMINAL environment variable.

### DIFF
--- a/Services/SystemUpdateService.qml
+++ b/Services/SystemUpdateService.qml
@@ -191,11 +191,11 @@ Singleton {
             finalCommand.push(updateCommand)
             updater.command = finalCommand
         } else {
-        const params = packageManagerParams[pkgManager].upgradeSettings.params.join(" ")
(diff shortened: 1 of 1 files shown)


TowelSniffer:
For me the Use Custom Command toggle option doesn't work. I don't seem to get an error but toggling this doesn't update `"updaterUseCustomCommand": true,` in my settings.json. If I edit the json file manually it will work. 

Disabling also removes the custom commands but doesn't change the `updaterUseCustomCommand` value

bbedward:
> For me the Use Custom Command toggle option doesn't work. I don't seem to get an error but toggling this doesn't update `"updaterUseCustomCommand": true,` in my settings.json. If I edit the json file manually it will work.
> 
> Disabling also removes the custom commands but doesn't change the `updaterUseCustomCommand` value

Should be fixed via cc0984db14c1ba566ae993bd3521f4ef96a55e56

PR #425, opened 2025-10-14
Add gSlapper video wallpaper support with thumbnail generation

Nomadcxx:
This PR adds video wallpaper support using gSlapper with the following features:

## New Services
- **GslapperService.qml**: Manages gSlapper processes with crash recovery, mode switching, and safety monitoring
- **ThumbnailService.qml**: Generates and caches thumbnails for both images and videos with resource management
- **VideoThumbnailService.qml**: Specialized service for video thumbnail generation

## Key Improvements
-  process monitoring with safety limits and emergency shutdown
- generation with automatic caching and validation
- sesource management with limits on concurrent processes
- support for both global and per-monitor video wallpaper modes
- automatic video thumbnail generation with ffmpeg/ImageMagick

### DIFF
--- a/Modals/FileBrowser/FileBrowserModal.qml
+++ b/Modals/FileBrowser/FileBrowserModal.qml
@@ -47,6 +47,14 @@ DankModal {
         return ['jpg', 'jpeg', 'png', 'gif', 'bmp', 'webp', 'svg'].includes(ext)
     }
 
+    function isVideoFile(fileName) {
+        if (!fileName) {
+            return false
+        }
+        const ext = fileName.toLowerCase().split('.').pop()
+        return ['mp4', 'mkv', 'webm', 'avi', 'mov'].includes(ext)
+    }
+
     function getLastPath() {
         const lastPath = browserType === "wallpaper" ? SessionData.wallpaperLastPath : browserType === "profile" ? SessionData.profileLastPath : ""
         return (lastPath && lastPath !== "") ? lastPath : homeDir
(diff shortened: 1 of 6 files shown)


bbedward:
Could you add some context about it, I'm not familiar. Like what dependencies this requires, a sample wallpaper file to test with.

purian23:
Screenshots or Video overview would also be helpful when introducing a new PR this feature rich.

Nomadcxx:
> Could you add some context about it, I'm not familiar. Like what dependencies this requires, a sample wallpaper file to test with.
> Screenshots or Video overview would also be helpful when introducing a new PR this feature rich.

Hey lads, sorry for the late reply I forgot all about this PR.  This was something I put together pretty quickly because I had already made the changes on my system for personal use (wanted thumbnails for video wallpapers in the file browser and added gslapper support so could play video wallpapers (supported multi-monitor) both gslapperservice and thumbnailservice) were ancillary to that end and just made thumbnail generation more resource efficient. Looking at recent upstream though there are some significant changes since last week? CachceData.qml and WallpaperBackground.qml etc etc.

It wouldn't take a huge amount of time to fix merge conflicts if you want this feature. The hardest part probably CacheData.qml since this looks like it was removed entirely.

*Let me know if this is something you want and I'll update the PR and add screenshots. As for dependencies, if the user wants to utilize this feature they would need gSlapper, if they don't then they wouldn't need it, it would be entirely optional since obviously normal wallpapers and file browser still work.

PR #431, opened 2025-10-14
Multi batteries support

bb10root:
Support for displaying battery information when multiple batteries are installed in the device.
image

### DIFF
--- a/Services/BatteryService.qml
+++ b/Services/BatteryService.qml
@@ -25,63 +25,14 @@ Singleton {
     readonly property string preferredBatteryOverride: Quickshell.env("DMS_PREFERRED_BATTERY")
 
     // List of laptop batteries
-    property var batteries: []
+    readonly property var batteries: UPower.devices.values.filter(dev => dev.isLaptopBattery)
 
-    // Connections to monitor UPower devices model changes
-    property var devicesConnection: Connections {
-        target: UPower.devices
-
-        function onRowsInserted() {
-            Qt.callLater(root.updateBatteries)
-        }
-        function onRowsRemoved() {
-            Qt.callLater(root.updateBatteries)
-        }
-        function onModelReset() {
-            Qt.callLater(root.updateBatteries)
-        }
-        function onDataChanged() {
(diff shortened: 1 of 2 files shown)


Services/BatteryService.qml
```diff
@@ -24,15 +24,76 @@ Singleton {
...
+        }
+    }
+
+    // Timer to periodically refresh battery information
+    property var updateTimer: Timer {
+        interval: 2000
+        running: true
+        repeat: true
+        triggeredOnStart: true
+        onTriggered: root.updateBatteries()
```
bbedward: Don't need a timer here, UPower.devices would be reactive as a property binding

Services/BatteryService.qml
```diff
@@ -24,15 +24,76 @@ Singleton {
...
+        }
+
+        batteries = arr
+    }
+
+    // Main battery (for backward compatibility)
     readonly property UPowerDevice device: {
         var preferredDev
         if (preferredBatteryOverride && preferredBatteryOverride.length > 0) {
-            preferredDev = UPower.devices.values.find(dev => dev.nativePath.toLowerCase().includes(preferredBatteryOverride.toLowerCase()))
```
bbedward, line 30: We have cases where, people have > 1 battery but only want to display one (because an internal one is dead or something) - should make sure we can still handle that case

bb10root: Am I correct in understanding that if DMS_PREFERRED_BATTERY is set, all information should be for this battery only?

Services/BatteryService.qml
bb10root: ok

Services/BatteryService.qml
bbedward: the current version, only shows 1 battery regardless. But that was overriding the one it picked (it just piicked first laptop battery).

So there's just a few cases people want I guess
- Show the one and only battery 
- Change the battery displayed
- Show all batteries

bb10root: fixed

PR #432, opened 2025-10-14
feat: Use libcanberra for themeable sound effects

Vantesh:
Dms can use `libcanberra` for sound effects, which allows for themeable sounds. To use this feature, we require the `libcanberra` package

```bash
# Arch Linux
sudo pacman -S libcanberra

# Fedora
sudo dnf install libcanberra
```

By default it will default to the freedesktop sounds that are in `/usr/share/sounds/freedesktop/**` but users can change the sound theme by installing new sound themes and then setting the theme name with `gsettings`:

```bash
gsettings set org.gnome.desktop.sound theme-name "$NAME"
```

Where `$NAME` is the name of the sound theme you want to use.

Example:

```bash

# install some sounds 
sudo pacman -S oxygen-sounds # used in kde

# Change to this sound

gsettings set org.gnome.desktop.sound theme-name "'oxygen'"

# To reset to default sound

gsettings reset org.gnome.desktop.sound theme-name

```


bbedward:
I don't think I want to do it exactly like this (but I want to do it), I think I'd rather still wrap in QtMultimedia - and resolve the sound files at runtime based on the theme picked - with maybe an option to still use default sounds and not the system one.

That would probably look like:

- See if  gsettings get org.gnome.desktop.sound theme-name exits  0 (can access libcanberra themes)
- Add a scanner for XDG_DATA_DIRS to detect installed sound themes
- Add a setting "Use System Theme" -> when checked, reveal a dropdown with a list of themes (defaulting to currently picked one via gsettings)

I think it gives more flexibility and don't have to do process spawns.

Vantesh:
Thanks for the feedback! Makes sense on the spawning process part. Closing this PR for now.

bbedward:
Here's the variant I had in mind: 90aab9f4dbdce0f84af7911d4e037a2b611af57b

Not all sound themes have all the same effects under the same name so there is some fallbacks and trickery there - "smooth" one audio changed doesnt play anything audible so I just added a way to have special cases. A little trickier than I thought, but it is functional

Vantesh:
works great! Just some unknown logs idk if they are related

```
[vorbis @ 0x7f192e250a80] Could not update timestamps for discarded samples.
[vorbis @ 0x7f1966e4b380] Could not update timestamps for discarded samples```

bbedward:
> works great! Just some unknown logs idk if they are related
> 
> ```
> [vorbis @ 0x7f192e250a80] Could not update timestamps for discarded samples.
> [vorbis @ 0x7f1966e4b380] Could not update timestamps for discarded samples```
> ```

This is just some log from Qt MediaPlayer when playing oga/ogg files. I'm sure there's some way to suppress it I just havent figured it out yet

PR #473, opened 2025-10-17
enhancement: managed NetworkManager ethernet configurations connectio…

max72bra:
…n from control panel


max72bra:
followed by the modification on the DMS side

bbedward:
Very nice contribution thank you, I ended up bumping expectedApiVersion back down since that makes it a breaking change when it really shouldn't be.

max72bra:
>>  I ended up bumping expectedApiVersion back down since that makes it a breaking change when it really shouldn't be.
no problem!!

thanks for the work of you all

PR #475, opened 2025-10-17
Add Workspace Indicator scrolling

maggster165:
Adds a settings option to scroll through the windows on a workspace rather than scrolling through the workspaces themselves when scrolling on the workspace indicator.

### DIFF
--- a/Modules/Settings/WidgetTweaksTab.qml
+++ b/Modules/Settings/WidgetTweaksTab.qml
@@ -65,6 +65,15 @@ Item {
                                            checked)
                                    }
                     }
+                    DankToggle {
+                        width: parent.width
+                        text: I18n.tr("Window Scrolling")
+                        description: "Scroll through windows, rather than workspaces"
+                        checked: SettingsData.workspaceScrolling
+                        onToggled: checked => {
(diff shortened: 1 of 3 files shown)


Modules/Settings/WidgetTweaksTab.qml
```diff
@@ -65,6 +65,15 @@ Item {
                                            checked)
                                    }
                     }
+                    DankToggle {
+                        width: parent.width
```
bbedward, line 69: I think this should be hidden when !CompositorService.isNiri (same with the functionality)

bbedward (changes requested):

Modules/Settings/WidgetTweaksTab.qml
maggster165: This works with niri, though I have not tested it with Hyprland

maggster165: how would one make that work?

maggster165: just tested, works fine if mouse warping is disabled. I suppose there's no need to do it

PR #497, opened 2025-10-19
feat: Only colour Bluetooth icon in DankBar if a device is connected

### LINKED ISSUE #487: Only color bluetooth icon if a device is connected
## Feature Description

The bluetooth icon **in the bar** should only be colored when there is a device connected.

## Use Case

To quickly be able to determine if there is a bluetooth device connected or not.

## Compositor

Not sure, I only use Niri.

- [ ] All compositors
- [ ] niri
- [ ] Hyprland

## Proposed Solution:

I propose the following scenarios:

Adapter disabled: no icon
Adapter enabled no devices connected: greyed out
Adapter enabled  with devices connected: coloured

## Screenshot

Image

avesst:
This adds a new property `BluetoothService.connected` which determines if there is any currently connected devices and then we use that to determine which colour the bluetooth icon in the DankBar should have. Showcase below:

Adapter disabled:
image

Adapter enabled, no devices connected:
image

Adapter enabled, with device connected:
image

Fixes #487.

### DIFF
--- a/Modules/DankBar/Widgets/ControlCenterButton.qml
+++ b/Modules/DankBar/Widgets/ControlCenterButton.qml
@@ -67,7 +67,7 @@ Rectangle {
         DankIcon {
             name: "bluetooth"
             size: Theme.barIconSize(barThickness)
-            color: BluetoothService.enabled ? Theme.primary : Theme.outlineButton
+            color: BluetoothService.connected ? Theme.primary : Theme.outlineButton
             anchors.horizontalCenter: parent.horizontalCenter
             visible: root.showBluetoothIcon && BluetoothService.available && BluetoothService.enabled
         }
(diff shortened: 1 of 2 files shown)


Modules/DankBar/Widgets/ControlCenterButton.qml
```diff
@@ -169,7 +169,7 @@ Rectangle {
 
             name: "bluetooth"
             size: Theme.barIconSize(barThickness)
-            color: BluetoothService.enabled ? Theme.primary : Theme.outlineButton
+            color: BluetoothService.connected ? Theme.primary : Theme.outlineButton
```
bbedward, line 172: I think we should make it, Theme.surfaceText instead - to match everything else on the bar. And primary when connected.

avesst: I tested it with `Theme.surfaceText` but I thought it was quite hard to discern the difference when using a couple of colour schemes, especially if the colours are leaning toward yellow/orange and you have night mode enabled.

Also, `Theme.outlineButton` is the colour for `NetworkService.networkstatus == "disconnected"`, so I'm not sure what's best here.

PR #525, opened 2025-10-22
feat: a new daily wallpaper from a famous portal

max72bra:
bing wallpaper of the day integration

### DIFF
--- a/Modules/DankBar/Widgets/NotificationCenterButton.qml
+++ b/Modules/DankBar/Widgets/NotificationCenterButton.qml
@@ -15,6 +15,7 @@ BasePill {
             implicitHeight: root.widgetThickness - root.horizontalPadding * 2
 
             DankIcon {
+                id: notifIcon
                 anchors.centerIn: parent
                 name: SessionData.doNotDisturb ? "notifications_off" : "notifications"
                 size: Theme.barIconSize(root.barThickness, -4)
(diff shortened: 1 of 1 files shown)


Common/Paths.qml
```diff
@@ -50,6 +51,25 @@ Singleton {
...
+            }
+        }
+    }
+    function pathExists(path: url, callback) {
+        var stripped = strip(path)
+        pathChecker.command = ["sh", "-c", `test -e '${stripped}'`]
+        pathChecker.callback = callback
+        pathChecker.running = true
+    }
+
```
bbedward: You could use the Proc.runCommand() helper which abstracts basically this

bbedward:
I'd really prefer to see this as a [Plugin](https://github.com/AvengeMedia/DankMaterialShell/blob/master/PLUGINS/README.md) - I think it fits the use case pretty well, then added to the registry - it should actually be fairly straight forward to refactor it as such from what I can see. It's sort of becoming the preference for these types of features.

max72bra:
mmm, this will conflict with other wallpaper management...

max72bra:
need to study plugins and see what i can do, for now refuse this

bbedward:
It should translate pretty well to a plugin, its actually even easier/less code that way. SessionData and that stuff is still accessible (because we dont enforce any restrictions for plugins currently) - when it's done it can go in the [registry](https://github.com/AvengeMedia/dms-plugin-registry)

max72bra:
where do you want the PR for a new plugin? here or in the plugin repo?

bbedward:
You can leave it in your own github account, and then you can open a pull request to the registry: https://github.com/AvengeMedia/dms-plugin-registry - which would just contain some metadata and a link to it.  (Registry plugins show up in DMS, when browsing for plugins)

PR #544, opened 2025-10-24
Feat: add nerd font app icon support

kylerm42:
Before:
image

After:
image

This PR adds support for Nerd Fonts to provide a consistent app icon appearance. I've updated the documentation to recommend FiraCode Nerd Font instead of regular FiraCode, but any Nerd Font can be used. It does require manual mapping of app -> icon, and I've added some to start. It should fall back to the normal app icon if none is mapped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,7 +268,7 @@ sudo dnf copr enable avengemedia/danklinux && sudo dnf install quickshell-git
 ```
 
 #### 2. Install fonts
-*Inter Variable* and *Fira Code* are not strictly required, but they are the default fonts of dms.
+*Inter Variable* and *FiraCode Nerd Font* are not strictly required, but they are the default fonts of dms.
 
 #### 2.1 Install Material Symbols
 ```bash
@@ -279,10 +279,15 @@ sudo curl -L "https://github.com/google/material-design-icons/raw/master/variabl
 sudo curl -L "https://github.com/rsms/inter/raw/refs/tags/v4.1/docs/font-files/InterVariable.ttf" -o /usr/share/fonts/InterVariable.ttf
 ```
 
-#### 2.3 Install Fira Code (monospace font)
+#### 2.3 Install FiraCode Nerd Font (monospace font with icon glyphs)
 ```bash
-sudo curl -L "https://github.com/tonsky/FiraCode/releases/latest/download/FiraCode-Regular.ttf" -o /usr/share/fonts/FiraCode-Regular.ttf
+sudo curl -L "https://github.com/ryanoasis/nerd-fonts/releases/latest/download/FiraCode.tar.xz" -o /tmp/FiraCode.tar.xz
+sudo mkdir -p /usr/share/fonts/nerd-fonts
+sudo tar -xJf /tmp/FiraCode.tar.xz -C /usr/share/fonts/nerd-fonts/
+rm /tmp/FiraCode.tar.xz
 ```
+
+**Note:** FiraCode Nerd Font includes the standard Fira Code font plus thousands of icon glyphs from Font Awesome, Material Design Icons, and more. This enables brand-specific application icons in the workspace indicators.
 
 #### 2.4 Refresh font cache
 ```bash
--- a/Services/AppIconService.qml
+++ b/Services/AppIconService.qml
@@ -1,0 +1,136 @@
+pragma Singleton
+pragma ComponentBehavior: Bound
+
+import QtQuick
+import Quickshell
+import qs.Common
+
+Singleton {
+    id: root
+
+    readonly property var brandIcons: ({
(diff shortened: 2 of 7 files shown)


Services/AppIconService.qml
```diff
@@ -0,0 +1,136 @@
...
+
+        "evolution": "\uf0e0",
+        "kmail": "\uf0e0",
+        "calendar": "\uf073",
+
+        "ticktick": "\udb80\udd34",
+        "home assistant desktop": "\udb81\udfd0",
+        "aws vpn client": "\ue7ad",
+        "tv.plex.plex": "\udb81\udeba",
+        "protonvpn-app": "\udb81\udd82"
```
bbedward, line 108: I think it looks better but I'm not super thrilled about this mapping strategy, I feel like there should be a way to pull all of the app icon unicodes from nerd fonts, and I'm not sure how complete the nerd font catalog is but it would be really annoying to me if I didn't have the icon for an app I use often and it was colored while some weren't - so in that case, I might want a way to not use nerd fonts for it.

Services/AppIconService.qml
```diff
@@ -0,0 +1,136 @@
...
+        if (brandIcons[cleanAppId]) {
+            return brandIcons[cleanAppId];
+        }
+
+        for (const key in brandIcons) {
+            if (cleanAppId.includes(key) || key.includes(cleanAppId)) {
+                return brandIcons[key];
+            }
+        }
+
```
bbedward, line 127: We might want some property, to make sure the font is available and always falling back if not.

```js
    readonly property bool hasNerdIconFont: Qt.fontFamilies().includes("Fira Code Nerd Font")
```

Or something along these lines

README.md
```diff
@@ -279,11 +279,16 @@ sudo curl -L "https://github.com/google/material-design-icons/raw/master/variabl
...
 ```bash
-sudo curl -L "https://github.com/tonsky/FiraCode/releases/latest/download/FiraCode-Regular.ttf" -o /usr/share/fonts/FiraCode-Regular.ttf
+sudo curl -L "https://github.com/ryanoasis/nerd-fonts/releases/latest/download/FiraCode.tar.xz" -o /tmp/FiraCode.tar.xz
+sudo mkdir -p /usr/share/fonts/nerd-fonts
+sudo tar -xJf /tmp/FiraCode.tar.xz -C /usr/share/fonts/nerd-fonts/
+rm /tmp/FiraCode.tar.xz
 ```
 
+**Note:** FiraCode Nerd Font includes the standard Fira Code font plus thousands of icon glyphs from Font Awesome, Material Design Icons, and more. This enables brand-specific application icons in the workspace indicators.
+
```
bbedward, line 291: Also should make a note to update dankinstall, to use the nerd font version across all distros during installation

Services/AppIconService.qml
kylerm42: > I'm not super thrilled about this mapping strategy

Yeah to be honest, neither am I. The other option I considered was [sketchybar-app-font](https://github.com/kvndrsslr/sketchybar-app-font). It has way fewer icons and was made for Mac, but has a nicer mapping by app name instead of unicodes. But either way there would be some manual mapping involved, since the app ids can be weird and won't match the icon name. 

> it would be really annoying to me if I didn't have the icon for an app I use often and it was colored while some weren't - so in that case, I might want a way to not use nerd fonts for it.

Yeah that'll always be a problem with something like this, there's too many apps out there to cover them all. I think a setting to toggle this feature on/off completely would be helpful for people who use a lot of the unmapped apps, but there's a few things we could do for those unmapped ones. Nerd fonts have letter icons, so we could pull the first letter (I think there's already a similar fallback like that). Or we could find some generic default icon. Or we could just display the regular app icon, despite the visual differences. I think I like the first option, but would be open to anything.

Is this something you'd eventually want to merge? I know it's not the most straightforward feature and would take some maintenance, so let me know if I should continue down this path

Services/AppIconService.qml
kylerm42: Yeah I think I'll add a settings toggle for this feature, set it to off and disabled if the font isn't available.

Services/AppIconService.qml
bbedward: Yea, I think it looks better, I actually bundled the font  into the shell so it doesn't need to be a dependency at all (if you reference StyledText or DankIcon for how to load it)

There's probably some way to build a pretty complete mapping of the icons with app IDs, even if some of them are mapped wrong it'd be nice to get a fairly complete mapping that can be fixed later.

kylerm42: Sounds good, I'll put together a larger mapping this week

purian23:
Feel free to reopen in the future.

PR #547, opened 2025-10-25
feat(notification): Add j/k keyboard navigation

avesst:
This just adds `j/k` as an alternative to `up/down` for navigating the notifications in Notification Center.

Not sure if it's necessary to express these key bindings in the cheat sheet, but added just in case.

### DIFF
--- a/Modules/Notifications/Center/NotificationKeyboardController.qml
+++ b/Modules/Notifications/Center/NotificationKeyboardController.qml
@@ -368,7 +368,7 @@ QtObject {
                     onClose()
                 event.accepted = true
             }
-        } else if (event.key === Qt.Key_Down || event.key === 16777237) {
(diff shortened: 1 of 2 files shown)


bbedward:
It already has this with control modifier, IE ctrl+j/k - is that not sufficient? It's like this in pretty much all of the keyboard navigat-able things, some it is required since like spotlight - you want want to steal k/j from teh search field. (Notifications doesnt have a text input, though)

avesst:
Oh, my bad. I noticed you can use `ctrl+j/k` in the clipboard history, but didn't even try it in the notifications. As you say, there is no text input in the notifications and imho without `ctrl` makes it slightly more ergonomic, but I'll leave the decision to you.

bbedward:
I'm going to opt out, I think it's better to keep it consistent rather than have another path.

PR #555, opened 2025-10-26
filebrowser: use xdg paths

max72bra:


max72bra:
very very off-topic: could you please consider my PR on the DANK HOOKS plugin?

PR #585, opened 2025-10-30
fix(dms-greeter,rpm): greeter user is supplied by sysusers and having manual user on the spec breaks it

tulilirockz:
This makes it so this RPM works fine on fedora 43, the greeter user
should be created and configured by systemd sysusers anyways

### DIFF
--- a/dms-greeter.spec
+++ b/dms-greeter.spec
@@ -16,6 +16,8 @@ Source0:        {{{ git_dir_pack }}}
 
 BuildRequires:  git-core
 BuildRequires:  rpkg
+# For the _tmpfilesdir macro.
+BuildRequires: systemd-rpm-macros
 
 Requires:       greetd
 Requires:       (quickshell-git or quickshell)


bbedward:
I don't know about fedora personally, but would this brick the greeter in anyway? The intention was to make sure the greeter user can get permission to write to /var/cache/greeter - primarily for its session + last user memory.

tulilirockz:
That just kinda works fine on greetd, I'll just make sure to test this first on a virtual machine

bbedward:
The main thing is just to ensure the greeter has permission to store its memory, that would just relate to it remembering the last session and successful user sign in. Obviously it's broken entirely right now on F43, Im just wondering if something else is needed to make sure that functionality is preserved on F43

tulilirockz:
Works perfectly fine on a VM

image

purian23:
Thanks again for this one. I will test it in the morning to see if we run into any oddities on F43.

tulilirockz:
Added this build to the Zirconium COPR for better testing: https://copr.fedorainfracloud.org/coprs/zirconium/packages/package/dms-greeter/

PR #591, opened 2025-10-30
base activation for cups capability

max72bra:
related to the commit for cups on the dms repo

### DIFF
--- a/translations/en.json
+++ b/translations/en.json
@@ -932,7 +932,7 @@
   {
     "term": "DMS out of date",
     "context": "DMS out of date",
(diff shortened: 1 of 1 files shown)


bbedward:
I'd like to see a native printer plugin in control center, but the cups dbus integration looks good

max72bra:
It's coming, the time to get it up and running with the service and widget. Since the project is progressing very quickly, I didn't want the sources to deviate too much.
Even so, to give you a clean PR, I had to argue with dwl (only a day late on the commit).   :sob:

Issue #508, opened 2025-10-20
Notification sounds always play through laptop speakers when Bluetooth headset is default


MrMondego:
## Compositor

- [x] niri
- [ ] Hyprland
- [ ] Other (specify)

## Distribution

CachyOS

## dms version

DankLinux Manager v0.1.15

## Description

System/application notification sounds (volume change, new message, etc.) generated by DankMaterialShell are always routed to the built-in laptop speakers even though the Bluetooth headset is set as the default output device in Dank bar or Pavucontrol.
Media players and other apps correctly use the headset; only DMS notifications ignore the setting.

## Expected Behavior

Notification sounds respect the current default sink and play through the same device as every other application.

## Steps to Reproduce

1. Pair and connect Bluetooth headset.
2. Open pavucontrol --> “Output Devices” -->  set headset as default (green check).
3. Verify that music/youtube plays through headset.
4. Trigger any DMS notification:
  - Change volume with media keys, or
  - Receive a chat message that fires a notification sound.
5. Pavucontrol “Playback” tab lists the notification on the wrong sink

## Screenshots/Recordings

https://github.com/user-attachments/assets/9272b02e-c960-4ac8-bbc6-78ee3e3838e4

bbedward:
I'm not able to reproduce it but idk, maybe theres some quirk in QtMultimedia. Does the control center reflect the default audio device always as set in pavucontrol?

That would be, whaht shhows here 

<img width="280" height="87" alt="Image" src="https://github.com/user-attachments/assets/263bdcee-a50a-49a7-b42f-f6fd34d31b9a" />

MrMondego:
Yes, the control center always mirrors the device that is set as fallback in pavucontrol – you can see it in the video attached above.  
<img width="1192" height="98" alt="Image" src="https://github.com/user-attachments/assets/f96b569b-2965-4709-9c9a-320f4286c8d7" />
You can also notice that the notification stream briefly appears on the analogue sink for a split-second in 'Playback' tab.

Extra detail I just discovered:  if I change the headset volume via pavucontrol → ‘Output devices’ tab → Bluetooth device slider, the notification beep is correctly played in the headphones. But every time I use the laptop’s hardware volume keys — those are bound in my Niri config to invoke `dms ipc call audio increment/decrement` — the notification sound comes out of the laptop speakers instead of the current default (Bluetooth) device.

zsolt-donca:
[...]

PR #600, opened 2025-11-01
Trying to make sound effects play on default output

zsolt-donca:
Fixes https://github.com/AvengeMedia/DankMaterialShell/issues/508. Note that this was entirely made by Claude Code; I don't understand how it works or why it works. Feel free to reject this as AI slop. 😅 But it seems to work. I tested it with my Bluetooth headset that I use as a default output device.

### DIFF
--- a/Services/AudioService.qml
+++ b/Services/AudioService.qml
@@ -7,6 +7,7 @@ import QtQuick
 import Quickshell
 import Quickshell.Io
 import Quickshell.Services.Pipewire
+import QtMultimedia
 import qs.Common
 
 Singleton {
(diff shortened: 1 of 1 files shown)


Services/AudioService.qml
```diff
@@ -7,6 +7,7 @@ import QtQuick
 import Quickshell
 import Quickshell.Io
 import Quickshell.Services.Pipewire
+import QtMultimedia
```
bbedward, line 10: This is fine but we can't import QtMultimedia here, just because we call qt6 multimedia an optional dependency - this will break it as compile time if not installed.

I can tweak it to not import it directly though

bbedward:
Try this 5f22347d7aadeb829748a7f435a4ccb0467f341f

zsolt-donca:
> Try this [5f22347](https://github.com/AvengeMedia/DankMaterialShell/commit/5f22347d7aadeb829748a7f435a4ccb0467f341f)

It works great, thanks!

PR #601, opened 2025-11-01
Remove wallpaper engine support in favor of plugin

sgtaziz:
Wallpaper Engine support has been moved to a plugin:
https://github.com/AvengeMedia/dms-plugin-registry/pull/22
https://github.com/sgtaziz/dms-wallpaperengine

### DIFF
--- a/Common/SessionData.qml
+++ b/Common/SessionData.qml
@@ -83,7 +83,21 @@ Singleton {
             if (content && content.trim()) {
                 var settings = JSON.parse(content)
                 isLightMode = settings.isLightMode !== undefined ? settings.isLightMode : false
-                wallpaperPath = settings.wallpaperPath !== undefined ? settings.wallpaperPath : ""
+
+                if (settings.wallpaperPath && settings.wallpaperPath.startsWith("we:")) {
+                    console.warn("WallpaperEngine wallpaper detected, resetting wallpaper")
+                    wallpaperPath = ""
+                    Quickshell.execDetached([
+                        "notify-send",
+                        "-u", "critical",
+                        "-a", "DMS",
+                        "-i", "dialog-warning",
+                        "WallpaperEngine Support Moved",
+                        "WallpaperEngine support has been moved to a plugin. Please enable the Linux Wallpaper Engine plugin in Settings → Plugins to continue using WallpaperEngine."
+                    ])
+                } else {
+                    wallpaperPath = settings.wallpaperPath !== undefined ? settings.wallpaperPath : ""
+                }
                 perMonitorWallpaper = settings.perMonitorWallpaper !== undefined ? settings.perMonitorWallpaper : false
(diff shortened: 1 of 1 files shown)


bbedward:
This is great, i'm going to defer review to @rochacbruno  for the plugin stuff then merge this after plugin is in registry

rochacbruno:
+1 on moving this functionality to a plugin.

I am not an WE user myself, but I am concerned with a breaking change like this without user notice.

Not sure if we can provide an automated migration path, but maybe a notice?


Is WE support something pretty new that would not impact users?

What do you think about adding something like:

```qml
  // In SessionData.qml or shell.qml Component.onCompleted
  if (SessionData.wallpaperPath.startsWith("we:")) {
      // Have to check if plugin is not installed 
      console.warn("Wallpaper Engine moved to plugin - install from plugin registry")
      // OR EVEN SPAWN A NOTIFICATION HERE THAT IS MORE VISIBLE!
      SessionData.wallpaperPath = "" // Reset to default/stock
  }
```

there will be also n orphaned cache on `~/.cache/DankMaterialShell/we_screenshots/` (will the plugin recover the same cache later?)


@bbedward WDYT?

bbedward:
Could show some notification or toast maybe to say it moved to a plugin. It is an old feature, that I don't think many people use just because they probably don't even know it exists - so it may not matter too much.

rochacbruno (approved):
Lets ensure there is a good notice for the breaking functionality and migration path for users.

sgtaziz:
Sure thing! I think toast makes sense here. I'll add it.
Should we do this on only wallpaperPath, or also the others? From a quick glance, I can see:
- wallpaperPath
- monitorWallpapers
- wallpaperPathLight
- wallpaperPathDark
- monitorWallpapersLight
- monitorWallpapersDark

> there will be also n orphaned cache on `~/.cache/DankMaterialShell/we_screenshots/` (will the plugin recover the same cache later?)
> 

Regarding this, the plugin actually uses that same directory still to set the "static" wallpaper. From DMS point of view, it thinks thats what the actual wallpaper is, and matugen/lockscreen works properly. Plugin calls linux-wallpaperengine which draws the animated wallpaper on top of it.

bbedward:
I'm not sure wallpaper engine even supports the per monitor wallpapers, it probably doesn't so just checking the main one is fine. (I could be wrong, there's so many wallpaper options it's hard to track)

sgtaziz:
> I'm not sure wallpaper engine even supports the per monitor wallpapers, it probably doesn't so just checking the main one is fine. (I could be wrong, there's so many wallpaper options it's hard to track)

It does actually, I was using it :smiley:
Realistically, DMS won't crash if there are no wallpapers set. So as long as we catch one and notify, user can go in and change whats needed. I'll go ahead and implement it on wallpaperPath for now.

sgtaziz:
Should be good! For some reason, Toast wasn't working, it kept disappearing shortly after it appears. Made it a notification instead.

PR #620, opened 2025-11-03
matugen: missing foot.ini colors section

ffoebel:


bbedward:
This probably came up, by my best guess, as a result of `dms dank16` not being available, maybe a mismatch between the backend and QML versions, since it adds [colors] when it assembles the final config enriched with dank16.

https://github.com/AvengeMedia/DankMaterialShell/blob/43fbbc07f5c0c185cd8ae5e48bf2888ebb798fe1/scripts/matugen-worker.sh#L365

ffoebel:
Missed that. But I guess it doesn't hurt to have it in the template in case it fails.
Checked the git version I built yesterday from AUR, didn't find `dank16`.
Checked today's git, it's there and I do have the additional colors. :shrug:

PR #632, opened 2025-11-05
CUPS cc integrated widget and service

max72bra:
as per title

### DIFF
--- a/translations/en.json
+++ b/translations/en.json
@@ -588,6 +588,18 @@
     "comment": ""
   },
   {
+    "term": "CUPS Insecure Filter Warning",
+    "context": "CUPS Insecure Filter Warning",
+    "reference": "Services/CupsService.qml:283",
+    "comment": ""
+  },
+  {
+    "term": "CUPS Missing Filter Warning",
+    "context": "CUPS Missing Filter Warning",
+    "reference": "Services/CupsService.qml:282",
+    "comment": ""
+  },
+  {
     "term": "Cancel",
     "context": "Cancel",
     "reference": "Modals/BluetoothPairingModal.qml:249, Modals/PolkitAuthModal.qml:289, Modals/DankColorPickerModal.qml:516, Modals/WifiPasswordModal.qml:492, Modules/Settings/PluginBrowser.qml:627, Modals/FileBrowser/FileBrowserOverwriteDialog.qml:83",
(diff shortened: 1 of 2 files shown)


bbedward:
I don't want to merge this yet, until I figure out how to get it to work at least.. I have cups/arch linux, 1 printer, and my user is in the `cups` group.

```
  WARN  go: [CUPS] D-Bus interface not available. Fallback to log parsing
  INFO  go: CUPS manager initialized
```

For me, I see no printers in the widget - and I dont exactly understand why the dbus interface isn't available. 

But I also am thinking that, maybe the entire cups backend should be re-thought. A better approach might be:

- Use [go-ipp](https://github.com/phin1x/go-ipp) - seems like the best option to me at first look, they implement a [cups client](https://github.com/phin1x/go-ipp/blob/091da977c0e98bca4d813bd1ada1069fc7227df3/cups-client.go#L26) which extends with various methods
  - We can just look at localhost:631 by default, and let users specify an alternative server, with credentials (optionally) if they wanted
- Alternative is using [printer-manager](https://github.com/korylprince/printer-manager-cups) - but that introduces an extra dependency and may not be desirable.

max72bra:
ok, no problem, but unfortunately I've already invested a lot of time and I don't have any more.

max72bra:
I'll be back to make some small corrections, see you around and thank you for your work

max72bra:
Sorry one last time, I would never allow myself to deliver broken code obviously:

immagine

Kooha-2025-11-05-19-16-00

bbedward:
I'm ok to tackle the cups backend - I did flip to go-ipp already and it works fine, but dbus seems like its not present on newer cups, and its lacking ipp subscription (which cups support), so I'm looking into that still.

max72bra:
as I was saying, the dbus status in CUPS is really unclear

bbedward:
Backend revisions for pure IPP implementation: https://github.com/AvengeMedia/danklinux/commit/b1fce7b7f971c90e0639349b6444afc829ec9983 

Please  check with your widget/clinet (some thhings were removed, like `getState`)

bbedward:
I'm just going to make the very minor API tweaks, after merging - I like it a lot, generally.

max72bra:
ok, thanks. I was adapting the widget to the changes, but I think you're already doing them. basically, it's already working (90%)

bbedward:
It seems to work pretty well, made a few fixes. I wanted to get it in for next release (which has to happen urgently since theres a breaking bug for a few users)

bbedward:
Figured out the dbus part too, apparently that works, but requires an ipp event first. That cut down idle cpu usage to effectively 0 again

PR #674, opened 2025-11-09
themes/docs: added a gruvbox light/dark json theme file

motaz-shokry:
2025-11-09-131736_hyprshot


bbedward:
I am OK to add more popular themes to the built-ins, like catpuccin is. Things like gruvbox, nord, rose pine - some of those.

PR #686, opened 2025-11-11
[LICENSE] Relicense from GPL-3.0 to MIT

bbedward:
# Contributors to DankMaterialShell

This document lists all significant contributors to DankMaterialShell who hold copyright over their contributions. For the purpose of license changes, consent is required from contributors listed below.

**Note**: Contributors with only trivial changes (whitespace, minor documentation, <50 lines) are not listed here, as such changes typically do not meet the threshold of originality for copyright.

---

## Primary Authors

These individuals have made extensive contributions (10,000+ lines of code):

1. **bbedward** (BB)
   - Role: Co-Maintainer

2. **purian23**
   - Role: Co-Maintainer

---

## Major Contributors

These individuals have made substantial contributions (1,000+ lines of code OR 10+ commits):

3. **Bruno Cesar Rocha**
   - Contributions: 2,969 lines QML, 183 lines other code
   - Commits: 11 commits

4. **Massimo Branchini** (max72bra)
   - Contributions: 2,088 lines QML, 717 lines other code
   - Commits: 20 commits

5. **Jon Rogers**
   - Contributions: 1,512 lines QML
   - Commits: 13 commits

6. **Aziz Hasanain**
   - Contributions: 1,290 lines QML, 122 lines other code
   - Commits: 5 commits

7. **Aleksandr Lebedev** (KyleKrein)
   - Contributions: 997 lines QML
   - Commits: 11 commits

8. **Mattias**
   - Contributions: 763 lines QML
   - Commits: 4 commits

9. **bokicoder**
   - Contributions: Configuration and build system (96 lines config)
   - Commits: 14 commits

---

## Significant Contributors

These individuals have made notable contributions (100+ lines of code):

10. **Oleksandr**
    - Contributions: 439 lines QML
    - Commits: 3 commits

11. **lonerorz**
    - Contributions: 301 lines QML
    - Commits: 1 commit

12. **Eduardo B. A.** (sezaru)
    - Contributions: 304 lines config/theme files
    - Commits: 5 commits

13. **blue linden**
    - Contributions: 254 lines config/theme files
    - Commits: 5 commits

14. **Rishi Vora**
    - Contributions: 253 lines config/theme files
    - Commits: 3 commits

15. **asaadmohammed74**
    - Contributions: 251 lines QML
    - Commits: 2 commits

16. **Parthiv Seetharaman**
    - Contributions: 87 lines QML, 192 lines config
    - Commits: 4 commits

17. **Gonen Gazit**
    - Contributions: 128 lines QML
[...]

PR #693, opened 2025-11-11
feat: more layer namespaces

claymorwan:
noticed some stuff that i missed that can have layer namespaces so i added theme, i don't think i missed any this time


bbedward:
Can you agree to this, then I'll merge it https://github.com/AvengeMedia/DankMaterialShell/pull/686

Trying to get re-licensed under MIT and avoid any GPL violations

PR #714, opened 2025-11-14
adding tokyonight, everforest, nord and rose-pine themes

kritag:
Added some popular themes, I don't use light themes myself so kinda hard to verify how well they work.


bbedward:
I still stand by that we should add these to the Common/StockThemes ... the more popular ones at least.

PR #715, opened 2025-11-14
Feature/privacy widget - Settings to force icons on

pihomeserver:
Here is a proposal to select wich icons can be forced to be always displayed even if inactive

It adds a new context menu in the settings to select icons to display

image

Maybe default options should be set to false to keep the same behavior ?

PS : I'm not a dev and first time with QML so please be tolerent :-)

### DIFF
--- a/quickshell/Modules/Settings/WidgetsTabSection.qml
+++ b/quickshell/Modules/Settings/WidgetsTabSection.qml
@@ -991,7 +991,7 @@ Column {
 
 
         width: 200
-        height: 120
+        height: 160
         padding: 0
         modal: true
         focus: true


quickshell/Common/SettingsData.qml
```diff
@@ -100,6 +100,11 @@ Singleton {
     property bool controlCenterShowNetworkIcon: true
     property bool controlCenterShowBluetoothIcon: true
     property bool controlCenterShowAudioIcon: true
+    property bool showPrivacyButton: true
```
bbedward, line 103: I think this looks mostly okay, I would prefer to keep the current behavior though. Because when you add it to the bar, you'll see the different options you can configure. But for people who already have it there, they would get a different behavior than before and have to go into the settings to find it.

pihomeserver (commented):
Set default values to false (not visible)

Not a GitHub expert so I hope that's ok

PR #720, opened 2025-11-14
Change default config path to account for monorepo update

sxlphuric:
This pull request changes the line
`primaryPaths = append(primaryPaths, filepath.Join(dir, "quickshell", "dms"))` to `primaryPaths = append(primaryPaths, filepath.Join(dir, "quickshell", "dms", "quickshell"))` to account for the monorepo changes. Because of the monorepo update, the quickshell location was changed and this file was not changed to account for it.

### DIFF
--- a/core/internal/config/dms.go
+++ b/core/internal/config/dms.go
@@ -31,7 +31,7 @@ func LocateDMSConfig() (string, error) {
 
 	for _, dir := range strings.Split(configDirs, ":") {
 		if dir != "" {
-			primaryPaths = append(primaryPaths, filepath.Join(dir, "quickshell", "dms"))
+			primaryPaths = append(primaryPaths, filepath.Join(dir, "quickshell", "dms", "quickshell"))
 		}
 	}
 


core/internal/config/dms.go
```diff
@@ -31,7 +31,7 @@ func LocateDMSConfig() (string, error) {
 
 	for _, dir := range strings.Split(configDirs, ":") {
 		if dir != "" {
-			primaryPaths = append(primaryPaths, filepath.Join(dir, "quickshell", "dms"))
+			primaryPaths = append(primaryPaths, filepath.Join(dir, "quickshell", "dms", "quickshell"))
```
bbedward, line 34: Does it not account for it? It keeps both paths in the search interleaved, so it checks non mono repo path, then monorepo path in the same order as the higher level directories.

bbedward (changes requested):

core/internal/config/dms.go
sxlphuric: When I run the DMS on my computer (i have updated) it does not run the config, and I need to manually pass `-c` for DMS to find it.

bbedward: What's your config path? And what is `dms version` the monorepo was accounted for in the current release

sxlphuric: My config path is at `~/.config/quickshell/dms/quickshell`, and `dms version` says version 0.5.0.

bbedward: You need to build/download v0.5.1 then it should be good. That was the first monorepo release

PR #734, opened 2025-11-16
Add NixOS module

Hikiru:
Adapted the HM module, removed the config stuff and changed the quickshell installation method.

### DIFF
--- a/distro/nix/options.nix
+++ b/distro/nix/options.nix
@@ -1,4 +1,8 @@
-{pkgs, lib, ...}: let
+{
+    pkgs,
+    lib,
+    ...
+}: let
     inherit (lib) types;
 in {
     options.programs.dankMaterialShell = {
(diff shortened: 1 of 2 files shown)


distro/nix/nixos.nix
```diff
@@ -0,0 +1,118 @@
...
+    config,
+    pkgs,
+    lib,
+    dmsPkgs,
+    ...
+}: let
+    cfg = config.programs.dankMaterialShell;
+in {
+    imports = [
+        (lib.mkRemovedOptionModule ["programs" "dankMaterialShell" "enableNightMode"] "Night mode is now always available.")
```
LuckShiba: This option never existed (since the NixOS module, to begin with, never existed), so I think it's not needed here.

distro/nix/nixos.nix
```diff
@@ -0,0 +1,118 @@
...
+
+            Service = {
+                ExecStart = lib.getExe dmsPkgs.dmsCli + " run --session";
+                Restart = "on-failure";
+            };
+
+            Install.WantedBy = [config.wayland.systemd.target];
+        };
+
+        environment.systemPackages =
```
LuckShiba: I think that the basic module options, specially the `enable*` ones should be in a single shared `.nix` file, imported by both home-manager and NixOS modules, since the packages and options are verbatim the same. If someone wanted to add a new dependency or modify the needed, they would always have to unnecessarily have to modify both files.

I can do this in another PR, if you want.

PR #768, opened 2025-11-19
Fedora(dms-greeter): Tell users to use graphical.target so greeter loads at startup

brunodsf05:
This PR updates the Fedora-specific post-install guide of `dms-greeter` by adding an instruction on how to **enable the greeter at startup**.

Users on Fedora 43 Minimal will find that `dms-greeter` doesn't load at startup. That's because:
1. The system defaults to `multi-user.target`.
2. `dms-greeter` only launches automatically if the default target is `graphical.target`.

The fix to this is to set `graphical.target` as the default:
```bash
sudo systemctl set-default graphical.target
```

Here is a video demonstrating the solution: Before the fix the system boots into **tty**, and after the fix it boots directly into **dms-greeter**.

https://github.com/user-attachments/assets/6138d630-e20d-4fe8-a38c-4f7d62e42fe5

### DIFF
--- a/distro/fedora/dms-greeter.spec
+++ b/distro/fedora/dms-greeter.spec
@@ -234,7 +234,10 @@ Next steps:
 2. Enable greetd service:
      sudo systemctl enable greetd
 
-3. (Optional) Sync your theme with the greeter:
+3. Make your system load the greeter at startup:
+     sudo systemctl set-default graphical.target
+
+4. (Optional) Sync your theme with the greeter:
      If you have DankMaterialShell (DMS) installed, you can sync with:
      dms greeter sync
 


purian23:
Thanks for this. Actually there's a newer method that we've built into the DMS Binary where it enables the syncing and such by default.  If we put this message in there as it is, folks would to think they need to run that command on each environment when they don't. It may be needed or optional for minimal installs, but it isn't on top of Gnome and KDE. 

https://danklinux.com/docs/dankgreeter/installation#completing-setup

brunodsf05:
Sorry if this is a dumb question but, is it already implemented?
Because when I tested it like an hour ago it didn't change the default to `graphical.target` thing.

Maybe I could change the message into (optional) or simmilar

purian23:
Hey @brunodsf05 apologies for the delay. I was busy packaging DMS for multi-distro support. From your findings here, I was able to confirm the same on my end with Fedora minimal installs. When users had a base Gnome/KDE installed, it wasn't a factor as it installed correctly as-is. 

We've now made enhancements to the dms-greeter installer itself and updated the post message for Fedora users to leverage the new automation via dms-cli. Feel free to give it a try. 
The changes were made here: https://github.com/AvengeMedia/DankMaterialShell/commit/022102107866ad658e88c3ce21491991c9dcd14d

PR #786, opened 2025-11-22
feat: add autohide and settings ipc functions

mbpowers:
Added IPC functions for `autHide`, `manualHide` and `toggleAutoHide`, as well as general settings getter `get arg` and setter `set arg value`.


bbedward:
I don't want to merge this one yet, because the multi bar branch is nearly ready to be merged and I opted to not keep backwards compatibility with bar related IPCs (in that world, auto hide becomes a per-bar setting not a global one) so the IPCs change like `bar toggle 1-4` (because optional arguments aren't possible via IpcHandler afaik) so the bar related one at least, should follow the sam paradigm basically.

mbpowers:
Ah, makes sense. I can update this for multi-bar when that arrives unless the IPC is included, in which case I can separate out just the general settings getter and setter.

I think you are right about the optional arguments, I came across that as well.

bbedward:
Multi bar has been merged to master

mbpowers:
Updated :+1: Also I changed `reveal` to `show` as we kinda had mixed language

quickshell/Common/SettingsData.qml
```diff
@@ -1392,7 +1392,57 @@ rm -rf '${home}'/.cache/icon-cache '${home}'/.cache/thumbnails 2>/dev/null || tr
...
+                console.warn("Failed to set property:", arg, "error:", e)
+                return "SETTINGS_SET_FAILURE"
+            }
+        }
+
+        target: "settings"
+    }
+
+    IpcHandler {
+        function show(): string {
```
bbedward: `show` doesnt work in quickshell IPC context, its a reserved word I guess and it just makes it show the available functions. That's why they were `reveal` instead

bbedward (changes requested):

quickshell/Common/SettingsData.qml
mbpowers: dang it, of course I change that right at the end and didn't retest lol

quickshell/Common/SettingsData.qml
```diff
@@ -1413,4 +1463,72 @@ rm -rf '${home}'/.cache/icon-cache '${home}'/.cache/thumbnails 2>/dev/null || tr
...
+        function toggleAutoHide(selector: string, value: string): string {
+            if (!barSelectors.includes(selector)) { return "BAR_INVALID_SELECTOR" }
+            const index = selector == "index" ? value : barConfigs.findIndex((bar)=> bar[selector] == value)
+            const barConfig = barConfigs[index]
+            if (!barConfig) { return "BAR_NOT_FOUND" }
+            updateBarConfig(barConfig.id, {autoHide: !barConfig.autoHide})
+            return root.dankBarAutoHide ?  "BAR_MANUAL_HIDE_SUCCESS": "BAR_AUTO_HIDE_SUCCESS"
+        }
+
+        target: "bar"
```
bbedward: We already have target "bar" in DMSShellIPC.qml , some of it is duplicated. Better to keep a singular handler registered.

mbpowers: Ah I missed those. I moved all three of the IpcHandlers from SettingsData into DMSShellIPC. I had a duplicate `settings` handler as well.

PR #787, opened 2025-11-22
feat: add sun and moon view to WeatherTab

mbpowers:
Added a sun and moon with moonphases view to the weatherTab. All data is calculated locally using location and date with `suncalc.js`. Also adds buttons to change the date and time.

https://github.com/user-attachments/assets/19c63e97-4499-4e15-a90c-e3c366ca3eed

I wanted to send this as a draft PR to discuss changes and how to integrate it if wanted. Also this is my first time really working with QML so if I am doing something in the wrong way or with a bad patter you notice please call me out.

Some questions/thoughts:

Should I add this as an option in the settings? Maybe it should be hidden and you can click for it to expand? Not sure if this should be on by default, it distracts somewhat from the clean look.

I can keep or remove the day/time selector as I was just using it for testing. I think it could be a nice feature though, not just to see the sun/moon move, but to be able to see the forecast further into the past/future. I think I can make the UI a little nicer without needing 4 sets of arrows as well. Also I think the weather info above the 7 day forecast should probably change with the given day selected unlike what is show in the video.

Also for slightly more accurate sunrise/set times the functions in suncalc take an optional altitude in meters that I have left out for now, maybe we could fetch that with location somehow? Or we can always ask for that manually in the settings.

I was also thinking about adding the weather in the sky and some stars.

### DIFF
--- a/quickshell/Services/WeatherService.qml
+++ b/quickshell/Services/WeatherService.qml
@@ -1,12 +1,11 @@
 pragma Singleton
-
 pragma ComponentBehavior: Bound
 
 import QtQuick
 import Quickshell
 import Quickshell.Io
 import qs.Common
-import '../Common/suncalc.js' as SunCalc
+import "../Common/suncalc.js" as SunCalc
 
 Singleton {
     id: root
@@ -15,25 +14,25 @@ Singleton {
 
     property var selectedDate: new Date()
     property var weather: ({
-                               "available": false,
-                               "loading": true,
-                               "temp": 0,
-                               "tempF": 0,
-                               "feelsLike": 0,
-                               "feelsLikeF": 0,
-                               "city": "",
-                               "country": "",
-                               "wCode": 0,
-                               "humidity": 0,
-                               "wind": "",
-                               "sunrise": "06:00",
-                               "sunset": "18:00",
-                               "uv": 0,
-                               "pressure": 0,
-                               "precipitationProbability": 0,
-                               "isDay": true,
-                               "forecast": []
-                           })
+            "available": false,
+            "loading": true,
+            "temp": 0,
+            "tempF": 0,
+            "feelsLike": 0,
+            "feelsLikeF": 0,
+            "city": "",
+            "country": "",
+            "wCode": 0,
+            "humidity": 0,
+            "wind": "",
+            "sunrise": "06:00",
+            "sunset": "18:00",
+            "uv": 0,
+            "pressure": 0,
+            "precipitationProbability": 0,
+            "isDay": true,
+            "forecast": []
+        })
 
     property var location: null
     property int updateInterval: 900000 // 15 minutes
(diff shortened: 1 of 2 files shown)


quickshell/Common/suncalc.js
```diff
@@ -0,0 +1,307 @@
+// Copyright (c) 2025, Vladimir Agafonkin
```
bbedward, line 2: it's good to add `.pragma library` here, it may not matter much because there isnt any shared state here - but it might prevent QML from re-creating new contexts or something every time its imported (not sure how QT handles that under the hood)

quickshell/Services/WeatherService.qml
```diff
@@ -149,6 +150,127 @@ Singleton {
...
+    function getCurrentSunTime(date) {
+        const times = getSunTimes(date)
+        const dateObj = new Date(date)
+
+        const periods = [
+            { name: "Dawn (Astronomical Twilight)", start: new Date(times.nightEnd),      end: new Date(times.nauticalDawn) },
+            { name: "Dawn (Nautical Twilight)",     start: new Date(times.nauticalDawn),  end: new Date(times.dawn) },
+            { name: "Dawn (Civil Twilight)",        start: new Date(times.dawn),          end: new Date(times.sunrise) },
+            { name: "Sunrise",                      start: new Date(times.sunrise),       end: new Date(times.sunriseEnd) },
+            { name: "Golden Hour",                  start: new Date(times.sunriseEnd),    end: new Date(times.goldenHourEnd) },
```
bbedward: Good to add I18n.tr() around all user-facing strings, that will get them parsed out for translations (I knoww that some are missed in the codebase like the other weather ones above)

bbedward (changes requested):

bbedward:
> Should I add this as an option in the settings? Maybe it should be hidden and you can click for it to expand? Not sure if this should be on by default, it distracts somewhat from the clean look.

I think it should just be on, I like it as an option - doesnt take anything away just adds more detail.

> I can keep or remove the day/time selector as I was just using it for testing. I think it could be a nice feature though, not just to see the sun/moon move, but to be able to see the forecast further into the past/future. I think I can make the UI a little nicer without needing 4 sets of arrows as well. Also I think the weather info above the 7 day forecast should probably change with the given day selected unlike what is show in the video.

The day/time selector indeed would be better if it could show hourly forecast and moon phase - would give some way to show hourly forecast as well at least, Also the button UI, I'm not a fan but not sure what the best approach is. Maybe a slider of some kind.

> Also for slightly more accurate sunrise/set times the functions in suncalc take an optional altitude in meters that I have left out for now, maybe we could fetch that with location somehow? Or we can always ask for that manually in the settings.

I don't think slightly more accurate is worth it for a manual setting, or another API call. But if it comes back fron open-meteo or one of our 

> I was also thinking about adding the weather in the sky and some stars.

Could maybe make some shaders to get some effects too, but I'm not sure its worth it really

bbedward:
My issue with this right now is how slow it makes opening the weather tab (like clicking the weather widget), like something is blocking the UI. I tried briefly to tweak the loading but didn't see anything immediately obvious as to waht the bottleneck is

mbpowers:
Not sure if this is the ideal way but, I think this fixed it (at least it feels as fast as the other popups to me). For some reason the listviews were blocking when created

bbedward:
I'll get to reviewing this soon, probably tomorrow. I have a couple changes in mind I'll probably push to your fork tomorrow.

bbedward:
I pushed a bunch of stuff to your branch, wrapping tab content in loaders, some spacing adjustments, etc. A lot of the transitions made it feel glitchy when opening so I Just disabled a lot of the. Check how it is and see if I broke anything

mbpowers:
All looks good to me :+1: except for maybe removing the shift scrolling for mouses to scroll horizontally, unless that was intentional. If so why so?

PR #789, opened 2025-11-22
Update keybindings: remap Settings to Mod+S and add Mod+Comma for column management.

MamunAlshawa:
- Added Mod+Comma binding for column management, which was missing.
- Updated Settings binding from Mod+Comma to Mod+S to avoid conflict with the newly added Mod+Comma functionality.

### DIFF
--- a/backend/internal/config/embedded/niri.kdl
+++ b/backend/internal/config/embedded/niri.kdl
@@ -235,7 +235,7 @@ binds {
     Mod+M hotkey-overlay-title="Task Manager" {
         spawn "dms" "ipc" "call" "processlist" "toggle";
     }
-    Mod+Comma hotkey-overlay-title="Settings" {
+    Mod+S hotkey-overlay-title="Settings" {
         spawn "dms" "ipc" "call" "settings" "toggle";
     }
     Mod+Y hotkey-overlay-title="Browse Wallpapers" {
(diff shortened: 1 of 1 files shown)


bbedward:
Changing a default keybind is a bigger deal, while I dont exactly disagree with the change (I'd like to get other opinions) changing a default bind would need to be much more comprehensive. Like changing docs, hyprland, etc.

MamunAlshawa:
The default Niri keybinds already include:

Mod+Comma  { consume-window-into-column; }
Mod+Period { expel-window-from-column; }

(Ref: https://github.com/YaLTeR/niri/blob/main/resources/default-config.kdl)

The missing binding was Mod+Comma, and when this was added, it conflicted with the shortcut that opens the system settings.
To resolve this, the settings shortcut was changed to Mod+S.

This ensures that the newly added binding works correctly while keeping the old settings accessible.
This change only affects Niri and does not impact other window managers.

Please correct me if I have made a mistake.

bbedward:
I mean that, I would also want to see the hyprland default config updated to match the settings bind.

And the [docs](https://github.com/AvengeMedia/DankLinux-Docs) get an update

MamunAlshawa:
I’ve updated the docs and Hyprland config to match the Niri settings bindings and submitted a PR.

PR #795, opened 2025-11-23
Media volume scroll on DankBar widget and media volume OSD

### LINKED ISSUE #595: Playerctl volume controls on media player widget
## Feature Description

A volume slider to control player volume with playerctl instead of just global system volume. I would also like scrolling on the widget on the bar to increase/decrease the volume.

## Use Case

Quick music volume adjustments without switching to my music player. 

## Proposed Solution

It should be quick to implement since DMS just has to issue `playerctl volume [LEVEL][+/-] `. It might be better to only have the scrolling since a GUI slider wouldn't do anything if the player doesn't support MPRIS controls (unless there's a way to check this that I'm not aware of).

LuckShiba:
Closes #595 

This PR changes:
- Scrolling the mouse wheel on the DankBar Media widget will now change the media volume;
- Changes the volume control in the Media Tab of DankDash to be the player volume instead of the device volume;
- Adds an OSD for the current media player volume.

Note that some applications (most notably, [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1674654) doesn't support the MPRIS volume interface. For these cases, I made it disable the volume control in the DankDash and noop on the widget (but still accepting the event, so it won't have a mixed behavior of sometimes changing workspaces and sometimes changing the volume).

### DIFF
--- a/quickshell/Modules/DankDash/MediaPlayerTab.qml
+++ b/quickshell/Modules/DankDash/MediaPlayerTab.qml
@@ -18,7 +18,7 @@ Item {
     property var allPlayers: MprisController.availablePlayers
 
     readonly property bool isRightEdge: (SettingsData.barConfigs[0]?.position ?? SettingsData.Position.Top) === SettingsData.Position.Right
-    property var defaultSink: AudioService.sink
+    readonly property bool volumeAvailable: activePlayer && activePlayer.volumeSupported
 
     // Palette that stays stable across track switches until new colors are ready
     property color dom: Qt.rgba(Theme.surface.r, Theme.surface.g, Theme.surface.b, 1.0)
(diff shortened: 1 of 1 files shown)


quickshell/Common/SettingsData.qml
```diff
@@ -282,6 +282,7 @@ Singleton {
     property bool osdAlwaysShowValue: false
     property int osdPosition: SettingsData.Position.BottomCenter
     property bool osdVolumeEnabled: true
+    property bool osdMediaVolumeEnabled: true
```
bbedward, line 285: This also has to be in the Common/Settings/SettingsSpec.js otherwise it won't be persisted.

bbedward (changes requested):

quickshell/Common/SettingsData.qml
LuckShiba: Fixed!

PR #803, opened 2025-11-23
Update keybindings: remap Settings to Mod+S and add Mod+Comma for column management.

MamunAlshawa:
I’ve updated the docs and Hyprland config to match the Niri settings bindings and submitted a PR.


bbedward:
I'm optting not do accept this, while mod+comma is a deafult niri thing I prefer mod+comma to default to settings since its consistent iwth MacOS and a bunch of other apps.

Its easy enough for users to change it themselves, and soon we're going to have a full settings interface for binds so it'll be even easier then

PR #813, opened 2025-11-25
Add middle-click close and collapse popout

BlackCherryCat:
Collapse all components on close in order to prevent having a different height in the control center when opening it again. It's pretty convenient to close the applications with the middle click

### DIFF
--- a/quickshell/Modules/ControlCenter/ControlCenterPopout.qml
+++ b/quickshell/Modules/ControlCenter/ControlCenterPopout.qml
@@ -92,8 +92,6 @@ DankPopout {
             });
         } else {
             Qt.callLater(() => {
-                collapseAll();
-
                 if (NetworkService.activeService) {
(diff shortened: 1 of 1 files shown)


BlackCherryCat:
Ok, I've noticed that the control center popout behaviour it's fixed, what about the middle click close?

quickshell/Modules/ControlCenter/ControlCenterPopout.qml
```diff
@@ -92,6 +92,8 @@ DankPopout {
             });
         } else {
             Qt.callLater(() => {
+                collapseAll();
```
bbedward: fwiw I think it's better to do this when shouldBeVisible is set, rather than vice versa. Probably prevents any visual glitch since shouldBeVisible is set, before any animation begins.

The rest is good, if you can just revert this part I'll merge the middle click

BlackCherryCat:
All right, thanks. Reverted

PR #826, opened 2025-11-27
Add a widget to display the power menu

pihomeserver:
Create a new widget that will popup the power menu (using the `dms call ipc` command) to avoid going through other widgets

### DIFF
--- a/quickshell/Services/DisplayService.qml
+++ b/quickshell/Services/DisplayService.qml
@@ -750,11 +750,18 @@ Singleton {
         }
     }
 
+    Timer {
+        id: screenChangeRescanTimer
+        interval: 3000
+        repeat: false
+        onTriggered: rescanDevices()
+    }
+
     Connections {
         target: Quickshell
 
         function onScreensChanged() {
-            rescanDevices();
+            screenChangeRescanTimer.restart();
         }
     }
 


quickshell/Modules/DankBar/Widgets/PowerButton.qml
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    MouseArea {
+        anchors.fill: parent
+        acceptedButtons: Qt.LeftButton
+        onClicked: {
+            var process = Qt.createQmlObject(`
+                                            import Quickshell.Io
+                                            Process {
+                                                command: ["dms", "ipc", "call", "powermenu", "toggle"]
```
bbedward: This widget is a fine addition, but it should  not use IPC here. It should invoke the powerMenuModalLoader directly, like control center button does, etc.

pihomeserver: Agree even if I was not able to understand how to call it so I ended with IPC. I will continue to search

bbedward: We have PopoutService, that is a relatively new-ish thing, it might be sensible to add it to that (even though its not a popout) - it makes it easier to access the components in DMSShell.qml without needing to pass a reference all the way downstream

pihomeserver (commented):
Updated to use the popupservice and open the modal centered on the screen

quickshell/Modules/Settings/DankBarTab.qml
```diff
@@ -192,6 +192,12 @@ Item {
             "description": I18n.tr("Quick access to notepad"),
             "icon": "assignment",
             "enabled": true
+        }, {
+            "id": "powerButton",
+            "text": I18n.tr("Power"),
+            "description": I18n.tr("Display the power system menu"),
+            "icon": "power_settings_new",
+            "enabled": true
```
bbedward: We completely re-org'd the settings (for thhe better), but hopefully these conflicts are not too hard to resolve

bbedward (changes requested):

PR #836, opened 2025-11-28
nix: install completion support for dms cli

marcusramberg:
Use the built in completion command to install shell completion for dms on nixos

Also slightly tidied buildDmsPkgs by using inherit


LuckShiba:
Thanks!

LuckShiba (approved):

PR #815, opened 2025-11-26
feat: Add browser picker modal for URL handling

devnullvoid:
## Summary
Add a **generic, extensible application picker** that allows users to select which application to open URLs and files with, similar to the app launcher interface. This feature is designed to be **highly reusable** and can handle any MIME type or application category, similar to [Junction](https://github.com/sonnyp/Junction).

The browser picker is now just one specialized use case of the generic `AppPickerModal` component - the architecture makes it trivial to create pickers for PDFs, images, text files, or any other file type.

**✨ NEW: Generic File Opening**
- Automatically detects MIME types from file extensions
- Maps MIME types to appropriate application categories
- Handles `file://` URIs and local file paths
- Shows relevant applications based on file type (TextEditor for .txt, Graphics for .png, etc.)
- Per-file-type usage tracking
- Works with `xdg-open` through MIME type associations

### UI/UX
- **Search functionality** - Filter applications by name with DankTextField
- **List/Grid view modes** - Toggle between list and grid layouts
- **View mode persistence** - Remembers user's preferred view (list/grid)
- **Clean layout** - Matches AppLauncher design (520x500)
- **Target display** - Shows URL/file path at bottom with proper truncation
- **Keyboard navigation** - Full keyboard support with auto-scrolling
- **Grid column sync** - Browser picker grid uses same column setting as app launcher

### Functionality
- **Usage tracking** - Sorts applications by most recently used (per picker instance)
- **Application filtering** - Search to quickly find specific applications
- **Persistent preferences** - Stores view mode and usage history in SettingsData
- **Smart sorting** - Most used applications appear first, with alphabetical fallback
- **CLI integration** - `dms open` command communicates with running DMS instance
- **Extensible architecture** - Easy to create specialized pickers for different file types

### Generic AppPickerModal Component

The core `AppPickerModal` component (~450 lines) is completely generic and reusable:

```qml
AppPickerModal {
    title: "Select Application"
    categoryFilter: ["Office", "Viewer"]  // Filter by any categories
    targetData: "document.pdf"             // What we're opening
    targetDataLabel: "PDF"                 // Optional label
    usageHistoryKey: "pdfPickerHistory"   // Separate usage tracking

    onApplicationSelected: (app, target) => {
        // Handle launch however you want
    }
}
```

### Browser Picker as Example

The browser picker is now just a **46-line wrapper** demonstrating how to create specialized pickers:

```qml
AppPickerModal {
    title: I18n.tr("Open with...")
[...]

PR #848, opened 2025-11-29
core: replace all use of interface{} with any in the go code

marcusramberg:
Motivation: any is an alias for interface{} and recommended for new code.

imo it makes the code easier to read.


bbedward:
Old habits I guess, but I'm not inclined to merge this right now since I have a bunch of local work on the cups stuff, and this will create many conflicts.

marcusramberg:
If you give me a ping, I'd be happy to rebase at a more suitable time.

bbedward:
Also prefer to wait for #815  - or at least not touch any of the go files that one is changing. CUPS I'll have done within the hour

PR #856, opened 2025-11-30
refactor(layout): implement geometric centering for widget group

gbPagano:
Previously, the layout logic attempted to anchor the specific "middle widget" to the center of the screen. This caused visual asymmetry when the surrounding widgets differed significantly in size, resulting in a visually unbalanced bar.

I replaced the old index-based anchoring with a geometric-center approach. This guarantees proper centering regardless of individual widget widths.

### Visual Comparison

**Before:**

**After:**

**Before:**

**After:**

| Before | After |
| :---: | :---: |
|  |  |

### DIFF
--- a/quickshell/Common/SettingsData.qml
+++ b/quickshell/Common/SettingsData.qml
@@ -175,6 +175,7 @@ Singleton {
     property bool keyboardLayoutNameCompactMode: false
     property bool runningAppsCurrentWorkspace: false
     property bool runningAppsGroupByApp: false
+    property string centeringMode: "index"
     property string clockDateFormat: ""
     property string lockDateFormat: ""
     property int mediaSize: 1
@@ -227,6 +228,7 @@ Singleton {
     onNotepadFontFamilyChanged: saveSettings()
     onNotepadFontSizeChanged: saveSettings()
     onNotepadShowLineNumbersChanged: saveSettings()
+    // onCenteringModeChanged: saveSettings()
     onNotepadTransparencyOverrideChanged: {
         if (notepadTransparencyOverride > 0) {
             notepadLastCustomTransparency = notepadTransparencyOverride;
--- a/quickshell/Common/settings/SettingsSpec.js
+++ b/quickshell/Common/settings/SettingsSpec.js
@@ -90,6 +90,7 @@ var SPEC = {
     keyboardLayoutNameCompactMode: { def: false },
     runningAppsCurrentWorkspace: { def: false },
     runningAppsGroupByApp: { def: false },
+    centeringMode: { def: "index" },
     clockDateFormat: { def: "" },
     lockDateFormat: { def: "" },
     mediaSize: { def: 1 },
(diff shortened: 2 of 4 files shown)


bbedward:
My issue with this, as-is at least is that , in my setup, I have media/clock/weather in top center - and when media starts playing/reveals itself I prefer it to just appear next to the clock, without shifting anything. With this patch, it causes the clock to shift right when m edia starts playing

gbPagano:
That makes sense, but I believe the tradeoff here is justified. The shifting only happens during significant state changes (opening/closing the player), so it doesn't feel erratic.

By accepting this momentary shift, we gain a perfectly centered layout for the vast majority of the time the system is in use, which I think looks much more polished. This aligns with how most modern shells handle dynamic center widgets.

Maybe we could make this behavior configurable (e.g., centerStrategy: 'geometric' | 'fixed_middle')? However, I strongly believe 'geometric' should be the default for better aesthetics.

bbedward:
> That makes sense, but I believe the tradeoff here is justified. The shifting only happens during significant state changes (opening/closing the player), so it doesn't feel erratic.
> 
> By accepting this momentary shift, we gain a perfectly centered layout for the vast majority of the time the system is in use, which I think looks much more polished. This aligns with how most modern shells handle dynamic center widgets.
> 
> Maybe we could make this behavior configurable (e.g., centerStrategy: 'geometric' | 'fixed_middle')? However, I strongly believe 'geometric' should be the default for better aesthetics.

I don't disagree but I really don't want to change at least the default behavior of media not shifting the clock when its revealed. That's sort of a special case widget - along with maybe the privacy indicator, that just reveal themselves conditionally.

Eitehr way a centerStrategy globally is ok, or just a special case for the media widget specifically.

gbPagano:
To address this while keeping the centering logic, I've pushed a new commit.

Instead of unloading the Media widget when no player is active (which caused the resize/shift), I modified it to handle visibility via opacity and the enabled property.

The widget now remains in the scene graph (invisible and unclickable) when idle. This prevents the "re-center" calculation from triggering abruptly when media starts/stops. The clock and other widgets remain stable, and the media player simply fades in without disrupting the layout.

Example:

bbedward:
It's still not the behavior that I want it to be at least. With my 3 widget group it just shifts the clock off center all the time. I want my clock anchored to the center at all times, as it is before any changes. For vertical bars, I think it may matter much less. I am not ok with changing the behavior of the current default config, but am ok with changing behavior optionally, not by default.

pihomeserver:
What about a section option : default is to keep current behavior and once activated then the new "layout"applies ?
The option could be displayed via a 3 dots button on the right of the section title

bbedward:
That seems good to me, an alternative is the ability to anchor a specific widget to the center but that's more complicated and  excessive probably

gbPagano:
I added this option to set the centralization method. Here is how it looks in the settings menu:

bbedward (approved):

PR #867, opened 2025-12-01
feat: add maxSystemVolume and maxMediaVolume, DankSlider changes

mbpowers:
add vertical version of DankSlider
add reference bar to DankSlider
add tooltipPlacement to DankSlider
replace some custom volume sliders with DankSlider


mbpowers:
I'm pretty sure I got every volume slider.

I can also add in mic input volume max but wasn't sure if people would actually need that above the default 100%.

Also the tooltip in the `MediaDropdownOverlay` was clipping on the right for large values and I couldn't figure that out.

purian23:
Feel free to rework this PR for updated compatibility or close to reopen in the future.

PR #886, opened 2025-12-02
WIP: parallax scrolling wallpaper

PandorasFox:
needs integration with: lockscreen, wallpaper transition shaders, and a performance/optimization pass

a bit choppier than my lazy WpViewport solution, but I'm using a 75MiB image that's approx 7500x11000 with my 3440x1440 display - right now I'm testing with `QT_IMAGEIO_MAXALLOC=0 qs -p .` just to be able to load my gigantic wallpaper, so the perf issues I see might be solved by optimizing the image itself a bit, but I'm still doing a pass over the anchoring/positioning, rects & clipping, etc to tighten that down a bit first.

vibe-coded most of the first-pass impl here; rest of the work will be manual bits

current thoughts:
- when finishing a scroll, we could try and copy over the display-size crop over to the active wallpaper image slot for the shaders to use
- similar stuff for loading/cropping NextWallpaper 
- (wallpaper picker widget would benefit from enabling mipmaps on the thumbnailed wallpapers since they're downscaled?)
- should just be able to export the scrolling percentage (or raw x/y offsets) to the lockscreen/blurred wallpaper widgets?


purian23:
Feel free to update to finish the PR, and for updated compatibility or close to reopen in the future.

PandorasFox:
yeah, I've been meaning to come back to this with a proper shader implementation, I've just been a bit busy the past few months.

PandorasFox:
Will be reopening this under a fresh PR, since I chucked out this approach for a GSL fragment shader.

It works much better now, with a note that `QSG_USE_SIMPLE_ANIMATION_DRIVER=1` is required for the animation loop to not be locked to 60Hz regardless of display refresh rate (required so that this all looks smooth on my 240hz panel :) )

PandorasFox:
#1802

PR #923, opened 2025-12-06
evdev: improve capslock detection for no led device

yuxqiu:
The current Caps Lock detection does not work correctly on my Sofle keyboard: the state instantly flips from true back to false.  

After checking the logs, the cause is clear:

```go
if event.Type == evKeyType && event.Code == keyCapslockKey && event.Value == keyStateOn {
    time.Sleep(50 * time.Millisecond)
    m.readAndUpdateCapsLockState(deviceIndex) // triggered here
} else if event.Type == evLedType && event.Code == ledCapslockKey {
    capsLockState := event.Value == keyStateOn
    m.updateCapsLockStateDirect(capsLockState)
}
```

and inside the called function:

```go
ledStates, err := device.State(evLedType)
...
capsLockState := ledStates[ledCapslockKey] // root cause
```

My keyboard has no physical Caps Lock LED, so `device.State(EV_LED)` returns an empty map.  
Accessing it yields `false`, which forces the state back to off every time.

This PR adds a simple check: if `ledStates` is completely empty, we skip the update entirely instead of treating Caps Lock as off.

The change has minimal impact on other devices:
- When `ledStates` is non-empty (normal case), behaviour is unchanged.
- When it is empty, we now avoid the incorrect forced-off state that some keyboards currently suffer from.

---

Another alternative to the current fix would be to fall back to using `event.Value` directly whenever the LED map is detected to be empty.

### DIFF
--- a/core/internal/server/evdev/manager.go
+++ b/core/internal/server/evdev/manager.go
@@ -306,14 +306,14 @@ func (m *Manager) readAndUpdateCapsLockState(deviceIndex int) {
 		return
 	}
 
-    if len(ledStates) == 0 {
-        log.Debug("No LED state available (empty map)")
+	if len(ledStates) == 0 {
+		log.Debug("No LED state available (empty map)")
 
-        // This means the device either:
-        // - doesn't support LED reporting at all, or
-        // - the kernel returned an empty state
-        return
-    }
+		// This means the device either:
+		// - doesn't support LED reporting at all, or
+		// - the kernel returned an empty state
+		return
+	}
 
 	capsLockState := ledStates[ledCapslockKey]
 	m.updateCapsLockStateDirect(capsLockState)


yuxqiu:
@bbedward This is ready for review. Feel free to take a look when you have time. Thanks!

bbedward:
> @bbedward This is ready for review. Feel free to take a look when you have time. Thanks!

It's failing the pipeline still

yuxqiu:
Oh that is because of an irrelevant commit before mine. Probably I should rebase it.

yuxqiu:
> > @bbedward This is ready for review. Feel free to take a look when you have time. Thanks!
> 
> It's failing the pipeline still

Done.

bbedward:
fwiw I didnt see a better way to monitor the state than LED, the other event normally just gives us if it was pressed or not, but that doesnt reflect the "state" of capslock. There might be a better solution than listening to the LEDs, though.

PR #930, opened 2025-12-06
fix: handle ipc arguments

marcusramberg:
If a command has function arguments, insert them as placeholders on autocomplete. Also stops it from incorrectly autocompleting ipc calls that has no arguments.


marcusramberg:
Accidentally pushed older version here. ci will still fail due to #935 - can rebase after that's been applied.

PR #967, opened 2025-12-11
nix: add QML dependencies to dms-shell package

LuckShiba:
Removes `enableSystemSound` option in the modules and adds `qtmultimedia`, `kirigami` and `sonnet` QML import paths into `dms-shell` package instead. qtmultimedia, in the way it was before, wouldn't work without adding the import path.


LuckShiba:
Rebased on master.

PR #981, opened 2025-12-11
Add lock screen layout settings

pihomeserver:
Add a layout setting block for the lock screen to make it more customisable or minimalist
- Show/hide system time
- Show/hide system date
- Show/hide profile icon
- Show/hide system icons and weather
- Show/hide password field until a key is pressed

If this PR is ok I will push another one with more settings (font color, size, opacity, etc.)

### DIFF
--- a/quickshell/Modules/Settings/LockScreenTab.qml
+++ b/quickshell/Modules/Settings/LockScreenTab.qml
@@ -26,43 +26,38 @@ Item {
                 title: I18n.tr("Lock Screen layout")
 
                 SettingsToggleRow {
-                    text: I18n.tr("Show Power Actions")
-                    description: I18n.tr("Show power, restart, and logout buttons on the lock screen")
+                    text: I18n.tr("Show Power Actions", "Enable power action icon on the lock screen window")
                     checked: SettingsData.lockScreenShowPowerActions
                     onToggled: checked => SettingsData.set("lockScreenShowPowerActions", checked)
                 }
 
                 SettingsToggleRow {
-                    text: I18n.tr("Show System icons")
-                    description: I18n.tr("Show system icons on the lock screen")
+                    text: I18n.tr("Show System Icons", "Enable system status icons on the lock screen window")
                     checked: SettingsData.lockScreenShowSystemIcons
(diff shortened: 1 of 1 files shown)


quickshell/Modules/Settings/LockScreenTab.qml
```diff
@@ -32,6 +32,47 @@ Item {
                     onToggled: checked => SettingsData.set("lockScreenShowPowerActions", checked)
                 }
 
+                SettingsToggleRow {
+                    text: I18n.tr("Show System icons")
```
bbedward: The casing of these should be consistent, `Show System Icons` or `Show system icons`  (do as I say not as I do, I know there's not a consistent pattern right now). I'd probably up for `Show System Icons`

It's important just because these become localization keys and we can't change them later without breaking translations.

Also add context ideally (again do as I say not as I do...localization started as an afterthought). That is just a second parameter of I18n.tr, like `I18n.tr("Show System Icons", "Enable system status icons on lock screen settings window")`

quickshell/Modules/Settings/LockScreenTab.qml
```diff
@@ -32,6 +32,47 @@ Item {
...
+                    description: I18n.tr("Show system icons on the lock screen")
+                    checked: SettingsData.lockScreenShowSystemIcons
+                    onToggled: checked => SettingsData.set("lockScreenShowSystemIcons", checked)
+                }
+
+                SettingsToggleRow {
+                    text: I18n.tr("Show Time")
+                    description: I18n.tr("Show system time on the lock screen")
```
bbedward: May prefer to get rid of these descriptions entirely since the toggles may provide enough context, for most of them (again, just to reduce i18n burdens)

quickshell/Modules/Settings/LockScreenTab.qml
pihomeserver: Should be ok now. I have also updated the existing parameter

quickshell/Modules/Settings/LockScreenTab.qml
pihomeserver: I have removed all of them except for the password field to ensure that the hidden behavior is understood

bbedward (approved):

PR #1021, opened 2025-12-14
fix: Add monitor configuration for Hyprland greeter

arfan21:
in my case, the greetd resolution not 1920x1080

### DIFF
--- a/quickshell/Modules/Greetd/assets/dms-greeter
+++ b/quickshell/Modules/Greetd/assets/dms-greeter
@@ -174,6 +174,8 @@ misc {
     disable_hyprland_logo = true
 }
 
+monitor = , preferred, auto, 1
+
 exec = sh -c "$QS_CMD; hyprctl dispatch exit"
 HYPRLAND_EOF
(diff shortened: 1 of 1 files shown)


bbedward:
Wouldn't this be a better default? `monitor = , preferred,auto,auto` setting scaling to 1 might make it less-accessibility friendly out of the box.

arfan21:
image

I'm setting up a multi-monitor configuration, and I've noticed that scaling levels differ between displays when auto (for example, my 1080p laptop defaults to 1.5).
The default 1.5 on the 1080p laptop screen makes text and UI too big for my preference.

bbedward:
The question isn't about that really, but more about accessibility. Hyprland might default to a larger scale so that text and things are more legible by default. It just seems like it may be better to rely on Hyprland's default  behavior since dms-greeter users [can override it anyway](https://danklinux.com/docs/dankgreeter/configuration#hyprland)

arfan21:
Ah okay, then I'll use the override method.

PR #1022, opened 2025-12-14
Fixed having to click twice to exit out of Spotlight/Cliphist/Powermenu

tsxkasa:
fix the need of having to click the background twice to close those modals.
not sure if this is intentional behavior


bbedward:
We may want to do this for other HyprlandFocusGrab uses, as long as it doesnt break keyboard focus of the primary surface

PR #1030, opened 2025-12-14
dankmodal: removed backgroundWindow to fix clicking twice

tsxkasa:
removed 'backgroundWindow' but combined it with 'contentWindow',
adding backgroundWindow to HyprlandFocusGrab's window kinda broke keyboard focus on the content window (sorry)

### DIFF
--- a/quickshell/Modals/Clipboard/ClipboardHistoryModal.qml
+++ b/quickshell/Modals/Clipboard/ClipboardHistoryModal.qml
@@ -13,7 +13,7 @@ DankModal {
     layerNamespace: "dms:clipboard"
 
     HyprlandFocusGrab {
-        windows: [clipboardHistoryModal.contentWindow]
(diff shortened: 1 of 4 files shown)


bbedward:
This is not ok, unless it's specific to hyprland. The point of two windows is to workaround for QML surface damage - damaging the entire window. Otherwise the GPU will redraw the entire screen surface when something changes.

The only reason hyprland uses a focus grab workaround is because it misbehaves with exclusive keyboard focus for some reason. We have to figure something else out but we can't abandon the two window solution (unless we only did that on hyprland)

bbedward:
Im ok to revert to the less power friendly behavior of 1 window on hyprland only (but it has to be hyprland only)

tsxkasa:
Thanks for your response, so it is okay if I make this change specific to Hyprland?

bbedward:
> Thanks for your response, so it is okay if I make this change specific to Hyprland?

Yea hyprland is fine, we just can't do it across all compositors.

bbedward (approved):

PR #1078, opened 2025-12-17
Add media control bindings using playerctl

adarsh-67r:
* Added `XF86AudioPlay`, `XF86AudioNext`, and `XF86AudioPrev` key bindings to control media playback (play/pause, next, previous) via `playerctl`.

### DIFF
--- a/core/internal/config/embedded/niri-binds.kdl
+++ b/core/internal/config/embedded/niri-binds.kdl
@@ -45,6 +45,15 @@ binds {
     XF86AudioLowerVolume allow-when-locked=true {
         spawn "dms" "ipc" "call" "audio" "decrement" "3";
     }
+    XF86AudioPlay allow-when-locked=true { 
+        spawn "playerctl" "play-pause"; 
+    }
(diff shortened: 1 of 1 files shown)


core/internal/config/embedded/niri-binds.kdl
```diff
@@ -45,6 +45,15 @@ binds {
...
     }
+    XF86AudioPlay allow-when-locked=true { 
+        spawn "playerctl" "play-pause"; 
+    }
+    XF86AudioNext allow-when-locked=true { 
+        spawn "playerctl" "next"; 
+    }
+    XF86AudioPrev allow-when-locked=true { 
+        spawn "playerctl" "previous"; 
+    }
```
bbedward, line 56: Shouldn't we use DMS IPC for this? Since playerctl is using mpris anyway. Takes away a needed dependency.

Also good to add to hyprland stock config I think

Shorakie:
call executes a command. we can omit it here.
fun fact we can do
`dms ipc call call... mpris playPause`

bbedward:
closed for #1240

PR #1081, opened 2025-12-17
Enable printer device and driver search

misiasty3:
It can often be impossible to select your desired printer driver when there is a lot of them installed, for example when using [Gutenprint](https://gimp-print.sourceforge.io/)

https://github.com/user-attachments/assets/04075c53-9f82-4040-a6b5-cbddedc1b399

https://github.com/user-attachments/assets/ec5c8018-fc0c-4db2-91fe-9585abab8f77

Draft because:
-  Performance could be better, as shown in the video. Every keystroke generates a small, noticeable lag. 
- It would make sense to hide the search field when there is only a small number of devices/drivers


bbedward:
The performance is probably a fail in the DankDropdown implementation, guessing it should use a ScriptModel and is probably rebuilding the delegate on every search query (unrelated to your PR, unless you wanna try to tackle it).

Feel free to take this out of draft then I'll review it

bbedward:
Done in c9331b73387eb434ad5052b41944b5ebd13dbb24

Issue #1093, opened 2025-12-19
Add IPC call to dismiss all notification popups


marcinjahn:
## Feature Description
Notification popups cannot be controlled with keyboard right now. When a popup appears, you can dismiss it only with a mouse click.

## Use Case

IPC command for dismissing notification popups would allow to bind it to any keyboard shortcut.

## Compositor

Is this feature specific to one compositor?

- [x] All compositors

## Proposed Solution

Add a new IPC command.

ewtodd:
I added an IPC command ( `dms ipc call notifications clearAll` ) for dismissing all notifications in #1082, which also works to dismiss popups. This addresses the title, but are you also suggesting a command for dismissing a single active popup?

marcinjahn:
By “clearAll”, so you mean that the Notification Center will be empty after your command’s invocation? I think it’d be useful to have a command that just closes the current popup(s) and places them in the Notification Center.

ewtodd:
Understood. See #1100, which adds `dms ipc call notifications dismissAllPopups`, which, as the name suggests, will send all active popups to the notification center (but not clear them). Let me know if this is what you had in mind!

PR #1100, opened 2025-12-19
notifications: add modal IPC command to dismiss all popups

ewtodd:
Add function to notifications IPC modal for dismissing all popups to the notification center. I renamed the existing clearAllPopups() function to dismissAllPopups(), since the clearAllNotifications function actually deletes the notifications but the clearAllPopups function simply sends them to the notification center. This addresses #1093.


bbedward:
I'm going to hold this one because I'm traveling and don't wanna forget to add it to the documentation but it looks fine

PR #1140, opened 2025-12-24
Add getDoNotDisturb to IPC

aaronjamt:
This PR adds a new function to the IPC interface for fetching current Do Not Disturb status.


aaronjamt:
Just realized there was already an undocumented interface for toggling the state, all it actually needs is a way to fetch the current state. PR updated.

PR #1149, opened 2025-12-24
Update translations

aaronjamt:
I noticed that the translation files are a bit outdated, this PR updates them.

### DIFF
--- a/quickshell/translations/en.json
+++ b/quickshell/translations/en.json
@@ -122,7 +122,7 @@
   {
     "term": "180°",
     "context": "180°",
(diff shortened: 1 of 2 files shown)


bbedward:
I generate these with a script periodically, not manually.

aaronjamt:
If you're referring to `quickshell/translations/extract_translations.py`, that's what I used to make this PR.

bbedward:
I run the sync one which extracts, syncs with poeditor API, and pulls in all the languages. It requires the poeditor API key , sometimes I lag because I don't always wanna send new terms to translators if a new feature is going to change still.

aaronjamt:
Should I remove the translation file updates from my other PRs then? I only added the new terms in those.

bbedward:
> Should I remove the translation file updates from my other PRs then? I only added the new terms in those.

It doesn't matter too much, besides conflicts that will come up. I used to have it on a pre-commit hook but because it's a little slow I just run it manually. Since it ships the new terms to translators I just try to make it so there's a decent sized batch for them to go over.

PR #1174, opened 2025-12-27
Notification: Right-click to toggle Do Not Disturb

xxyangyoulin:


quickshell/Modules/DankBar/Widgets/NotificationCenterButton.qml
```diff
@@ -33,4 +33,13 @@ BasePill {
             }
         }
     }
+
+    MouseArea {
```
bbedward: BasePill provides a signal, onRightClicked, it would be better to just connect to that instead of having a separate mouse handler.

bbedward (changes requested):

quickshell/Modules/DankBar/Widgets/NotificationCenterButton.qml
xxyangyoulin: Thanks for the reply, I updated

bbedward (approved):

PR #1179, opened 2025-12-27
feat: Added reverse workspace scrolling

vha:
Added an option to reverse workspace switch direction when scrolling over the bar and workspace plugin.
image

### DIFF
--- a/quickshell/translations/settings_search_index.json
+++ b/quickshell/translations/settings_search_index.json
@@ -803,16 +803,15 @@
       "contain",
       "desktop",
       "desktops",
-      "scroll", 
-      "scrolling", 
-      "reverse", 
+      "scroll",
+      "scrolling",
+      "reverse",
       "direction",
       "windows",
       "workspace",
       "workspaces"
     ],
-    "description": "Display only workspaces that contain windows",
-    "conditionKey": "isNiri"
+    "description": "Display only workspaces that contain windows"
   },
   {
     "section": "showWorkspaceApps",


quickshell/Modules/DankBar/Widgets/WorkspaceSwitcher.qml
```diff
@@ -675,7 +675,7 @@ Item {
 
             const delta = wheel.angleDelta.y;
             const isMouseWheel = Math.abs(delta) >= 120 && (Math.abs(delta) % 120) === 0;
-            const direction = delta < 0 ? 1 : -1;
+            const direction = delta * -SettingsData.reverseScrolling < 0 ? 1 : -1;
```
LoayGhreeb, line 678: This broke the normal scroll direction. When `reverseScrolling` is `false`, `delta * -false` evaluates to `delta * 0 = 0`, which makes the direction always `-1` regardless of scroll input.

I think it should be like:
```suggestion
            const direction = delta * (SettingsData.reverseScrolling ? -1 : 1) < 0 ? 1 : -1;
```

bbedward: Can you or someone open a PR

vha: #1182

LoayGhreeb: Thanks for the quick fix.

Issue #1187, opened 2025-12-28
Can't build on NixOS after commit removed DGop Input.


HayyaSylvy:
## Compositor

- [* ] niri

## Distribution

NixOS

## Description
After updating my flake to match the latest commit, a reversion caused by the removal of the dgop input from the flake results on a build failure with "An option declaration for `home-manager.users.(myuser).programs.dank-material-shell.dgop.package has type `lambda' rather than an attribute set." as an error message. i haven't changed a line of my configuration other than updating my flake.

```

Postboote1:
should be fixed by #1188.
only waiting for merge

HayyaSylvy:
Yep, I can say the issue has been fixed after this commit :D

PR #1188, opened 2025-12-28
fixed lambda issue on nixos

Postboote1:
Because of missing brackets the error:
```
       error: An option declaration for `home-manager.users.matthias.programs.dank-material-shell.dgop.package' has type
       `lambda' rather than an attribute set.
```
occured. Fixed by adding missing brackets in options.nix

### DIFF
--- a/distro/nix/options.nix
+++ b/distro/nix/options.nix
@@ -37,7 +37,7 @@ in
     };
 
     dgop = {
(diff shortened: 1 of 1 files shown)


Postboote1:
should fix #1187

marcusramberg (approved):

PR #1189, opened 2025-12-28
fix: new dgop package option is missing function call

marcusramberg:
has type 'lambda' rather than attribute set.


Postboote1:
#1188

PR #1194, opened 2025-12-28
Further tweaks for Zen theming on DMS

johngalt:
A few tweaks to the CSS that have been brought up (mainly the workspaces container). Also will keep this PR in draft until I figure out how to handle the whole `zen` vs `zen-browser` binary matching issue with matugen.go .

### DIFF
--- a/core/internal/config/deployer_test.go
+++ b/core/internal/config/deployer_test.go
@@ -462,7 +462,7 @@ func TestHyprlandConfigStructure(t *testing.T) {
 	assert.Contains(t, HyprlandConfig, "# KEYBINDINGS")
 	assert.Contains(t, HyprlandConfig, "bind = $mod, T, exec, {{TERMINAL_COMMAND}}")
 	assert.Contains(t, HyprlandConfig, "bind = $mod, space, exec, dms ipc call spotlight toggle")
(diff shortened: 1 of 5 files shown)


johngalt:
So for discussion: 

DMS looks for the `zen` binary name to determine whether the theme should be generated or not. NixOS and some other distros use this as the default binary name. Arch users who install the browser from AUR end up with a `zen-browser` binary. In this instance, the theme isn't generated because the binary name is different than what is expected by DMS.

I don't know any Go to figure out how to tweak `appendConfig()` in `core/internal/matugen/matugen.go` to check either `zen` OR `zen-browser` with the `utils.CommandExists()` function without breaking anything.

An alternative would be to add a duplicate appendConfig line to match either binary names, but that seems hacky?
```go
	if !opts.ShouldSkipTemplate("zenbrowser") {
		appendConfig(opts, cfgFile, "zen", "zenbrowser.toml")
		appendConfig(opts, cfgFile, "zen-browser", "zenbrowser.toml")
	}
```

Open to suggestions

johngalt:
Commit https://github.com/AvengeMedia/DankMaterialShell/commit/cffee0fae6471133b5a399fbc89c9b432f68b85d fixes the `zen` vs `zen-browser` issue. This PR will just fix the workspaces color in the userChrome.css template.

PR #1211, opened 2025-12-29
feat: implement smart compositor entry point (start-hyprland vs Hyprland)

mesteryui:
Description

This PR introduces a version-aware check to optimize the Hyprland startup process within DMS scripts, following the latest recommendations from the Hyprland development team.
What is start-hyprland?

Starting from recent versions ([v0.53.0+](https://hypr.land/news/update53/) ecosystem), the developers have introduced start-hyprland. This is a wrapper script designed to handle the environment initialization more safely than calling the Hyprland binary directly.
Key improvements in this PR:

    Safety & Environment: start-hyprland ensures that essential environment variables and session requirements are correctly set up before the compositor launches.

    Warning Mitigation: Running the Hyprland binary directly in newer versions often triggers warnings regarding “unsafe” or “improper” startup methods. This wrapper eliminates those warnings.

    Production Readiness: While the Hyprland binary is not deprecated, the official stance is that start-hyprland is the recommended entry point for production environments to ensure a stable Wayland session.

    Backward Compatibility: The script includes a version check (MINIMUM_VERSION="0.53.0") to ensure that users on older tags can still boot using the legacy binary method without breakage.

Technical Logic:

The script now dynamically detects the Hyprland version:

    If the version is ≥ 0.53.0, it utilizes start-hyprland -- --config ....

    If the version is older, it falls back to the standard Hyprland -c ... command.

This change aligns DMS with the current “best practices” of the Hyprland ecosystem, providing a cleaner and more robust user experience.
Reference:

    [Hyprland News - Update v0.53.0+](https://hypr.land/news/update53/)

### DIFF
--- a/quickshell/Modules/Greetd/assets/greet-hyprland.sh
+++ b/quickshell/Modules/Greetd/assets/greet-hyprland.sh
@@ -13,4 +13,3 @@ if [ "$(printf '%s\n%s' "$MINIMUM_VERSION" "$CURRENT_VERSION" | sort -V | head -
 else
     exec Hyprland -c /etc/greetd/dms-hypr.conf
 fi
-


alsaiduq-lab:
Sorry to inject here but this commit broke my greeter module I was using from here on NixOS. I am currently on a reverted lock file but from what I understand

```
CURRENT_VERSION=$(hyprctl version | grep "Tag:" | awk '{print $2}' | tr -d 'v,')
	MINIMUM_VERSION="0.53.0"
	if [ "$(printf '%s\n%s' "$MINIMUM_VERSION" "$CURRENT_VERSION" | sort -V | head -n1)" = "$MINIMUM_VERSION" ]; then
	    exec start-hyprland -- --config "$COMPOSITOR_CONFIG"
	else
	    exec Hyprland -c "$COMPOSITOR_CONFIG"
	fi
        ;;
```

How does part work if hyprland hasn't started yet?

mesteryui:
I don't know I only now that when I load it works at least for me and if that was merged I think doesn't happen anything, I started from ArchLinux using a package that installs the greet module and everything needed from git repository and it doesn't break, I fix it to use the ```hyprland --version``` command it gives the same information that hyprctl version but without depending of IPC or something like that

alsaiduq-lab:
? It's checking for version through hyprctl not hyprland --version

mesteryui:
Now it uses hyprland --version they accept the pull request for that

sweenu:
I'm on master, and still get the warning that hyprland is not starting with `start-hyprland`.
And this issue might also be related: https://github.com/AvengeMedia/DankMaterialShell/issues/1225

PR #1220, opened 2025-12-31
Improving the logic for start-hyprland wrapper use

mesteryui:
This includes a change to use hyprland --version and solves any problem of using hyprctl because hyprctl normally serve to use Hyprland IPC from command line this uses the hyprland command for that and after try it in a tty it works without any problem to acquire the information so it's more stable in that sense


bbedward:
It would probably be easier and more reliable to just check for the existence of start-hyprland in the path tbh

PR #1230, opened 2026-01-01
feat: Added a latest news panel to the system updates widget

vha:
This PR adds a news ticker with the latest news item from a news feed to the System Updates widget. On click, it expands to show all available news and their descriptions.

I made it to check Arch news before updating the system, but you can configure the feed to show something else. By default it works on Arch, Manjaro, CachyOS, EndeavourOS and Fedora as an honorable guest. Didn't really know what feed to set on Fedora, so I went with https://fedoramagazine.org/

Since Arch's feed has a known and stable structure, I'm using curl + regex instead of any xml parser to avoid extra dependencies. 

The custom feed option is pretty limited. You need to give it a regex with 4 matching groups similar to arch's structure so it fits the layout.

https://github.com/user-attachments/assets/aa0f69f9-d98c-4a3b-8765-cb53a07e8aec

### DIFF
--- a/quickshell/Modules/SystemUpdatePopout.qml
+++ b/quickshell/Modules/SystemUpdatePopout.qml
@@ -23,6 +23,7 @@ DankPopout {
     shouldBeVisible: false
 
     onBackgroundClicked: close()
+
     onShouldBeVisibleChanged: {
         if (shouldBeVisible) {
             if (SystemUpdateService.updateCount === 0 && !SystemUpdateService.isChecking) {
@@ -34,6 +35,7 @@ DankPopout {
     content: Component {
         Rectangle {
             id: updaterPanel
+
             color: "transparent"
             radius: Theme.cornerRadius
             antialiasing: true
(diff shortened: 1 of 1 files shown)


purian23:
Hi there, thank you for your interest to contribute to DMS. I think this is a cool feature, however we don't feel it belongs in the core code by default. We would request to convert this into a plugin where users are able to modularly install and use the feature in the future. 

https://danklinux.com/docs/dankmaterialshell/plugin-development

vha:
> Hi there, thank you for your interest to contribute to DMS. I think this is a cool feature, however we don't feel it belongs in the core code by default. We would request to convert this into a plugin where users are able to modularly install and use the feature in the future.
> 
> https://danklinux.com/docs/dankmaterialshell/plugin-development

Can I inject my qml inside the system updater widget from a plugin? I can only see ways to connect to the control center. 
The idea was to check the news before updating; adding the newsfeed somewhere else defeats the point, so I guess I'm abandoning the feature.

purian23:
What might be even cooler is converting it to a desktop plugin widget. 
https://danklinux.com/docs/dankmaterialshell/plugin-development#desktop-plugins

PR #1231, opened 2026-01-01
feat: Persistent Launcher instances and Async UI updates

pcortellezzi:
## Overview
This PR refactors how Launcher plugins are handled to support state persistence and asynchronous UI updates. Previously, Launcher plugins were instantiated and destroyed for every single query, preventing them from maintaining internal state or performing async tasks.

## Concrete Example: The Qalculate Plugin
This refactoring was driven by the needs of advanced plugins like [Qalculate](https://github.com/pcortellezzi/dms_plugins/tree/main/qalculate).
- **Persistence**: The plugin keeps a background `qalc` process alive for instant performance. Without this PR, the process would be killed and restarted at every keystroke.
- **Asynchronicity**: When a calculation is sent, the plugin can now immediately return a "Calculating..." item and then signal the UI to refresh with the final result once the background process responds, without freezing or blocking the launcher.

## Key Changes

1.  **Persistent Launcher Instances**: 
    - Launcher plugins are now instantiated upon loading and stored in `PluginService.pluginInstances`, similar to `Daemons`.
    - This allows plugins to maintain a background state (e.g., a long-running process or an active connection).

2.  **Asynchronous UI Updates**:
    - Introduced a `requestLauncherUpdate(pluginId)` signal in `PluginService`.
    - `AppLauncher` now listens to this signal to trigger a refresh. Plugins can now update the items list at any time, not just during the initial query.

3.  **AppSearchService Optimization**:
    - `AppSearchService` now reuses the persistent instance from `PluginService`. 
    - It maintains backward compatibility by falling back to the old create/destroy behavior if no persistent instance is found.

4.  **Dependency Injection**:
    - The `PluginService` singleton is injected into plugins during creation (`pluginService: root`), enabling easier access to shared services and data.

## Why this is needed
This architecture is essential for any plugin that interacts with external binaries, APIs, or performs heavy computations. It makes the launcher capable of handling non-instant results while remaining fluid for the user.

### DIFF
--- a/quickshell/Modules/AppDrawer/AppLauncher.qml
+++ b/quickshell/Modules/AppDrawer/AppLauncher.qml
@@ -51,6 +51,10 @@ Item {
         function onPluginLoaded() { updateCategories() }
         function onPluginUnloaded() { updateCategories() }
         function onPluginListUpdated() { updateCategories() }
+        function onRequestLauncherUpdate(pluginId) {
+            // Only update if we are actually looking at this plugin or in All category
+            updateFilteredModel()
+        }
     }
 
     Connections {
--- a/quickshell/Services/AppSearchService.qml
+++ b/quickshell/Services/AppSearchService.qml
@@ -411,26 +411,43 @@ Singleton {
             return []
         }
 
-        const component = PluginService.pluginLauncherComponents[pluginId]
-        if (!component)
+        let instance = PluginService.pluginInstances[pluginId]
+        let isPersistent = true
+
+        if (!instance) {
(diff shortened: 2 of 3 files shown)


bbedward:
Great thank you, I'll get a chance to review this later. But you should refactor your qalculate plugin to use QtObject and PluginSettings, I made PRs to existing launcher plugins for the same (like DankCalculator)

pcortellezzi:
Thanks for the suggestion! 🙂 I've refactored both the Qalculate and Powermenu plugins to use `QtObject` and `PluginSettings` following your advice, thanks!

bbedward (approved):

Issue #824, opened 2025-11-27
CurrentThemeName in default.settings not respected


HirschBerge:
## Compositor

- [ ] niri
- [X] Hyprland
- [ ] dwl (MangoWC)
- [ ] sway
- [ ] Other (specify)

## Distribution

NixOS

## dms version

0.6.2+date=2025-11-23_2ae9112

## Description

When setting currentThemeName as such
```
  currentThemeName = "cat-mauve";
```
via home-manager module, all settings besides theme are set.

## Expected Behavior

cat-mauve theme to be displayed

## Steps to Reproduce

https://gist.github.com/HirschBerge/a47d8887c8e34d99b7feefb43d96fe06

## Actual Behavior
Default Blue scheme is applied.

## Other
Not particularly painful, since it's easy enough to go into the settings and update the theme, but Nix is about declarative configs, so it would be nice to have it work properly.

Thanks!

HayyaSylvy:
Can report the same issue here, it seems for me that the Shell can't access the file on the read-only Nix Store like nix sets it so it doesn't detect the theme if its there for some reason. A workaround might be possible on Home Manager by using a custom theme, then using the xdg module to generate it somewhere in your user's home folder then setting that as a file on the DMS module, haven't tested it yet but it SHOULD work :P

Update: It does work, the declared theme is indeed read if set to a file outside of the /nix/store folder even if that file is still read-only. :D. Not a definitive solution, but it does work as a makeshift in the meantime.

HirschBerge:
The more I think about it, the more I think the real solution is just respecting the declaration, instead of this wacky "default-settings" stuff. Have it read directly from nix, then if you make changes, it go to changed-settings.json or something where you can then use json2nix to add in new settings

ewtodd:
@HirschBerge I completely agree. Noctalia has something similar, where if it cannot write to the standard settings file it will instead write to a file called gui-settings.json. It is then easy to diff the two files and convert the new json to nix. The home.nix file could be changed to something like the following: 

```
{ config, pkgs, lib, dmsPkgs, ... }@args:
let
  cfg = config.programs.dankMaterialShell;
  jsonFormat = pkgs.formats.json { };
  common = import ./common.nix { inherit config pkgs lib dmsPkgs; };
in {
  imports = [
    (import ./options.nix args)
    (lib.mkRemovedOptionModule [
      "programs"
      "dankMaterialShell"
      "enableNightMode"
[...]

PR #1233, opened 2026-01-01
nix: update home-manager module to remove default*, add clsettings

ewtodd:
Update home-manager module to match removal of default* paradigm as discussed in #824 .

### DIFF
--- a/distro/nix/home.nix
+++ b/distro/nix/home.nix
@@ -16,6 +16,12 @@ let
       dmsPkgs
       ;
   };
+  hasPluginSettings = lib.any (plugin: plugin.settings != { }) (
+    lib.attrValues (lib.filterAttrs (n: v: v.enable) cfg.plugins)
+  );
+  pluginSettings = lib.mapAttrs (name: plugin: { enabled = plugin.enable; } // plugin.settings) (
+    lib.filterAttrs (n: v: v.enable) cfg.plugins
+  );
 in
 {
   imports = [
@@ -24,22 +30,19 @@ in
       "programs"
       "dank-material-shell"
       "enableNightMode"
-    ] "Night mode is now always available.")
-    (lib.mkRemovedOptionModule
-      [
-        "programs"
-        "dank-material-shell"
-        "default"
-        "settings"
-      ]
-      "Default settings have been removed and been replaced with programs.dank-material-shell.settings."
-    )
+    ] "Night mode is now always available")
+    (lib.mkRemovedOptionModule [
+      "programs"
+      "dank-material-shell"
+      "default"
+      "settings"
+    ] "Default settings have been removed and been replaced with programs.dank-material-shell.settings")
     (lib.mkRemovedOptionModule [
       "programs"
       "dank-material-shell"
       "default"
       "session"
-    ] "Default session has been removed and been replaced with programs.dank-material-shell.session.")
+    ] "Default session has been removed and been replaced with programs.dank-material-shell.session")
     (lib.mkRenamedOptionModule
       [ "programs" "dank-material-shell" "enableSystemd" ]
       [ "programs" "dank-material-shell" "systemd" "enable" ]
@@ -63,6 +66,12 @@ in
       type = jsonFormat.type;
       default = { };
       description = "DankMaterialShell session settings as an attribute set, to be written to ~/.local/state/DankMaterialShell/session.json.";
+    };
+
+    managePluginSettings = lib.mkOption {
+      type = lib.types.bool;
+      default = hasPluginSettings;
+      description = ''Whether to manage plugin settings. Automatically enabled if any plugins have settings configured.'';
     };
   };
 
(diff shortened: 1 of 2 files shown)


distro/nix/home.nix
```diff
@@ -25,24 +25,41 @@ in
...
+      "dank-material-shell"
+      "default"
+      "settings"
+    ] "Declared settings are now respected.")
+    (lib.mkRemovedOptionModule [
+      "programs"
+      "dank-material-shell"
+      "default"
+      "session"
+    ] "Declared session settings are now respected.")
```
LuckShiba: I think this message may be a bit confusing, I'd rather something around the lines of "Default [session] settings support got removed in favor of `programs.dank-material-shell.(settings/session)`." instead

distro/nix/home.nix
```diff
@@ -25,24 +25,41 @@ in
...
-        description = "The default session are only read if the session.json file don't exist";
-      };
+  options.programs.dank-material-shell = {
+    settings = lib.mkOption {
+      type = jsonFormat.type;
+      default = { };
+      description = "DankMaterialShell configuration settings as an attribute set, to be written to ~/.config/DankMaterialShell/settings.json.";
+    };
+
+    clsettings = lib.mkOption {
```
LuckShiba: What about naming this `clipboardSettings`?

distro/nix/home.nix
```diff
@@ -67,22 +84,25 @@ in
...
+    xdg.configFile = {
+      "DankMaterialShell/settings.json" = lib.mkIf (cfg.settings != { }) {
+        source = jsonFormat.generate "settings.json" cfg.settings;
+      };
+      "DankMaterialShell/clsettings.json" = lib.mkIf (cfg.clsettings != { }) {
+        source = jsonFormat.generate "clsettings.json" cfg.clsettings;
+      };
+      "DankMaterialShell/plugin_settings.json" = (
+        lib.mapAttrs' (name: value: {
+          name = "DankMaterialShell/plugins/${name}";
```
LuckShiba: Image

This is not working.

distro/nix/home.nix
ewtodd: That's a good point; I changed it to "Default settings have been removed and been replaced with programs.dank-material-shell.settings." and equivalent for session.

distro/nix/home.nix
ewtodd: That would definitely be more obvious; this is also done.

distro/nix/home.nix
ewtodd: I'm not sure about this because I think the plugin_settings have never worked for me as currently  written. I have been handling it manually like 
```
{ pkgs, ... }:
let
  dmsPlugins = pkgs.fetchFromGitHub {
    owner = "AvengeMedia";
    repo = "dms-plugins";
    rev = "8715ca35b61d7c6275a300fa4992e2b0490f70f7";
    hash = "sha256-6ScuVcvSaXc35Sf1iwtCy8aM/pFID3+0G4NRMw8aBcM=";
  };
  jsonFormat = pkgs.formats.json { };
in
{

  xdg.configFile."DankMaterialShell/plugin_settings.json" = {
    source = jsonFormat.generate "plugin_settings.json" {
      dankPomodoroTimer.enabled = true;
    };
  };

  xdg.configFile."DankMaterialShell/plugins/" = {
    source = dmsPlugins;
    recursive = true;
  };
}
```
But it seems like there should be a better way to handle this. It would be ideal for example if one could simply enable any of the plugins that are official, without needing to fetch the source themselves. Am I missing something, or is this the intended behavior? I'm happy to make any changes after the intended functionality is clear.

LuckShiba: We have the `plugins` option that will write to the plugins directory. For example:

```nix
programs.dank-material-shell.plugins = {
  dankPomodoroTimer.src = "${dmsPlugins}/dankPomodoroTimer";
};
```

And this will copy this to `.config/DankMaterialShell/plugins/dankPomodoroTimer`. this is not related to the plugin_settings.json, which also includes other settings, for example:
```json
{
    "calculator": {
      "enabled": true,
      "trigger": "=",
      "noTrigger": false
    }
  }
  ```

Maybe we can add a `settings` field in the plugins object to write to this file? And also link the .enable of `plugins.<name>.enable = true;` to write to this file?

For example:
```nix
{
  programs.dank-material-shell.plugins = {
    dankPomodoroTimer = {
      src = ...;
      settings = {
        trigger = "=";
        noTrigger = false;
      };
    };
  };
}
```

Note: if you're doing this, I'd like to have an option to disable this in case someone doesn't want to have a nix-managed `plugin_settings.json`.
  
Now about having plugins without source, I think a better approach for this is https://github.com/AvengeMedia/dms-plugin-registry/pull/69. It'll integrate nicely with this.

ewtodd: Thanks for the clarification. I think adding a settings field in the plugins object makes a lot of sense. The example you gave is how I'd assume most nix users would expect this to work, and having an option to disable it is definitely also reasonable. I will update this PR to include that. 

About https://github.com/AvengeMedia/dms-plugin-registry/pull/69, do you envision the user including that flake in inputs, and ending up with something like 

```
  programs.dank-material-shell.plugins = {
    dankPomodoroTimer = {
      src = inputs.dms-plugin-registry.packages.${pkgs.system}.dankPomodoroTimer;
      settings = {
        trigger = "=";
        noTrigger = false;
      };
    };
  };
```
The issue I see here is that the intended behavior of the module in the other flake is to not use the packages directly (like you would when combining with the regular dms module), but instead to use the enable option defined in the other module. It seems there is a bit of overlap there?

LuckShiba: `enable` option already exists in the current module, it's not defined in the other module. What it does is simply setting it by default as false, which means it won't be copied into the folder.

The way Nix works, with declarative rather than imperative options, even if defined in the other  flake, if you do
```nix
  programs.dank-material-shell.plugins = {
    dankPomodoroTimer = {
      enable = true;
      settings = {
        trigger = "=";
        noTrigger = false;
      };
    };
  };
  ```
  
  It _will_ work properly, with the `src` being defined by the registry flake, and since all others are with `enable = false`, this repo's module will simply ignore it. There's no overlap :)

ewtodd: I see. I should have looked more closely at the other module; I didn't realize it was only setting defaults rather than defining new options (particularly because the naming still follows the old camel case convention). I've not seen an example of an external module defining defaults for a separate module before, but it's always good to learn more Nix. I agree that it is definitely a more elegant solution for handling the plugin source. 

In any case, the most recent commit has the following behavior:
If there are no plugins with settings defined but managePluginSettings = true, a minimal plugin_settings.json is created that enables the plugins. If there are plugin settings defined, managePluginSettings defaults to true and those settings are placed in plugin_settings.json. If there are plugin settings defined and managePluginSettings is set to false, there will be a warning, but plugin_settings.json won't be created. 

If it looks good to you I can also submit a PR to the documentation repo.

LuckShiba: It does, contributions to the docs are welcome!

PR #1234, opened 2026-01-01
util: add flatpak introspection utilities

porkloin:
:wave: 

I'm using DMS on [zirconium](https://github.com/zirconium-dev/zirconium) which is a universal blue/bootc project, so many pieces of software are installed via flatpak rather than via package manager.

I noticed that matugen tries to determine whether or not to generate theme files for supported applications based on some shell execs in the utils dir. I'm hoping to contribute some work in a follow-on PR that allows the configuration in `core/internal/matugen/matugen.go` to take flatpak package IDs, for example something like:

```
appendConfig(opts, cfgFile, []string{"zen", "zen-browser", "app.zen_browser.zen"}, "zenbrowser.toml")
```

The semantics of how to implement that can be hashed out later, but this PR contributes some simple utils that would be needed to have matugen interact with flatpaks more broadly. The utils aren't rigged up or called from any other part of the codebase currently, but it provides the following util funcs for getting flatpak info:

- `FlatpakInPath` - determines if the user even has flatpak installed, that way we can fail fast if not
- `FlatpakExists` - for a given fully-qualified flatpak id (foo.bar.myapp) returns whether or not the package is installed on host system
- `FlatpakSearchBySubstring` - for a conventional app name ("calculator") returns whether or not a package with a matching name exists on host system (org.calc.calculator)
- `FlatpakInstallationDir` - gets the home sandbox dir for a sandbox. Probably helpful if future matugen work wants/needs to stuff configs directly into the sandbox dir of the flatpak

I have also added some steps to the go-ci.yml file for installing flatpak onto the runner so that my tests can run in the CI environment. Long term it would probably be best to just pick a runner that has flatpak pre-installed, or bake an image that adds flatpak onto ubuntu-latest to prevent needing to do all that network shit on every PR commit :upside_down_face: 

I wanted to just propose this smaller PR initially that stubs out the core set of interactions with flatpak, but I'm happy to also chat here more about ideas or concerns about integrating flatpak support into matugen.

Let me know what you folks think - I'm happy to adapt or adjust this if needed. Thanks for all the amazing work on DMS!

### DIFF
--- a/.github/workflows/go-ci.yml
+++ b/.github/workflows/go-ci.yml
@@ -28,6 +28,15 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - name: Install flatpak
+        run: sudo apt update && sudo apt install -y flatpak
+
+      - name: Add flathub
+        run: sudo flatpak remote-add --if-not-exists flathub https://flathub.org/repo/flathub.flatpakrepo
+
+      - name: Add a flatpak that mutagen could support
+        run: sudo flatpak install -y org.freedesktop.Platform/x86_64/24.08 app.zen_browser.zen
+
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
--- a/core/internal/utils/flatpak.go
+++ b/core/internal/utils/flatpak.go
@@ -1,0 +1,78 @@
+package utils
+
+import (
+	"bytes"
+	"errors"
+	"os/exec"
+	"strings"
+)
+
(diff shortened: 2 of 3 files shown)


porkloin:
Reworked a couple missing sudos in the GHA workflow, but it's running successfully and passing on my fork :+1:

bbedward:
This is good, but what's the next step for flatpak theme support?

PR #1239, opened 2026-01-02
feat(distro/nix/niri): add a hack for config includes with niri flake

NikSneMC:
It works fine but needs all dms-generated config files to be present

That hack should be replaced with proper config includes after https://github.com/sodiboo/niri-flake/pull/1548 merge

The only problem is that [optional includes](https://yalter.github.io/niri/Configuration%3A-Include.html#optional-includes) aren't released yet.


distro/nix/niri.nix
```diff
@@ -15,10 +15,43 @@ in
     niri = {
       enableKeybinds = lib.mkEnableOption "DankMaterialShell niri keybinds";
       enableSpawn = lib.mkEnableOption "DankMaterialShell niri spawn-at-startup";
+      enableImportsHack = lib.mkEnableOption "imports support for niri-flake";
```
LuckShiba: for future compability, I think we can call this simply `enableImports` (and have it default as true)

distro/nix/niri.nix
```diff
@@ -15,10 +15,43 @@ in
...
+            "outputs"
+            "wpblur"
+          ];
+        in
+        # HACK: niri-flake does not support config includes yet, but we can "fix" that
+        {
+          niri-config.target = lib.mkForce "niri/${hmFile}.kdl";
+          "niri/config.kdl".text = lib.pipe dmsFiles [
+            (map (filename: "dms/${filename}"))
+            (dmsFiles: [ hmFile ] ++ dmsFiles)
```
LuckShiba: We discussed about that in the linked PR, it can be relevant if the user wants DMS includes to be before or after the user settings since includes are positional. I suggest something like

```nix
{
  includes = {
    enable = true; # default
    override = true; # default, description: "Whether DMS settings will be prioritized or not over settings defined in niri-flake"
  };
}
```

LuckShiba (commented):
Thank you for this, this will be very good for users that use `niri-flake` while we don't get includes in the upstream! A few comments though.

distro/nix/niri.nix
NikSneMC: I think it would be nice to allow custom hm file name, DMS-generated file list and dms config dir path (if it could be theoretically changed anywhere)

distro/nix/niri.nix
NikSneMC: Yeah, I think that's a better approach

LuckShiba (approved):

LuckShiba:
Looks good to me, and is working here :)

NikSneMC:
I've got overrides wrong initially. Fixed

NikSneMC:
Looks like I've completely forgot about the [border special case](https://yalter.github.io/niri/Configuration%3A-Include.html#border-special-case). Working on a fix

UPD. fixed

LuckShiba:
> aLooks like I've completely forgot about [border special case](https://yalter.github.io/niri/Configuration%3A-Include.html#border-special-case). Working on a fix
> 
> UPD. fixed

Good catch. Can't we simply add the fixes in the generated config.kdl file rather than in another file? I guess this way may be good if we need more fixes which touches the same parts in the future, but will these cases happen?

NikSneMC:
> Can't we simply add the fixes in the generated config.kdl file rather than in another file?

Yes, we can. It's just a matter of preference ¯\\_(ツ)\_/¯

> I guess this way may be good if we need more fixes which touches the same parts in the future, but will these cases happen?

Theoretically, yes. Niri-flake may not add support for new stuff until [that](https://github.com/sodiboo/niri-flake/pull/1548) rewrite happens, but every new niri version can introduce new changes that may require fixes like this one. But I think my solution is definitely overengineered)

LuckShiba:
> But I think my solution is definitely overengineered

It surely is, but I guess we can keep it that way for now. I'll wait on @marcusramberg's input on this and we can merge :)

NikSneMC:
> > Can't we simply add the fixes in the generated config.kdl file rather than in another file?
> 
> Yes, we can. It's just a matter of preference ¯\\_(ツ)\_/¯

After a comparison I think that a single config.kdl is cleaner and a little bit easier to work with

marcusramberg:
Lgtm :)

tbaumann:
The fix is there in the merge request (the follow up one to be precise https://github.com/epireyn/niri-flake/pull/1)

I'm happy to use the custom branch. Perhaps time to work on a proper include version of this?  (switchable or auto-detect)

PR #1247, opened 2026-01-02
lockscreen/greetd: add 0 in front of single digit hours for 12 hour format. greetd: add option to hide profile image

ewtodd:
This adds an option to greetd to hide the profile image using the existing option for the locker. For both greetd and the locker, the time formatting for 12 hour formats is changed to add a preceding zero for single digit hours so that the clock always appears properly centered. This behavior is already present elsewhere in DMS such as in the vertical clock module.

Making a new pull request because it seems that I cannot reopen the original one.

### DIFF
--- a/quickshell/Common/SettingsData.qml
+++ b/quickshell/Common/SettingsData.qml
@@ -146,6 +146,7 @@ Singleton {
 
     property bool use24HourClock: true
     property bool showSeconds: false
+    property bool padHours12Hour: false
     property bool useFahrenheit: false
     property string windSpeedUnit: "kmh"
     property bool nightModeEnabled: false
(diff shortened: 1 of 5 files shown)


quickshell/Common/SettingsData.qml
```diff
@@ -935,7 +935,7 @@ Singleton {
         if (use24HourClock) {
             return showSeconds ? "hh:mm:ss" : "hh:mm";
         } else {
-            return showSeconds ? "h:mm:ss AP" : "h:mm AP";
+            return showSeconds ? "hh:mm:ss AP" : "hh:mm AP";
```
bbedward, line 1259: Personally I'd like this to be an option or fine-tuned a bit, I like this on the lock screen but not on the clock bar widget (personally at least), maybe better to add more options besides 24-hour such as custom format, locale-aware format, etc.

bbedward (changes requested):

quickshell/Common/SettingsData.qml
ewtodd: I think a custom format makes sense, or at least a list of different formats like there is for the date. I think it also makes sense to split it for top bar and lockscreen like the date currently is. If that sounds good to you, I can try and implement that.

bbedward: Yea I think that makes sense, maybe a configurable dropdown like calendar formats (or two for lock/greeter and bar/elsewhere). But it probably needs a migration to be backwards compat with current 24H setting.

ewtodd:
Sorry; did not mean to close. Will get back to this when I have time.

PR #1248, opened 2026-01-02
Bar (mediaplayer): Mouse wheel options for media player widget

philjackson:
Options for the media player widget:
* Scroll to change current player volume as it behaves now. This is still the default.
* Scroll to change tracks in the current player.
* Nothing

image


quickshell/Modules/Settings/MediaPlayerTab.qml
```diff
@@ -46,11 +46,23 @@ Item {
...
+                        "Change Volume": "volume",
+                        "Change Song": "song",
+                        "Nothing": "nothing"
+                    }
+
                     text: I18n.tr("Scroll Wheel")
-                    description: I18n.tr("Scroll on widget changes media volume")
-                    checked: SettingsData.audioScrollEnabled
-                    onToggled: checked => SettingsData.set("audioScrollEnabled", checked)
+                    description: I18n.tr("Scroll wheel behavior on media widget")
```
bbedward, line 57: lets add some tags to this, it will help the search/index `tags: ["media", "music", "scroll"]` something like that

bbedward (changes requested):

quickshell/Modules/Settings/MediaPlayerTab.qml
philjackson: @bbedward do I need to regenerate settings_search_index.json myself, or does that happen at build/runtime?

edit: looks like it's in git so I guess I do it.

PR #1251, opened 2026-01-03
feat: matugen detects flatpak installations of zenbrowser and vesktop

porkloin:
Follow-on PR for #1234. This PR actually updates the `matugen.go` logic in `appendConfig` to take an additional argument `checkFlatpaks` which can be a list of known flatpak IDs. For the majority of applications defined elsewhere in `matugen.go` the arg is assigned to `nil`. I've added two flatpaks which I happen to know the IDs for as an initial test case: zen browser and vesktop.

I also added some tests to show that this shouldn't block on cases where a user has both flatpak and bin installs, etc.

Let me know if you'd like to see anything else here!

### DIFF
--- a/core/internal/matugen/matugen.go
+++ b/core/internal/matugen/matugen.go
@@ -281,22 +281,22 @@ output_path = '%s'
 		appendConfig(opts, cfgFile, []string{"equibop"}, nil, "equibop.toml")
 	}
 	if !opts.ShouldSkipTemplate("ghostty") {
-		appendTerminalConfig(opts, cfgFile, tmpDir, []string{"ghostty"}, "ghostty.toml")
+		appendTerminalConfig(opts, cfgFile, tmpDir, []string{"ghostty"}, nil, "ghostty.toml")
 	}
 	if !opts.ShouldSkipTemplate("kitty") {
-		appendTerminalConfig(opts, cfgFile, tmpDir, []string{"kitty"}, "kitty.toml")
+		appendTerminalConfig(opts, cfgFile, tmpDir, []string{"kitty"}, nil, "kitty.toml")
 	}
 	if !opts.ShouldSkipTemplate("foot") {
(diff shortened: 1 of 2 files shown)


core/internal/matugen/matugen.go
```diff
@@ -342,14 +342,23 @@ output_path = '%s'
...
+	checkFlatpaks []string,
+	fileName string,
+) {
 	configPath := filepath.Join(opts.ShellDir, "matugen", "configs", fileName)
 	if _, err := os.Stat(configPath); err != nil {
 		return
 	}
 	if len(checkCmd) > 0 && !utils.AnyCommandExists(checkCmd...) {
 		return
 	}
```
bbedward: This will short circuit and not check the flatpaks if the first binary check doesnt find it in the path.

Should proably be like this:

```go
cmdExists := len(checkCmd) == 0 || utils.AnyCommandExists(checkCmd...)
flatpakExists := len(checkFlatpaks) == 0 || utils.AnyFlatpakExists(checkFlatpaks...)

if !cmdExists && !flatpakExists {
    return
}
```

Same for appendTerminalConfig (should also get the flatpak argument)

bbedward (changes requested):

core/internal/matugen/matugen.go
porkloin: Good catch! Just fixed in a new commit, but `== 0` check won't quite work on `nil` so it's slightly different but same general idea.

bbedward: len(slice) in golang results in 0 when the slice is `nil` , in many operations in GO nil slices behavior like empty slices.

Albeit, in this context it doesnt matter since the "skip evaluation" conditions are explicitly setting `nil` 

But the distinction is

`len(slice) == 0` evaluates to true for nil and pointers to an empty slice (not nil but empty slice)

`checkCmd == nil` only evaluates to true for nil pointer, but not for an empty slice.

PR #1259, opened 2026-01-03
Add doctor command

LuckShiba:
Free healthcare for your DankMaterialShell

image

### DIFF
--- a/core/cmd/dms/commands_doctor.go
+++ b/core/cmd/dms/commands_doctor.go
@@ -13,6 +13,7 @@ import (
 	"github.com/AvengeMedia/DankMaterialShell/core/internal/config"
 	"github.com/AvengeMedia/DankMaterialShell/core/internal/distros"
 	"github.com/AvengeMedia/DankMaterialShell/core/internal/server/brightness"
+	"github.com/AvengeMedia/DankMaterialShell/core/internal/server/network"
 	"github.com/AvengeMedia/DankMaterialShell/core/internal/tui"
 	"github.com/AvengeMedia/DankMaterialShell/core/internal/utils"
 	"github.com/AvengeMedia/DankMaterialShell/core/internal/version"
@@ -582,6 +583,29 @@ func checkI2CAvailability() checkResult {
 	return checkResult{catOptionalFeatures, "I2C/DDC", statusOK, fmt.Sprintf("%d monitor(s) detected", len(devices)), "External monitor brightness control"}
 }
 
+func detectNetworkBackend() string {
+	result, err := network.DetectNetworkStack()
+	if err != nil {
+		return ""
+	}
+
+	switch result.Backend {
+	case network.BackendNetworkManager:
+		return "NetworkManager"
+	case network.BackendIwd:
+		return "iwd"
+	case network.BackendNetworkd:
+		if result.HasIwd {
+			return "iwd + systemd-networkd"
+		}
+		return "systemd-networkd"
+	case network.BackendConnMan:
+		return "ConnMan"
+	default:
+		return ""
+	}
+}
+
 func checkOptionalDependencies() []checkResult {
 	var results []checkResult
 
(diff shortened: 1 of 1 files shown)


core/cmd/dms/commands_doctor.go
```diff
@@ -0,0 +1,756 @@
...
+		})
+	}
+
+	return results
+}
+
+func checkEnvironmentVars() []checkResult {
+	var results []checkResult
+	results = append(results, checkEnvVar("QT_QPA_PLATFORMTHEME")...)
+	results = append(results, checkEnvVar("QT_ICON_THEME")...)
```
bbedward: its QS_ICON_THEME in this instance, quickshell-specific variable for that

core/internal/utils/paths.go
```diff
@@ -6,6 +6,30 @@ import (
...
+func xdgDir(envVar string, defaultPath ...string) string {
+	if dir := os.Getenv(envVar); dir != "" {
+		return dir
+	}
+	home, _ := os.UserHomeDir()
+	return filepath.Join(append([]string{home}, defaultPath...)...)
+}
+
+func XDGConfigHome() string {
+	return xdgDir("XDG_CONFIG_HOME", ".config")
```
bbedward: Some of these prefer using internal go stuff

```
import os

dir, err := os.UserConfigDir() // Maps to XDG_CONFIG_HOME

// os.UserCacheDir()
```

Think it stops there though, they dont have a standard lib helper for data_home/state_home.

core/internal/utils/exec.go
```diff
@@ -15,3 +18,16 @@ func AnyCommandExists(cmds ...string) bool {
...
+	if !CommandExists("systemctl") {
+		return false
+	}
+
+	args := []string{"is-active", name}
+	if userService {
+		args = []string{"--user", "is-active", name}
+	}
+	output, _ := exec.Command("systemctl", args...).Output()
+	return strings.TrimSpace(string(output)) == "active"
```
bbedward: `strings.EqualFold(strings.TrimSpace(string(output)), "active")`  can be used to guarantee case insensitive matching.

core/cmd/dms/commands_doctor.go
```diff
@@ -0,0 +1,756 @@
...
+			currentCategory = r.category
+		}
+		printResultLine(r, styles)
+	}
+}
+
+func printResultLine(r checkResult, styles tui.Styles) {
+	icon, style := "○", styles.Subtle
+	switch r.status {
+	case statusOK:
```
bbedward: You could use a type alias and put a method on the type alias for stuff like this

```go
type status string

const (
  statusOK status = "OK"
  ....
)

func (s *status) IconStyle() (string, lipgloss.Style) {
   switch *s {...
}
```

bbedward: Or maybe instead of that a struct that holds more stauts information, in general

```go
type DoctorStatus struct {
  Errors ...
  Warnings...
}
```

core/cmd/dms/commands_doctor.go
```diff
@@ -0,0 +1,756 @@
...
+		if !found && d.altCmd != "" {
+			if utils.CommandExists(d.altCmd) {
+				found, foundCmd = true, d.altCmd
+			}
+		}
+
+		if found {
+			message := "Installed"
+			switch foundCmd {
+			case "nmcli":
```
bbedward: Unrelated but could enhance later to add the server logic that selects which backend to use (to show what is actually in-use) - that's like network manager (only one implementing VPN) || iwd (wifi only) || iwd+systemd-networkd  || networkd (eth only)

bbedward (changes requested):

core/internal/utils/paths.go
LuckShiba: Oh, didn't know this was a thing

PR #1304, opened 2026-01-08
fix typos

antonkesy:
Found some recurring typos and fixed many of them.

Most of them are only comments. However, I'm not sure if changing the text in `core/internal/proto/xml/dwl-ipc-unstable-v2.xml` will have any breaking effects (but I doubt it).

### DIFF
--- a/core/internal/config/embedded/hyprland.conf
+++ b/core/internal/config/embedded/hyprland.conf
@@ -106,7 +106,7 @@ windowrule = float on, match:class ^(firefox)$, match:title ^(Picture-in-Picture
 windowrule = float on, match:class ^(zoom)$
 
 # DMS windows floating by default
-# ! Hyprland doesnt size these windows correctly so disabling by default here
+# ! Hyprland doesn't size these windows correctly so disabling by default here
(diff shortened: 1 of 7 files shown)


core/internal/proto/xml/dwl-ipc-unstable-v2.xml
```diff
@@ -24,7 +24,7 @@ I would probably just submit raphi's patchset but I don't think that would be po
       This interface is exposed as a global in wl_registry.
 
       Clients can use this interface to get a dwl_ipc_output.
-      After binding the client will recieve the dwl_ipc_manager.tags and dwl_ipc_manager.layout events.
```
bbedward, line 27: I prefer not to change this, its just taken directly from the [MangoWC](https://github.com/DreamMaoMao/mangowc/blob/main/protocols/dwl-ipc-unstable-v2.xml) repository. It's just cleaner to keep this a 1:1 replica of that, and re-generate the protocol bindings as needed.

antonkesy: Totally agree 👍🏻 I've removed changes to any proto/* file

PR #1312, opened 2026-01-09
feat: steam game icons

shalevc1098:
steam proton games now show their actual game icons instead of the generic controller. maps steam_app_XXXXX window class to steam_icon_XXXXX which is what steam installs.

### DIFF
--- a/quickshell/Common/Paths.qml
+++ b/quickshell/Common/Paths.qml
@@ -59,6 +59,9 @@ Singleton {
                 return "transmission";
             return "transmission-gtk";
         }
+        const steamMatch = appId.match(/^steam_app_(\d+)$/);
+        if (steamMatch)
+            return `steam_icon_${steamMatch[1]}`;
         return appId;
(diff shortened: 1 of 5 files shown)


Bodify:
@bbedward @shalevc1098 Could this be tweaked to still fallback to the controller? I'm not sure the exact reasons but it is not working for me. For non-steam games I get it but seems it does not work with some steam games either. Deadlock is a game in closed alpha atm though so not sure if that impacts it.

Btw @bbedward as a sidenote to this, would it be possible to extend the "App ID Substitutions" to have an option to check window title? If running a game through gamescope for example app id comes back "Unknown" or null. Window title on the other hand is listed as "gamescope - Deadlock" or something similar I believe.
image
image

bbedward:
Check this for a fallback a8230953720ba0da1c56b9a1a296f64278446bd6 - open a new one for the window titles (just so it doesnt get lost)

PR #1317, opened 2026-01-10
feat: configurable app id substitutions

shalevc1098:
adds a setting to map window class names to icon names. includes pattern matching (exact, contains, regex) and icons update live when substitutions change.

also fixes cursor not showing on header action buttons in non-collapsible settings cards.

### DIFF
--- a/quickshell/Modules/Settings/RunningAppsTab.qml
+++ b/quickshell/Modules/Settings/RunningAppsTab.qml
@@ -38,6 +38,7 @@ Item {
                 iconName: "find_replace"
                 title: I18n.tr("App ID Substitutions")
                 settingKey: "appIdSubstitutions"
+                tags: ["app", "icon", "substitution", "replacement", "pattern", "window", "class", "regex"]
 
                 headerActions: [
                     DankActionButton {
@@ -105,7 +106,6 @@ Item {
                                         DankTextField {
                                             id: patternField
                                             width: parent.width
-                                            height: 40
                                             text: modelData.pattern
                                             font.pixelSize: Theme.fontSizeSmall
                                             onEditingFinished: SettingsData.updateAppIdSubstitution(index, text, replacementField.text, modelData.type)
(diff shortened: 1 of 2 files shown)


quickshell/Modules/Settings/RunningAppsTab.qml
```diff
@@ -32,6 +32,160 @@ Item {
                     onToggled: checked => SettingsData.set("runningAppsCurrentWorkspace", checked)
                 }
             }
+
+            SettingsCard {
+                width: parent.width
+                iconName: "find_replace"
+                title: I18n.tr("App ID Substitutions")
+                settingKey: "appIdSubstitutions"
```
bbedward, line 40: Add tags: here for search index, then regenerate with `python3 translations/extract_settings_index.json`

quickshell/Modules/Settings/RunningAppsTab.qml
```diff
@@ -32,6 +32,160 @@ Item {
...
+                                        StyledText {
+                                            text: I18n.tr("Pattern")
+                                            font.pixelSize: Theme.fontSizeSmall - 1
+                                            color: Theme.surfaceVariantText
+                                        }
+
+                                        DankTextField {
+                                            id: patternField
+                                            width: parent.width
+                                            height: 40
```
bbedward: Better to not hardcode height and let it size to accomodate the font size, in case its bigger.

quickshell/Modules/Settings/RunningAppsTab.qml
```diff
@@ -32,6 +32,160 @@ Item {
...
+                                        StyledText {
+                                            text: I18n.tr("Replacement")
+                                            font.pixelSize: Theme.fontSizeSmall - 1
+                                            color: Theme.surfaceVariantText
+                                        }
+
+                                        DankTextField {
+                                            id: replacementField
+                                            width: parent.width
+                                            height: 40
```
bbedward: Also remove hardcoded height

PR #1361, opened 2026-01-14
Add screencast indicator for niri

YaLTeR:
Uses the [new niri cast IPC](https://yalter.github.io/niri/niri_ipc/struct.Cast.html). Adds an indicator to the control center button (enabled by default, configurable in widget settings) and to the lock screen system icons.

Updating the privacy indicator widget is left as an exercise for the LLM agent.


Purian23 (approved):
Thanks Yalt!

PR #1385, opened 2026-01-16
nix: Support specifying systemd target

bailey-busc:
This is a pretty common pattern in home manager: https://home-manager-options.extranix.com/?query=systemd.target

It's not necessarily true that one will always want to bind to the default target, like in the case that one is running DMS in Niri but sometimes running other compositor sessions where one doesn't want to run DMS. This supports that case in a cleaner way than `mkForce`ing the unit configuration. I opted to just duplicate the option between the home manager module and the NixOS module due to NixOS not having anything analogous to `config.wayland.systemd.target`.

### DIFF
--- a/distro/nix/home.nix
+++ b/distro/nix/home.nix
@@ -73,6 +73,12 @@ in
       default = hasPluginSettings;
       description = ''Whether to manage plugin settings. Automatically enabled if any plugins have settings configured.'';
     };
+
+    systemd.target = lib.mkOption {
+      type = lib.types.str;
+      default = config.wayland.systemd.target;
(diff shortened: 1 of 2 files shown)


LuckShiba (approved):
We already have this option in the [nixpkgs module](https://search.nixos.org/options?channel=unstable&show=programs.dms-shell.systemd.target&query=dms-shell), so it's fair to have here too. Thanks :)

shalevc1098:
hey, this pr seems to have introduced a syntax error in distro/nix/home.nix - nix complains about a missing semicolon at the end of the file. opened #1386 about it

PR #1389, opened 2026-01-16
Add environment PATH in home-manager module.

luis-agm:
## Motivation

In the case when a user overrides the quickshell package, the service fails to start since it can't find the `qs` executable.

## Proposed changes

Add the quickshell package to the `Environment` configuration for the service.

I've tested this by overriding the property in my home-manager configuration and it restored the desired behavior of the shell activating after login.

### DIFF
--- a/distro/nix/home.nix
+++ b/distro/nix/home.nix
@@ -99,7 +99,7 @@ in
         ExecStart = lib.getExe dmsPkgs.dms-shell + " run --session";
         Restart = "on-failure";
         Environment = [
-          "PATH=${lib.makeBinPath common.packages}"
+          "PATH=${lib.makeBinPath common.packages}:/run/current-system/sw/bin:/usr/local/bin:/usr/bin:/bin"
         ];
       };
 


distro/nix/home.nix
```diff
@@ -98,6 +98,9 @@ in
       Service = {
         ExecStart = lib.getExe dmsPkgs.dms-shell + " run --session";
         Restart = "on-failure";
+        Environment = [
+          "PATH=${lib.makeBinPath [ dmsPkgs.dms-shell cfg.quickshell.package ]}:$PATH"
```
LuckShiba: I think that if we're adding this here, `common.packages` could be used instead for adding all deps. Also, have you checked if this $PATH substitution really works here?

luis-agm: Good point! I've added the quickshell package to the common.packages. Also I tested the end result and the $PATH indeed didn't work, so I've removed it.

LuckShiba: If we can't have the "normal" PATH then, this is a problem. Some cases are:

- DMS checks for application binaries for generating themes (e.g. check if the `vesktop` command is present for generating a vesktop theme);
- Plugins may also want to use a binary not defined in the service's PATH, such as the terminal desktop widget plugin that needs python in order to work.

Of course, being NixOS, the user could add themselves these binaries to the service's PATH, but this is less practical.

luis-agm: Interesting, I haven't found a way to expand the `PATH` into the service's environment. But doing this directly in the dms flake and loading it locally seems to work (up to a point, some errors with theme generation).
```
        Environment = [
          "PATH=${lib.makeBinPath common.packages}:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin"
        ];
```

The errors I see are:
```
Jan 18 12:20:46 fedora dms[133871]: error: unexpected argument '--import-json-string' found
Jan 18 12:20:46 fedora dms[133871]:   tip: to pass '--import-json-string' as a value, use '-- --import-json-string'
Jan 18 12:20:46 fedora dms[133871]: Usage: matugen color hex --mode <MODE> --type <TYPE> --config <FILE> <STRING>
Jan 18 12:20:46 fedora dms[133871]: For more information, try '--help'.
Jan 18 12:20:46 fedora dms[133565]:  ERROR  go: DMS API Error: id=1 error=exit status 2
Jan 18 12:20:46 fedora dms[133578]:   WARN qml: Theme worker:  FATAL  go: Theme generation failed: server error: exit status 2
Jan 18 12:20:46 fedora dms[133578]:   WARN qml: Theme: Matugen worker failed with exit code: 1
Jan 18 12:20:46 fedora dms[133578]:   INFO qml: HyprlandService: Generated layout config at /home/<user>/.config/hypr/dms/layout.conf
Jan 18 12:20:46 fedora dms[133578]:   INFO qml: CompositorService: Detected Hyprland
Jan 18 12:20:46 fedora dms[133578]:   INFO qml: HyprlandService: Generated layout config at /home/<user>/.config/hypr/dms/layout.conf
Jan 18 12:20:46 fedora dms[133578]:   INFO qml: SessionService: Native inhibitor available: true
```
These look solvable. However, you were right, several processes that failed before in the logs don't fail with this approach, meaning the process has access to stuff it needs. Open to suggestions on how to make this work. 
Also keep in mind this is a home-manager module, not a NixOs module, so the service is created as a user service. Maybe that sheds some light on alternatives.

The errors that appeared previously (some of them) point to the process not being able to find some very basic essential tools.
```
Jan 18 11:55:08 fedora dms[97078]:   WARN: Process failed to start, likely because the binary could not be found. Command: QList("sh", "-c", "command -v dsearch")
Jan 18 11:55:08 fedora dms[97078]:   WARN: Process failed to start, likely because the binary could not be found. Command: QList("which", "wtype")
Jan 18 11:55:08 fedora dms[97078]:   WARN: Process failed to start, likely because the binary could not be found. Command: QList("cat", "/etc/os-release")
Jan 18 11:55:08 fedora dms[97078]:   WARN: Process failed to start, likely because the binary could not be found. Command: QList("which", "dgop")
Jan 18 11:55:08 fedora dms[97078]:   WARN: Process failed to start, likely because the binary could not be found. Command: QList("which", "matugen")
Jan 18 11:55:08 fedora dms[97078]:   WARN: Process failed to start, likely because the binary could not be found. Command: QList("sh", "-c", "gsettings get org.gnome.desktop.sound theme-name 2>/dev/null")
```

LuckShiba: Maybe you can instead of setting in Environment= setting them in a prefix in the run command? For example ``sh -c "PATH=${dmsPath}:$PATH dms run"`` (or even using coreutils' `env`)

Please try it out.

luis-agm: using `sh -c ...` works but swallows all the logs. Setting the env like this works fine:
```
Environment = [
      "PATH=${lib.makeBinPath common.packages}:/run/current-system/sw/bin:/usr/local/bin:/usr/bin:/bin"
];
```
And it shows all the logs from the service.

LuckShiba: It may work but is not complete... We can't really guess what the final user's path will be. I'll see if I can find an alternative

purian23:
Feel free to re-open in the future. Closing out.

PR #1396, opened 2026-01-16
Feat/niri workspace names

kamilchm:
## Summary
- switch niri workspace models to keep name fields and object data through the switcher
- fix niri workspace labels to align with 0-based indices and live name changes
- use the workspace index toggle to prefix named workspaces as `index: name`

## Testing
- Manual: restart dms, renamed workspaces with `niri msg action set-workspace-name`, verified labels update

image

### DIFF
--- a/quickshell/Modules/DankBar/Widgets/WorkspaceSwitcher.qml
+++ b/quickshell/Modules/DankBar/Widgets/WorkspaceSwitcher.qml
@@ -866,8 +866,10 @@ Item {
 
                 readonly property real baseWidth: root.isVertical ? (SettingsData.showWorkspaceApps ? widgetHeight * 0.7 : widgetHeight * 0.5) : (isActive ? root.widgetHeight * 1.05 : root.widgetHeight * 0.7)
                 readonly property real baseHeight: root.isVertical ? (isActive ? root.widgetHeight * 1.05 : root.widgetHeight * 0.7) : (SettingsData.showWorkspaceApps ? widgetHeight * 0.7 : widgetHeight * 0.5)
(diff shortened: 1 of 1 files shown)


quickshell/Modules/DankBar/Widgets/WorkspaceSwitcher.qml
```diff
@@ -802,8 +870,8 @@ Item {
                     return 0;
                 }
 
-                readonly property real visualWidth: baseWidth + iconsExtraWidth
-                readonly property real visualHeight: baseHeight + iconsExtraHeight
+                readonly property real visualWidth: Math.max(baseWidth + iconsExtraWidth, contentImplicitWidth > 0 ? contentImplicitWidth + Theme.spacingS : 0)
+                readonly property real visualHeight: Math.max(baseHeight + iconsExtraHeight, contentImplicitHeight > 0 ? contentImplicitHeight + Theme.spacingS : 0)
```
bbedward: For me it changes the height/behavior of the workspace pills, even if I don't have names. Just with index numbers it increases the height significantly of each pill. Better to not change current appearance like that, IMO

kamilchm: I think I now prefer add `spacingXS` to the height.

1. No spacing
image

2. `spacingXS`
image

3. `spacingS` - as used currenlty in the PR
 
image

Pick one, and I will make change to the PR.

bbedward: for names or icons and things it makes sense, I just mean when I have index numbers only (no names) it is also increasing the height. But I'll circle back to these PRs I can offer some help and get it merged.

PR #1402, opened 2026-01-16
lock screen power off monitors

### LINKED ISSUE #1157: [Feature request] Power off the monitor after Lock Screen
## Feature Description

Add an option to power off the monitors automatically when the lock screen is activated.

## Use Case

This feature helps save energy on laptops. Its behavior mirrors that of GNOME: as soon as the lock screen appears, the monitor turns off instantly.

## Compositor

- [x] All compositors
- [ ] niri
- [ ] Hyprland
- [ ] dwl (MangoWC)
- [ ] sway

## Proposed Solution
Include an option in the “Lock Screen” section of the DMS settings panel.

kamilchm:
Add ability to immediately power off monitors when the lock screen activates, controlled by a new setting "Power off monitors on lock". Uses a 100ms polling timer to detect when the session lock actually becomes active, then invokes compositor-specific DPMS commands.

For niri, uses the new power-off-monitors action via niri msg CLI with socket fallback.

Wake on input: first input after lock arms wake, second input actually powers monitors back on while keeping the lock screen visible.

Closes #1157

### DIFF
--- a/quickshell/Common/SettingsData.qml
+++ b/quickshell/Common/SettingsData.qml
@@ -394,6 +394,7 @@ Singleton {
     property bool lockScreenShowDate: true
     property bool lockScreenShowProfileImage: true
     property bool lockScreenShowPasswordField: true
+    property bool lockScreenPowerOffMonitorsOnLock: false
 
     property bool enableFprint: false
     property int maxFprintTries: 15
--- a/quickshell/Common/settings/SettingsSpec.js
+++ b/quickshell/Common/settings/SettingsSpec.js
@@ -259,6 +259,7 @@ var SPEC = {
     lockScreenShowDate: { def: true },
     lockScreenShowProfileImage: { def: true },
     lockScreenShowPasswordField: { def: true },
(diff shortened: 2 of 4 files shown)


quickshell/Services/NiriService.qml
```diff
@@ -774,19 +774,31 @@ Singleton {
     }
 
     function powerOffMonitors() {
-        return send({
-            "Action": {
-                "PowerOffMonitors": {}
-            }
+        Proc.runCommand("niri-power-off-monitors", ["niri", "msg", "action", "power-off-monitors"], (output, exitCode) => {
```
bbedward: What's the rationale of using the CLI command at all?

bbedward (changes requested):

quickshell/Services/NiriService.qml
kamilchm: I couldn't make the API command works. But it actually might works now as I found another bug that fixed it. Will test again.

kamilchm: Fixed, no niri cli call now.

PR #1406, opened 2026-01-16
Add French translation 🇫🇷

Landeli7:


bbedward:
This needs to be done through poeditor, you should be able to request access and then import the json to there. https://poeditor.com/join/project/kyeCPTpmJM

Landeli7:
Thank you, I've requested approval on poeditor

bbedward:
Granted, thanks for the translation - we just have to go through poeditor then I sync them from there to the codebase periodically.

PR #1425, opened 2026-01-18
feat(bar): enlarge bar icons if widget background is off

Eggrror404:
After turning off widget background on the bar, the padding becomes empty space, and the icons seem to small:

image

So this PR changes the icon size to `iconSizeLarge` if `noBackground` is true.

image

### DIFF
--- a/quickshell/Common/Theme.qml
+++ b/quickshell/Common/Theme.qml
@@ -752,9 +752,9 @@ Singleton {
         return (0.299 * c.r + 0.587 * c.g + 0.114 * c.b) < 0.5;
     }
 
-    function barIconSize(barThickness, offset) {
+    function barIconSize(barThickness, offset, noBackground) {
         const defaultOffset = offset !== undefined ? offset : -6;
-        const size = (SettingsData.barConfigs[0] || SettingsData.getBarConfig("default")).noBackground ? iconSizeLarge : iconSize;
+        const size = (noBackground ?? false) ? iconSizeLarge : iconSize;
 
         return Math.round((barThickness / 48) * (size + defaultOffset));
     }
--- a/quickshell/Modules/DankBar/Widgets/Battery.qml
+++ b/quickshell/Modules/DankBar/Widgets/Battery.qml
@@ -42,7 +42,7 @@ BasePill {
 
                 DankIcon {
                     name: BatteryService.getBatteryIcon()
-                    size: Theme.barIconSize(battery.barThickness)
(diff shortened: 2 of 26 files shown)


quickshell/Common/Theme.qml
```diff
@@ -754,7 +754,9 @@ Singleton {
 
     function barIconSize(barThickness, offset) {
         const defaultOffset = offset !== undefined ? offset : -6;
-        return Math.round((barThickness / 48) * (iconSize + defaultOffset));
+        const size = (SettingsData.barConfigs[0] || SettingsData.getBarConfig("default")).noBackground ? iconSizeLarge : iconSize;
```
bbedward: This should probably be passed into the function as a parameter, instead of  only working for the default/first bar config.

Eggrror404: True, I did not think of that. But this way we may have to explicitly pass in `undefined` for `offset` to use the third parameter.

Eggrror404:
@bbedward I updated the PR and this now works with multiple bars. `noBackground` is passed in as a function parameter so it uses the bar config the widget is on. Please take a look.

I also updated every use of the function, adding `undefined` for offset if necessary. looks a bit ugly though

quickshell/Common/Theme.qml
```diff
@@ -752,9 +752,11 @@ Singleton {
         return (0.299 * c.r + 0.587 * c.g + 0.114 * c.b) < 0.5;
     }
 
-    function barIconSize(barThickness, offset) {
+    function barIconSize(barThickness, offset, noBackground) {
```
bbedward, line 755: optional arguments should be supported `function barIconSize(barThickness, offset, noBackground = undefined)`

Then all function arguments don't need to be updated to pass undefined, explicitly.

Eggrror404: I believe the value would be undefined if the parameter is not passed in, by default. So the `= undefined` is unnecessary.
The reason I pass in noBackground for every call is that I want the icon to always update its size based on the value.

And since Javascript doesn't support named parameters like `barIconSize(noBackground = something)` I have to add in explicit undefined.

bbedward: We might wanna enlarge text too because it looks a little funky, but we'll see who complains

Eggrror404: Oh yeah the text too

There's a setting for text size for bars so I guess it's fine...?

PR #1429, opened 2026-01-19
feat: add workspace rename dialog

kamilchm:
- Adds a modal dialog to rename the current workspace
- Supports both Niri (via IPC socket) and Hyprland (via hyprctl dispatch)
- Default keybinding: Ctrl+Shift+R to open the dialog
- Pre-fills with current workspace name
- Allows setting empty name to reset to default

image

Works the best with #1396

### DIFF
--- a/quickshell/DMSShell.qml
+++ b/quickshell/DMSShell.qml
@@ -631,8 +631,16 @@ Item {
         }
     }
 
-    WorkspaceRenameModal {
-        id: workspaceRenameModal
+    LazyLoader {
+        id: workspaceRenameModalLoader
+
+        active: false
+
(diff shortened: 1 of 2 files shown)


quickshell/DMSShell.qml
```diff
@@ -631,6 +631,10 @@ Item {
         }
     }
 
+    WorkspaceRenameModal {
```
bbedward: I'd like to see this wrapped in a LazyLoader, similar to pattern of other windows.  Keeps a little bit lower of a memory footprint when it can be unloaded.

kamilchm: fixed

PR #1566, opened 2026-02-02
  fix(spotlight): prevent auto-selection on open without mouse movement

xxyangyoulin:
Fix Visual distraction: Users might see an item highlighted that they didn't intend to select.

### DIFF
--- a/quickshell/Modals/DankLauncherV2/Controller.qml
+++ b/quickshell/Modals/DankLauncherV2/Controller.qml
@@ -26,6 +26,9 @@ Item {
     property string activePluginId: ""
     property var collapsedSections: ({})
     property bool keyboardNavigationActive: false
+    property bool mouseHasMoved: false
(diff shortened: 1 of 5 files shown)


bbedward:
I dont exactly understand the problem that this is solving, can you elaborate more?

xxyangyoulin:
Modifies the Spotlight/DankLauncher behavior where items would be automatically highlighted when the launcher opens if the mouse cursor was resting over them, even without any actual mouse  movement. The solution adds a ouseHasMoved flag that tracks real mouse movement (with a 0.5px threshold), and updates the hover detection logic in GridItem, ResultItem, and TileItem to only  respond to hover states after the user has actively moved their mouse. This prevents unintended selection while maintaining normal hover behavior after mouse interaction.

d976b83e-f046-4486-bbef-2b1614371147

PR #1602, opened 2026-02-05
feat: media playback OSD

ArijanJ:
Adds OSD for media play/pause and track changes. SwayOSD has this and I missed it in DMS:
Recording 2026-02-05 at 22 43 35(1)

The play/pause button is clickable and toggles play/pause

Vertical variant only shows the button on play/pause - it's hidden on track changes, since there's no room to show the track title and artist:
image

(I can't get prek to work, so sorry if there are any formatting issues or similar that I missed)


ArijanJ:
Tbf this can be made to look much better if there's album art available, but I'm not very experienced with quickshell layouting/styling

purian23:
Cool idea, I updated just a few items to better match the code base. We can always update to show the album art backgrounds as an option in the future reusing the code from the Media Hub.

PR #1638, opened 2026-02-09
Making the new media playback OSD more beautiful

ArijanJ:
gif:

Recording 2026-02-09 at 23 20 40

There is a slight delay that is not very noticeable with a fast internet connections, but might be worse on slow connections.

I copied the UI/blurring code from the media hub,

However, I moved the art downloader code into a separate service so the OSD can use it as well ([quickshell/Services/TrackArtService.qml](https://github.com/AvengeMedia/DankMaterialShell/compare/master...ArijanJ:beautiful-playback-osd?expand=1#diff-a1b847e15b4ec6e0d75e14aca951f59edddfee10a5b8242caa7905cedd157bc4))
The "api" is basically the same, although I was able to get it to break if I go from a player with album art -> player without album art -> player with album art again (until track skip/end). But I doubt anyone with a regular workflow will hit that.

I chose a 0.3 blur instead of 0.8 for the bigger surface I copied it from, but even with less blur it would look fine most of the time (future customization potential?):
image

### DIFF
--- a/quickshell/Services/TrackArtService.qml
+++ b/quickshell/Services/TrackArtService.qml
@@ -5,6 +5,7 @@ import Quickshell
 import QtQuick
 
 import Quickshell.Io
+import Quickshell.Services.Mpris
 
 Singleton {
     id: root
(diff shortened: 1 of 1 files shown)


ArijanJ:
Sorry for the pushes, the bug I mentioned should be fixed now

bbedward (approved):

PR #1639, opened 2026-02-10
MangoWC added to Greeter.nix

yipfluoric:
MangoWC for the greeter seems to work out of box. its just that greeter.nix does not currently have it set to be put as an option, in the main repo.

### DIFF
--- a/quickshell/Modals/Clipboard/ClipboardEntry.qml
+++ b/quickshell/Modals/Clipboard/ClipboardEntry.qml
@@ -1,5 +1,6 @@
 import QtQuick
 import qs.Common
+import qs.Services
 import qs.Widgets
 
 Rectangle {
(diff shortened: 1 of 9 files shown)


yipfluoric:
putting this to draft. it appears almost done but obviously because i wanted the custom greet config to work that is the only part that doesnt work. going to see what in greeter.nix seems to make it the custom config is in /etc/mango/config.conf (does not work or exist for me)

yipfluoric:
hmm. no it does work, greetd does not restart on switch. but im unsure if it trying to pull for the default mango directory on other oses (without the custom config) is common? it may possibly just need a default one so that it doesnt try to use that edit: it does work perfectly fine, its just the fact it says it's trying to pull from that directory is weird. fixed by giving it a config via `dank-material-shell.enable.greeter.compositor.customConfig`

yipfluoric:
ill undraft it. all of it works but does seem to show it pulling from there without a conf. other than that it works

distro/nix/greeter.nix
```diff
@@ -71,6 +71,7 @@ in
         "hyprland"
         "sway"
         "labwc"
+        "mangowc"
```
LuckShiba, line 74: What about the package? Does mangowc exposes the package `mangowc` or is it only `mango`? Do you have any extra settings needed that need to be documented? Check the `compositorPackage` variable

yipfluoric: deleted a prior comment. it does seem to do mangowc? extra settings dont need to be documented. however i am gonna have to add something for mango flake users to decrease the disparity between `pkgs.mangowc` and `programs.mango`

yipfluoric:
did not see the pull request from my repo right so this one will be closed cause i also added scroll support as well anyways.

PR #1641, opened 2026-02-10
feat: configurable volume amount on scroll for Media widget

eboskma:
Adds a setting to change how much the volume is adjusted when the `audioScrollMode` setting for the Media widget is set to `volume`. 

image


JDKamalakar:
Hey Can You Make So That This Volume Indent Feature Is Applied To Volume Controls Triggered By Shortcuts or Volume Control Wheel.

PR #1647, opened 2026-02-10
MangoWC and Scroll Greeter Support for NixOS

yipfluoric:
Support for MangoWC in greeter from #1639 on both the nixpkgs, and Flake variations. along side this added Scroll. this is flake only but if it does go to nixpkgs at some point it will most likely use the package from there

### DIFF
--- a/quickshell/Modules/Greetd/assets/dms-greeter
+++ b/quickshell/Modules/Greetd/assets/dms-greeter
@@ -253,7 +253,7 @@ SCROLL_EOF
         fi
         ;;
 
-    mango)
+    mango|mangowc)
         if [[ -n "$COMPOSITOR_CONFIG" ]]; then
             exec mango -c "$COMPOSITOR_CONFIG" -s "$QS_CMD && mmsg -d quit"
         else


yipfluoric:
it all works as intended. mangowc installs without mango flake, the package from mango flake gets used when its enabled, and scroll greeter builds and works.

distro/nix/greeter.nix
```diff
@@ -13,7 +13,9 @@ let
 
   compositorPackage =
     let
-      configured = lib.attrByPath [ "programs" cfg.compositor.name "package" ] null config;
+      mango = lib.hasAttrByPath [ "programs" "mango" "package" ] true;
+      compositor = if mango != null then "mango" else cfg.compositor.name;
```
LuckShiba: If a user has mangowc enabled, it will always select it, no matter what you put in cfg.compositor.name? What if a user has mangowc enabled but to use want another compositor? Shouldn't this only apply if you have selected mangowc as package?

From what I can see, in mangowc Flake, the binary, package, and module is mango, so I think this call can be solved by reverting this and simply changing the string from `mangowc` to `mango`!

distro/nix/greeter.nix
```diff
@@ -71,6 +73,8 @@ in
         "hyprland"
         "sway"
         "labwc"
+        "mangowc"
+        "scroll"
```
LuckShiba, line 75: Now for scroll the module part seems to be alright.

distro/nix/greeter.nix
yipfluoric: I was thinking of doing that cause this solution is kind of eh. I'd either have to change all related to this from `mangowc` to `mango` in dms-greeter as well, which will most likely break greetd configs on other distros. or find a way to force this to use `mangowc` near the line where it runs the command instead. considering the command on all distros is `mango` anyways i might just do the former. i feel it is a reasonable enough breakage for making it less confusing give or take

yipfluoric: reverted it and changed the files to point to `mango` instead of `mangowc`

quickshell/Modules/Greetd/assets/dms-greeter
```diff
@@ -30,7 +30,7 @@ Examples:
     dms-greeter --command sway -p /home/user/.config/quickshell/custom-dms
     dms-greeter --command scroll -p /home/user/.config/quickshell/custom-dms
     dms-greeter --command niri --cache-dir /tmp/dmsgreeter
-    dms-greeter --command mangowc
+    dms-greeter --command mango
```
LuckShiba, line 33: @bbedward is it okay if we change here to mango too, considering this is the actual command?

quickshell/Modules/Greetd/assets/dms-greeter
```diff
@@ -253,7 +253,7 @@ SCROLL_EOF
         fi
         ;;
 
-    mangowc)
+    mango)
```
bbedward: This should be `mango|mangowc` to preserve compatibility

quickshell/Modules/Greetd/assets/dms-greeter
bbedward: Yes, but the switch/case should keep mangowc for compatibility. And should remember to update the docs too (https://danklinux.com/docs/dankgreeter/configuration)

LuckShiba (approved):

PR #1670, opened 2026-02-14
distro: add cups-pk-helper to recommended dependencies

cwelsys:
Was getting "The name is not activatable" when trying to interact/add/delete a printer. Noticed the dbus call in core/internal/server/cups/pkhelper.go.

Some distro's might have the pk-helper by default, I honestly don't know.

This commit is only a half measure.. the docs, AUR packages and system doctor would also need updating. I wanted to flag it first.

### DIFF
--- a/distro/debian/dms-git/debian/control
+++ b/distro/debian/dms-git/debian/control
@@ -29,6 +29,7 @@ Depends: ${misc:Depends},
          qml6-module-qtquick-templates,
          qml6-module-qtquick-window,
          qt6ct
+Suggests: cups-pk-helper
 Provides: dms
 Conflicts: dms
 Replaces: dms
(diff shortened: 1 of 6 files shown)


bbedward:
added to doctor and aur packages 14f92669c6e0b63d7128efa5e5ba6c4489159386 - but will have @Purian23  review these since he manages theother distros

purian23:
Hi there, thanks for PR. I'm debating whether I want to put this in Recommends or Suggests so users have more choice. I think BB put the package as optional in the AUR.

cwelsys:
the doctor/AUR change's are sufficient for my use. 

Can close this if you want, or I can draft out the changes to the docs. Actually nvm I see the docs @ cli-doctor#optional-features already shows it.

purian23:
If you'd like to rework these items to Suggests, that would work best as they appear to be installed by default in most distros.

PR #1685, opened 2026-02-15
greeter: Add support for Debian greetd user/group name

krig:
On most distros greetd uses the user and group name "greeter", but on Debian the user and group are called "_greetd".

Have tested `dms greeter install` and `dms greeter sync` with this fix on my system, and it works well.

### DIFF
--- a/core/internal/greeter/installer.go
+++ b/core/internal/greeter/installer.go
@@ -22,13 +22,18 @@ func DetectDMSPath() (string, error) {
 	return config.LocateDMSConfig()
 }
 
-// Tries to figure out the name of the
-// user and group used by greetd
 func DetectGreeterGroup() string {
-	osInfo, err := distros.GetOSInfo()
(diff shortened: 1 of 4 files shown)


core/internal/greeter/installer.go
```diff
@@ -22,6 +22,16 @@ func DetectDMSPath() (string, error) {
 	return config.LocateDMSConfig()
 }
 
+// Tries to figure out the name of the
+// user and group used by greetd
+func DetectGreeterGroup() string {
+	osInfo, err := distros.GetOSInfo()
+	if err == nil && osInfo.Distribution.ID == "debian" {
+		return "_greetd"
```
bbedward: We could probably do this instead by parsing /etc/group and checking for the existence of known greetd groups, greetd, _greetd , and pick the one that exists. That would make it work for debian and debian forks and maybe others.

PR #1687, opened 2026-02-15
Add terminal multiplexer launcher

salah-walid:
Add terminal multiplexer sessions listing launcher for tmux + zellij, possible to add other terminal multiplexers later

- Add a terminal multiplexer session manager modal with support for tmux and zellij backends
- New MuxService singleton that abstracts session listing, attaching, creating, killing, and renaming (only for tmux as zellij doesn't allow for rename outside of a session context [issue](https://github.com/zellij-org/zellij/issues/998#issuecomment-2536092028))
- New MuxModal with searchable session list, keyboard-driven navigation, and shortcuts bar
- New InputModal to allow for renaming and inputing a new session name (can be used later for other purposes)
- Settings tab under System > Multiplexers for configuring multiplexer type, custom commands, and session name filters (with regex support)


bbedward:
I'm gonna take a look at this soon, wanted to wait for 1.4 release

quickshell/DMSShell.qml
```diff
@@ -269,6 +270,10 @@ Item {
         }
     }
 
+    MuxModal {
```
bbedward: Should wrap in a LazyLoader here, like other modals - allow it to unload from memory

quickshell/Modals/MuxModal.qml
```diff
@@ -0,0 +1,644 @@
...
+                ScrollView {
+                    anchors.fill: parent
+                    clip: true
+
+                    Column {
+                        width: parent.width
+                        spacing: Theme.spacingXS
+
+                        Repeater {
+                            model: muxModal.filteredSessions
```
bbedward: Better to use [ScriptModel](https://quickshell.org/docs/v0.2.1/types/Quickshell/ScriptModel/) here

quickshell/Services/MuxService.qml
```diff
@@ -0,0 +1,181 @@
...
+
+            sessionList.push({
+                name: name.trim(),
+                windows: "N/A",
+                attached: !exited
+            })
+        }
+
+        if (sessionList.length !== root.sessions.length)
+            sessionsChanged()
```
bbedward: Don't need to emit this signal here, it will be emited on the below assignment implicitly.

quickshell/Services/MuxService.qml
```diff
@@ -0,0 +1,181 @@
...
+                sessionList.push({
+                    name: parts[0],
+                    windows: parts[1],
+                    attached: parts[2] === "1"
+                })
+            }
+        }
+
+        if (sessionList.length !== root.sessions.length)
+            sessionsChanged()
```
bbedward: Don't need to emit this signal here, it will be emited on the below assignment implicitly.

quickshell/Modals/MuxModal.qml
```diff
@@ -0,0 +1,644 @@
...
+        if (filter.length === 0)
+            return false
+        var parts = filter.split(",")
+        for (var i = 0; i < parts.length; i++) {
+            var pattern = parts[i].trim()
+            if (pattern.length === 0)
+                continue
+            if (pattern.startsWith("/") && pattern.endsWith("/") && pattern.length > 2) {
+                try {
+                    var re = new RegExp(pattern.slice(1, -1))
```
bbedward: It would be more efficient to compute this one time and re-use it, when mux session list changes. But it's not that critical

quickshell/Services/MuxService.qml
```diff
@@ -0,0 +1,181 @@
+pragma Singleton
+pragma ComponentBehavior: Bound
+
+import QtQuick
+import Quickshell
+import Quickshell.Io
+import qs.Common
+
+Singleton {
```
bbedward, line 9: Would be nice to get some onCompleted check here, like other services do, to check isZellijAvailable, isTmuxAvailable in the case they arent available in the PATH

bbedward (changes requested):

quickshell/DMSShell.qml
salah-walid: @bbedward since this is a modal and not a popout following [this similar example of the ClipboardHistoryModal](https://github.com/AvengeMedia/DankMaterialShell/blob/bb08e1233a2a57fc64ee024523cb7010aabb97ce/quickshell/DMSShell.qml#L622) it doesn't seem to be using a lazyloader, if you still prefer the mux modal to be lazy loaded it can still be done

salah-walid:
@bbedward can you please have another look, i also did a rebase, had a lot of conflicts with the translation files, i don't know if i did a good job, let me know if everything is in order

b0o:
Just curious, this seems like it might be better suited as a plugin?

salah-walid:
> Just curious, this seems like it might be better suited as a plugin?

@b0o i thought so too but there are some functionalities that i couldn't perform in a plugin, for example showing an input popup etc...

jpeeler:
Are there any docs for this feature? I don't understand how to use it. Does it require using the native launcher?

derchr:
> Are there any docs for this feature? I don't understand how to use it. Does it require using the native launcher?

I think running `dms ipc tmux open` opens a new tmux/zellij session manager.

Vatson112:
Also note that env variable `TERMINAL` must be set. Not sure if this is specific to my setup.

### Symptoms

- Sessions are visible via `dms ipc mux open`
- Cannot attach, kill, create, or delete any session

In logs nothing usefull:

```bash
May 04 12:17:35 systemd[6089]: Started app-niri-dms-453379.scope.
May 04 12:17:35 dms[6816]:   WARN qt.qpa.wayland.textinput: virtual void QtWaylandClient::QWaylandTextInputv3::enableSurface(wl_surface*) Try to enable surface 0x56269a48c370 with focusing surface 0x56269bf0adc0
```

### Fix

Set `TERMINA`L via a systemd drop-in for dms.service, as described in the [docs](https://danklinux.com/docs/dankmaterialshell/managing#environment-variables):

You may need something like this (modify according to your terminal)

```bash
# ~/.config/systemd/user/dms.service.d/override.conf
[Service]
Environment=TERMINAL=alacritty
```

salah-walid:
@derchr @Vatson112 I will document this in the official website when i have some free time

Titor8115:
@salah-walid Would there be any intent for integrating the `mux` action into "DMS Shell actions" type in keyboard shortcuts' setting instead of shell command run via keybinding? Or this action is intended for later inclusion into DMS Launcher bar.

(Sorry for asking in old Pulls since I haven't found any thing open for similar questions for this fantastic addition)

image

PR #1689, opened 2026-02-15
add support for globalprotect vpn using saml auth flow

dvjn:
This PR adds support for SAML authentication for GlobalProtect VPN connections that require external browser-based login.

**High-Level Flow:**
- Detects when SAML is required based on connection settings (saml-auth-method, missing username/authtype, or protocol indicators)
- Launches gp-saml-gui which opens a browser for user authentication
- Captures the prelogin cookie from gp-saml-gui output
- Converts it to a full auth cookie via openconnect --authenticate
- Caches the cookie to avoid re-authentication on reconnects
- Provides the cookie to NetworkManager's secret agent interface

**Implementation Notes:**
- Requires [gp-saml-gui](https://github.com/dlenski/gp-saml-gui) and [openconnect](https://gitlab.com/openconnect/openconnect).
- Currently only supports GlobalProtect (protocol=gp). Other protocols with SAML indicators fall back to credential auth with a warning log.
- 5-minute timeout for SAML flows to prevent indefinite hangs
- SAML cookies are cached per-connection UUID and cleared on success/failure
- Added unit cases covering detection logic, field inference, parsing, and secret formatting
- End to end testing done on a live GlobalProtect VPN.

**Questions:**
- Should I be documenting the new optional dependencies somewhere?
- Should this have a toast popup instead of a warning log whenever it tries SAML but doesn't find the dependencies?

### DIFF
--- a/core/internal/server/network/agent_networkmanager_test.go
+++ b/core/internal/server/network/agent_networkmanager_test.go
@@ -185,27 +185,27 @@ func TestVpnFieldMeta_GPSaml(t *testing.T) {
 
 func TestVpnFieldMeta_StandardFields(t *testing.T) {
 	tests := []struct {
-		field         string
-		vpnService    string
-		expectedLabel string
+		field          string
+		vpnService     string
+		expectedLabel  string
 		expectedSecret bool
 	}{
 		{
-			field:         "username",
-			vpnService:    "org.freedesktop.NetworkManager.openconnect",
-			expectedLabel: "Username",
+			field:          "username",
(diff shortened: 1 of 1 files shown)


bbedward:
This looks mostly good, documentation should go in https://github.com/AvengeMedia/DankLinux-Docs/tree/master/docs - somewhere in there relevant to optional features or dependencies. Maybe also DMS doctor, although in this case I think the missing deps should just be info-level not warning-level

PR #1692, opened 2026-02-15
fix(launcher): release DankLauncherV2 resources after close

higorprado:
## Summary
DankLauncherV2 could keep a high VRAM footprint after first open/close because the modal subtree stayed loaded. This change releases launcher resources after close while preserving all launcher features.

## What changed
- Added `PopoutService.unloadDankLauncherV2()` to deactivate the launcher `LazyLoader` and drop the modal reference.
- Wired `DankLauncherV2Modal.onDialogClosed` to call `PopoutService.unloadDankLauncherV2()`.
- Updated launcher `PanelWindow.visible` to only stay visible while opening/open/closing:
  - from `root._windowEnabled`
  - to `root._windowEnabled && (spotlightOpen || isClosing)`

## Files changed
- `quickshell/Services/PopoutService.qml`
- `quickshell/DMSShell.qml`
- `quickshell/Modals/DankLauncherV2/DankLauncherV2Modal.qml`

## Why
- Ensure launcher delegates/images/textures are released after close instead of staying pinned in memory.
- Reduce persistent VRAM usage after first open of DankLauncherV2.

## Behavior impact
- No feature removal.
- Tradeoff: launcher is re-instantiated on next open after close (small reopen cost), in exchange for VRAM release.

## Validation
Manual validation recommended:
1. Open `DankLauncherV2`, then close it.
2. Observe VRAM usage returning down after close animation finishes.
3. Reopen launcher and confirm normal behavior (search, navigation, actions, context menu).

PR #1695, opened 2026-02-15
fix(launcher): unload launcher content on close when optional

higorprado:
### Context

Follow-up to PR #1692 (fix(launcher): release DankLauncherV2 resources after close).

Feedback from @AvengeMedia on #1692: full launcher unload was made optional because it introduces a significant keyboard focus delay on reopen; suggested restructuring to unload content while keeping the main window loaded.

### Summary

This PR implements that restructuring.

When unload-on-close is enabled, it unloads only LauncherContent (heavy subtree), while keeping the DankLauncherV2Modal/window object alive.
This preserves optional VRAM savings while reducing reopen/focus latency compared with full modal re-instantiation.

### What changed

- Replaced direct LauncherContent instantiation with a Loader inside DankLauncherV2Modal.
- Added deferred initialization so show(), showWithQuery(), and showWithMode() still initialize correctly after content reload.
- Close cleanup now deactivates only the content loader (when setting is enabled), not the whole modal.

### Why

- Addresses maintainer feedback.
- Keeps behavior/features intact.
- Improves keyboard focus responsiveness on reopen versus full unload.

### File changed

- quickshell/Modals/DankLauncherV2/DankLauncherV2Modal.qml

### Manual validation

1. Enable launcher unload-on-close.
2. Open/close/reopen launcher and verify keyboard focus responsiveness.
3. Verify show, showWithQuery, and showWithMode paths.
4. Confirm heavy launcher content is released after close.

### DIFF
--- a/quickshell/Modals/DankLauncherV2/DankLauncherV2Modal.qml
+++ b/quickshell/Modals/DankLauncherV2/DankLauncherV2Modal.qml
@@ -284,7 +284,7 @@ Item {
 
     PanelWindow {
         id: launcherWindow
-        visible: root._windowEnabled
+        visible: root._windowEnabled && (!root.unloadContentOnClose || spotlightOpen || isClosing)
         color: "transparent"
         exclusionMode: ExclusionMode.Ignore
 


bbedward:
I did this 0d49acaaa860c6db71f5d7655c2bd4c3bb59cf05 to target subtree, more specifically the heavy deleages without unloading the entire LauncherContent. As unloading LauncherContent still effects focus speed/performance, but just the results delegate doesnt seem to have any negative effect. Is that still not good enough vram wise?

higorprado:
Is not a big deal, just was trying to optimize things but, with your change, in my tests, vram keep not being unloaded.

bbedward:
Apparently I accidentally included this in 2c9b22c01606a29b84c562459b697630e7e9d422 , which is fine - it can stay included since its an option.

PR #1727, opened 2026-02-18
Add Custom Media Source Integration via IPC Hooks

higorprado:
## Summary

Add support for external media players (rmpc, cmus, mpv, etc.) to integrate with DMS media widgets through IPC hooks, enabling non-MPRIS players to appear alongside native MPRIS players in the bar, dash, and OSD.

## Problem

DMS media widgets only support MPRIS-compatible players. Many popular terminal-based music players (rmpc/MPD, cmus, moc, etc.) don't expose MPRIS interfaces, leaving users without media controls in the shell UI.

## Solution

Implement a custom media source that receives metadata updates via IPC hooks and executes shell commands for playback control. The system integrates transparently with existing MPRIS infrastructure.

#### CustomMediaSource.qml

A singleton that mirrors the MprisPlayer interface:

- Stores media state (track info, playback state, position, etc.)
- Receives updates via `media.update` IPC command
- Executes configured shell commands for play/pause/next/previous
- Emits signals for OSD and widget change detection

#### MprisController.qml

Extended as the unified media interface:

- `currentPlayer` property returns either MPRIS player or CustomMediaSource
- Direct property bindings (`currentIsPlaying`, `currentTrackTitle`, etc.) for reliable UI updates
- Auto-selection logic: custom source takes precedence when it has content loaded
- Manual source selection via dropdown for user override

#### Widget Updates

All media widgets now use `MprisController.currentPlayer`:

- Bar media widget (play/pause, track scrolling)
- Dash media tab (full controls, seekbar, album art)
- Media OSD (playback state notifications)
- Volume OSD (per-player volume control)

#### IPC Commands

New `media` target for external integrations:

| Command | Description |
|---------|-------------|
| `media.update <json>` | Push track metadata and state |
| `media.setCommands <json>` | Configure playback shell commands |
| `media.status` | Query current state |
| `media.play` | Trigger play command |
| `media.pause` | Trigger pause command |
| `media.playPause` | Trigger toggle command |
| `media.next` | Trigger next track command |
| `media.previous` | Trigger previous track command |
| `media.clear` | Clear media state |

### Integration Example

External tools integrate by calling IPC from their hooks. Example with rmpc:

**~/.config/rmpc/dms_hook:**
```bash
#!/bin/bash
# Set playback commands (idempotent - safe to call every time)
dms ipc call media setCommands '{"play":"rmpc play","pause":"rmpc pause","toggle":"rmpc togglepause","next":"rmpc next","prev":"rmpc prev"}'

# Push current track info to DMS
dms ipc call media update "{\"title\":\"$TITLE\",\"artist\":\"$ARTIST\",\"album\":\"$ALBUM\",\"duration\":$DURATION,\"state\":1,\"available\":true,\"sourceId\":\"rmpc\"}"
```

**~/.config/rmpc/config.ron:**
```ron
on_song_change: [
    "~/.config/rmpc/dms_hook"
]
```

The hook pattern ensures commands are always available even after DMS restarts.

## Changes

| File | Change |
|------|--------|
| `Services/CustomMediaSource.qml` | New - Custom media source singleton |
| `Services/MprisController.qml` | Extended - Unified player interface with direct property bindings |
| `DMSShellIPC.qml` | Extended - Add `media` IPC target |
| `Modules/DankBar/Widgets/Media.qml` | Updated - Use unified interface |
| `Modules/DankDash/MediaPlayerTab.qml` | Updated - Use unified interface |
| `Modules/DankBar/Widgets/AudioVisualization.qml` | Updated - Use unified interface |
| `Modules/DankDash/Overview/MediaOverviewCard.qml` | Updated - Use unified interface |
| `Widgets/DankSeekbar.qml` | Updated - Use unified interface |
| Other media-related widgets | Updated - Consistent use of unified interface |
| `docs/IPC.md` | Extended - Documentation for `media` target |

## Testing

Tested with rmpc (MPD client):

- [x] Hook fires on song change → DMS receives metadata
- [x] Play/pause button toggles state and updates UI immediately
- [x] Next/previous buttons skip tracks
- [x] Album art displays correctly
- [x] Auto-selection keeps custom source active even when paused
- [x] Manual source selection via player dropdown works
- [x] MPRIS players (Firefox, Spotify) still work alongside custom source

### DIFF
--- a/quickshell/Services/MprisController.qml
+++ b/quickshell/Services/MprisController.qml
@@ -29,7 +29,7 @@ Singleton {
 
     // Direct property bindings for reliable UI updates (QML doesn't track nested var properties well)
     readonly property int currentPlaybackState: useCustomSource ? CustomMediaSource.playbackState : (activePlayer ? activePlayer.playbackState : 0)
-    readonly property bool currentIsPlaying: currentPlaybackState === 1
+    readonly property bool currentIsPlaying: currentPlaybackState === CustomMediaSource.statePlaying
     readonly property string currentTrackTitle: useCustomSource ? CustomMediaSource.trackTitle : (activePlayer ? activePlayer.trackTitle : "")
     readonly property string currentTrackArtist: useCustomSource ? CustomMediaSource.trackArtist : (activePlayer ? activePlayer.trackArtist : "")
     readonly property string currentTrackAlbum: useCustomSource ? CustomMediaSource.trackAlbum : (activePlayer ? activePlayer.trackAlbum : "")
--- a/quickshell/Widgets/DankAlbumArt.qml
+++ b/quickshell/Widgets/DankAlbumArt.qml
@@ -22,7 +22,7 @@ Item {
     }
 
     Loader {
-        active: activePlayer?.playbackState === 1 && showAnimation
+        active: activePlayer?.playbackState === CustomMediaSource.statePlaying && showAnimation
         sourceComponent: Component {
             Ref {
                 service: CavaService
@@ -35,7 +35,7 @@ Item {
         width: parent.width * 1.1
         height: parent.height * 1.1
         anchors.centerIn: parent
-        visible: CavaService.cavaAvailable && activePlayer?.playbackState === 1 && showAnimation
+        visible: CavaService.cavaAvailable && activePlayer?.playbackState === CustomMediaSource.statePlaying && showAnimation
         asynchronous: false
         antialiasing: true
         preferredRendererType: Shape.CurveRenderer
--- a/quickshell/Modules/DankDash/MediaPlayerTab.qml
+++ b/quickshell/Modules/DankDash/MediaPlayerTab.qml
@@ -67,7 +67,7 @@ Item {
     // Derived "no players" state: always correct, no timers.
     readonly property int _playerCount: allPlayers ? allPlayers.length : 0
     readonly property bool _noneAvailable: _playerCount === 0
-    readonly property bool _trulyIdle: activePlayer && activePlayer.playbackState === 0 && !activePlayer.trackTitle && !activePlayer.trackArtist
+    readonly property bool _trulyIdle: activePlayer && activePlayer.playbackState === CustomMediaSource.stateStopped && !activePlayer.trackTitle && !activePlayer.trackArtist
     readonly property bool showNoPlayerNow: (!_switchHold) && (_noneAvailable || _trulyIdle)
 
     property bool _switchHold: false
@@ -192,7 +192,7 @@ Item {
 
     Timer {
         interval: 1000
-        running: activePlayer?.playbackState === 1 && !isSeeking
+        running: activePlayer?.playbackState === CustomMediaSource.statePlaying && !isSeeking
         repeat: true
         onTriggered: activePlayer?.positionChanged()
     }
--- a/quickshell/Modules/DankDash/Overview/MediaOverviewCard.qml
+++ b/quickshell/Modules/DankDash/Overview/MediaOverviewCard.qml
@@ -33,7 +33,7 @@ Card {
 
     Timer {
         interval: 300
-        running: activePlayer?.playbackState === 1 && !isSeeking
+        running: activePlayer?.playbackState === CustomMediaSource.statePlaying && !isSeeking
         repeat: true
         onTriggered: activePlayer?.positionSupported && activePlayer.positionChanged()
     }
--- a/quickshell/Modules/Lock/LockScreenContent.qml
+++ b/quickshell/Modules/Lock/LockScreenContent.qml
@@ -1190,7 +1190,7 @@ Item {
                     anchors.verticalCenter: parent.verticalCenter
 
                     Loader {
-                        active: MprisController.currentPlayer?.playbackState === 1
+                        active: MprisController.currentPlayer?.playbackState === CustomMediaSource.statePlaying
 
                         sourceComponent: Component {
                             Ref {
@@ -1200,7 +1200,7 @@ Item {
                     }
 
                     Timer {
-                        running: !CavaService.cavaAvailable && MprisController.currentPlayer?.playbackState === 1
+                        running: !CavaService.cavaAvailable && MprisController.currentPlayer?.playbackState === CustomMediaSource.statePlaying
                         interval: 256
                         repeat: true
                         onTriggered: {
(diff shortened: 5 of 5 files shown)


quickshell/Services/CustomMediaSource.qml
```diff
@@ -0,0 +1,151 @@
...
+        if (cmds.pause !== undefined) _pauseCommand = cmds.pause
+        if (cmds.toggle !== undefined) _toggleCommand = cmds.toggle
+        if (cmds.next !== undefined) _nextCommand = cmds.next
+        if (cmds.previous !== undefined) _previousCommand = cmds.previous
+        if (cmds.prev !== undefined) _previousCommand = cmds.prev
+        if (cmds.volume !== undefined) _volumeCommand = cmds.volume
+    }
+
+    function _exec(cmd) {
+        if (cmd) Quickshell.execDetached(["sh", "-c", cmd])
```
bbedward, line 108: It's a little risky that shell commands could potentially be injected into the config, but I think it's a low risk.

quickshell/Widgets/DankAlbumArt.qml
```diff
@@ -22,7 +22,7 @@ Item {
     }
 
     Loader {
-        active: activePlayer?.playbackState === MprisPlaybackState.Playing && showAnimation
+        active: activePlayer?.playbackState === 1 && showAnimation
```
bbedward: Why replace the enum here with numbers?

quickshell/Widgets/DankAlbumArt.qml
```diff
@@ -7,7 +7,7 @@ import qs.Services
 Item {
     id: root
 
-    property MprisPlayer activePlayer
+    property var activePlayer
```
bbedward, line 10: I get the rationale I guess for making this a generic type, but I'm not too keen on losing the QML type here. Maybe a second property or a wrapped QtObject

quickshell/Services/MprisController.qml
```diff
@@ -10,4 +10,62 @@ Singleton {
...
+    property string forcedSource: "auto"
+
+    // Check if custom source has actual content
+    readonly property bool customHasContent: CustomMediaSource.available && CustomMediaSource.trackTitle !== ""
+
+    // Use custom source when forced or when auto-detect determines it
+    readonly property bool useCustomSource: {
+        if (forcedSource === "custom") return CustomMediaSource.available
+        if (forcedSource === "mpris") return false
+        // Auto: use custom when it has content (track loaded), regardless of playing state
```
bbedward, line 24: Why do we want to use it regardless of playing state?

quickshell/Widgets/DankAlbumArt.qml
higorprado: You're right, using magic numbers is not ideal. I've replaced them with `CustomMediaSource.statePlaying/stateStopped/statePaused` constants. These work for both `MprisPlayer` and `CustomMediaSource` since both use the same numeric values (0, 1, 2). This keeps the code readable while maintaining compatibility with the unified media interface.

quickshell/Widgets/DankAlbumArt.qml
higorprado: Using `var` is a pragmatic solution for polymorphism in QML. The `activePlayer` property can be either an `MprisPlayer` or `CustomMediaSource`, and QML doesn't support union types. Both types expose the same interface (`playbackState`, `trackTitle`, `trackArtUrl`, etc.), so a wrapper or dual properties would add complexity without practical benefit.

quickshell/Services/MprisController.qml
higorprado: This is intentional behavior. External players like rmpc or cmus pause frequently while users work. When paused, users still expect to see their current track in the UI. If we only used the custom source when actively playing, the UI would switch away to another MPRIS player (or show nothing) every time the user paused — which would be confusing. The `customHasContent` check means "a track is loaded", not "track is playing".

higorprado:
Thank you for the review.

### Regarding the enum vs magic numbers

You're right, using magic numbers is not ideal. I've replaced them with `CustomMediaSource.statePlaying/stateStopped/statePaused` constants. These work for both `MprisPlayer` and `CustomMediaSource` since both use the same numeric values (0, 1, 2). This keeps the code readable while maintaining compatibility with the unified media interface.

### Regarding the `var` type for `activePlayer`

Using `var` is a pragmatic solution for polymorphism in QML. The `activePlayer` property can be either an `MprisPlayer` or `CustomMediaSource`, and QML doesn't support union types. Both types expose the same interface (`playbackState`, `trackTitle`, `trackArtUrl`, etc.), so a wrapper or dual properties would add complexity without practical benefit.

### Regarding "use custom source regardless of playing state"

This is intentional behavior. External players like rmpc or cmus pause frequently while users work. When paused, users still expect to see their current track in the UI. If we only used the custom source when actively playing, the UI would switch away to another MPRIS player (or show nothing) every time the user paused — which would be confusing. The `customHasContent` check means "a track is loaded", not "track is playing".

quickshell/Widgets/DankAlbumArt.qml
```diff
@@ -22,7 +22,7 @@ Item {
     }
 
     Loader {
-        active: activePlayer?.playbackState === MprisPlaybackState.Playing && showAnimation
+        active: activePlayer?.playbackState === CustomMediaSource.statePlaying && showAnimation
```
bbedward, line 25: What I am wondering is why cant we re-use the MprisPlaybackState enum that is built into quickshell, instead of having a separate one or fixed numbers? We can re-purpose it for our custom media player values so it matches, same enum.

quickshell/Services/MprisController.qml
```diff
@@ -10,4 +10,62 @@ Singleton {
...
+    }
+
+    function selectSource(source) {
+        if (source && source.isCustom) {
+            forcedSource = "custom"
+        } else if (source && source.player) {
+            forcedSource = "mpris"
+            activePlayer = source.player
+        } else {
+            forcedSource = "auto"
```
bbedward, line 68: There is no path back to "auto" - selectSource seems like it's always called with an argument

quickshell/Modules/DankDash/MediaDropdownOverlay.qml
```diff
@@ -436,28 +444,18 @@ Item {
...
                                     StyledText {
-                                        text: {
-                                            if (!modelData)
-                                                return "";
-                                            const artist = modelData.trackArtist || "";
-                                            const isActive = modelData === activePlayer;
-                                            if (artist.length > 0)
-                                                return artist + (isActive ? " (Active)" : "");
-                                            return isActive ? "Active" : "Available";
-                                        }
```
bbedward, line 460: We lose the artist data here yea?

quickshell/Services/CustomMediaSource.qml
bbedward, line 108: I feel like instead of this, we should just have a list of known players and commands like rpmc, that we can add to later. Feels a little risky to allow it this way.

bbedward (changes requested):

PR #1750, opened 2026-02-19
Added plural support

Youseffo13:
I added some missing i18n strings, added plurarl support and updated it language

without plural support

with plural support


quickshell/Modules/Settings/PrinterTab.qml
```diff
@@ -1310,7 +1310,7 @@ Item {
                                         }
 
                                         StyledText {
-                                            text: I18n.tr("%1 printer(s)").arg(modelData.members?.length ?? 0)
+                                            text: I18n.ntr("%1 printer", "%1 printers", modelData.members?.length ?? 0).arg(modelData.members?.length ?? 0)
```
bbedward, line 1313: This f unction is missing from common/I18n.qml as well as the support for extracting them to poeditor upload in translations/extract_translations.py

PR #1759, opened 2026-02-20
clipboard: improve image thumbnail

lakapple:
- thumbnail image is now bigger
- circular mask has been replaced with rounded rectangular mask


lakapple:
image



This is the old vs new clipboard image thumbnail

PR #1786, opened 2026-02-21
Add support for 'evernight' distribution in Fedora

EvernightFedora:


EvernightFedora:
The Evernight Vista 44 or later it can support.

Issue #1592, opened 2026-02-05
Sharing Wifi via QR code.


nemu626:
### Feature Description

Windows and Android have a feature that generates a QR code from the information of the currently connected Wi-Fi and shares it with another device.
I think it would be quite convenient if DMS also had this built-in. What do you think? I'm thinking of implementing it and submitting a PR, what are your thoughts?

ref : https://www.elevenforum.com/t/share-wi-fi-password-with-qr-code-in-windows-11.26031/

### Use Case

I think it's a good thing for Linux Desktop to implement features found in other OSes like Windows.

### Compositor(s)

- [x] All compositors
- [ ] Niri
- [ ] Hyprland
- [ ] MangoWC (dwl)
- [ ] Sway
- [ ] Other (specify below)

### Proposed Solution

How about a mechanism where a "Share Icon" button is placed next to the currently connected SSID in the Wi-Fi settings of the Settings app, and pressing it brings up a QR code in a popup?

### Alternatives/Existing Solutions

already OSS app is exist.
https://github.com/kokoye2007/wifi-qr

PR #1794, opened 2026-02-21
Wifi QR code generation (requires testing)

Silzinc:
Attempts to solve #1592. This is a somewhat incomplete and best effort implementation, and I might not be able to do anything more than that for now.

I only added support for NetworkManager backend (I would not have been able to test the others). I also only could implement WPA-PSK security type. I did not find a way to know for sure where the password is to be found in all circumstances. Also, although I did extract a text snippet that seemed correct from my phone, I cannot be sure that it actually unlocks the WiFi network as I have no device recent enough to support this feature. Someone will have to check on his/her side...

There might be quite a few clunky blocks of code in there. It is my first time coding in Go and it is not like I have much more experience in QML. I might have not done all of what I had to do for the I18n support either. My apologies for that.

For every WiFi network that is saved and secured, a QR code icon is added on the network info panels. Upon clicking it, two QR codes are generated: one white on transparent background (to be displayed in a popout, as it is easy to colorize it with the theme's palette), and one more standard, black on white background. When the user asks for saving the code, the latter is copied with a rough `cp` command called in a `Quickshell.Io.Process`.

For some reason, I cannot edit the file name in the file browser that opens. I did not find the reason for that, maybe a maintainer will?

I also thought about adding a setting to display the standard code rather than the themed one. That could be done later I guess.

Thanks in advance for reviewing this!

### DIFF
--- a/core/go.mod
+++ b/core/go.mod
@@ -16,6 +16,8 @@ require (
 	github.com/sblinch/kdl-go v0.0.0-20260121213736-8b7053306ca6
 	github.com/spf13/cobra v1.10.2
 	github.com/stretchr/testify v1.11.1
+	github.com/yeqown/go-qrcode/v2 v2.2.5
+	github.com/yeqown/go-qrcode/writer/standard v1.3.0
 	github.com/yuin/goldmark v1.7.16
 	github.com/yuin/goldmark-highlighting/v2 v2.0.0-20230729083705-37449abec8cc
 	go.etcd.io/bbolt v1.4.3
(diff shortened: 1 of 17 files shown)


TwinFox20:
I tested this with the **iwd + systemd-networkd** backend, and this is what I got:

image

From what I can tell, this feature currently only supports the NetworkManager backend, so this behavior is expected.

Would it make sense to open a separate feature request for **iwd** support? I think there are enough users running iwd + networkd that it would be a useful addition.

Silzinc:
I actually think your problem is separate here, bbedward added support for IWD backend in his review before merging. Otherwise, you would have a message like "WiFi QR Code not supported by xxx backend". Systemd network is the only unsupported backend from what I can read in the source code.

Fyi the relevant function that caused this error is [here](https://github.com/AvengeMedia/DankMaterialShell/blob/master/core/internal/server/network/backend_iwd_unimplemented.go), `GetWiFiQRCodeContent`.

PR #1797, opened 2026-02-22
Fix dock visible when theres no app

lakapple:
**Problem**: Dock is still visible when there's no app
**Expected behavior**: Dock is completely hidden

When there's app

When there's no app


lakapple:
I forgot to sync the main branch, so the old clipboard commit is still there

Issue #1793, opened 2026-02-21
HTML elements get parsed in clipboard menu


mezoahmedii:
### Compositor

Niri

### Distribution

Fedora

### Select your Installation Method

DankInstaller

### Was your original Installation method different?

Yes

### dms doctor -vC

<details>
<summary>Click to expand</summary>

## DMS Doctor Report

**System**
- [ok] Operating System: Fedora Linux 43 (Workstation Edition)
  - ID: fedora, Version: 43, Arch: amd64
- [ok] Architecture: amd64
- [ok] Display Server: Wayland
  - WAYLAND_DISPLAY=wayland-1

**Versions**
- [ok] DMS CLI: dms v1.4.0
  - /usr/bin/dms
- [ok] Quickshell: quickshell 0.2.1, revision 26531fc46ef17e9365b03770edd3fb9206fcb460, distributed by: Fedora COPR (avengemedia/quickshell)
  - /usr/bin/qs
- [ok] DMS Shell: 1.4.0
  - /usr/share/quickshell/dms

**Installation**
- [ok] DMS Configuration: Found
  - /usr/share/quickshell/dms
- [ok] shell.qml: Present
  - /usr/share/quickshell/dms/shell.qml
- [info] Install Type: System package
  - /usr/share/quickshell/dms

**Compositor**
- [ok] niri: 25.11
  - /usr/bin/niri
- [info] Active: niri

**Quickshell Features**
- [ok] Polkit: Available
  - Authentication prompts
- [ok] IdleMonitor: Available
  - Idle detection
- [ok] IdleInhibitor: Available
  - Prevent idle/sleep
- [ok] ShortcutInhibitor: Available
  - Allow shortcut management (niri)

**Optional Features**
- [ok] accountsservice: Available
  - User accounts
- [ok] power-profiles-daemon: Available
  - Power profile management
- [ok] logind: Available
  - Session management
- [ok] cups-pk-helper: Available
  - Printer management
- [info] I2C/DDC: No monitors detected
  - External monitor brightness control
- [ok] kimageformats: Installed (4 formats)
  - Formats: AVIF, HEIF, JXL, EXR (/usr/lib64/qt6/plugins/imageformats)
- [ok] Terminal: ghostty
- [ok] Network: NetworkManager
  - NetworkManager present. Using NM API.
- [ok] matugen: Installed
  - Dynamic theming
- [ok] dgop: Installed
  - System monitoring
- [ok] cava: Installed
[...]

PR #1798, opened 2026-02-22
clipboard: fix html elements get parsed in clipboard entry

lakapple:
This pull request fix the issue in https://github.com/AvengeMedia/DankMaterialShell/issues/1793


quickshell/Modals/Clipboard/ClipboardEntry.qml
```diff
@@ -127,14 +127,15 @@ Rectangle {
                 elide: Text.ElideRight
             }
 
-            StyledText {
+            Text {
```
bbedward: Why can't we keep StyledText here? And just override textFormat

lakapple: I'm using zed. The qml extension lsp didn't suggest textFormat attribute for StyledText, so I assumed it doesn't exist.
I've switched back to StyledText.

PR #1802, opened 2026-02-23
shader-based scrolling wallpaper mode

PandorasFox:
some notes:
- needs to be run with `QSG_USE_SIMPLE_ANIMATION_DRIVER=1 qs -p quickshell` if your display refresh rate is >60hz (rolled into the .service file, potentially undesirable on devices that might want to throttle based on battery status? needs investigation beyond the desktop case..)
- plays nice with Lockscreen shader positioning, but haven't tested the greeter (seems fine/non-impacting since the greeter is an earlier stage?)
- haven't tested edge cases with display plugs/mode changes

overall much happier with this approach than the lazy software render impl; did require some boilerplating of the 8th fill mode across the shaders.


PandorasFox:
rebased for 85b00d3c76a03451bc57a9a27e1fc846397bfb0e, i think checks should pass now

bbedward:
Sorry, havent had time to take a look at this - will get to it soon.

purian23:
Thanks for getting it up to date, we will get this one reviewed. Apologies for the delay!

PandorasFox:
that's alright :) I needed to rebase & I got around to one last fix, it's otherwise been good for my daily driving for the past while

PandorasFox:
cleaned up the branch; should be a smaller diff now.

Finally figured out the weird wl_output surface reseating issue I was having, turns out it was racing against if the unlock surface had cleared before or after output reconnect.

Purian23:
Thanks @PandorasFox, I think the largest hurdle here is having to prefix `QSG_USE_SIMPLE_ANIMATION_DRIVER=1 qs -p quickshell` to use it. Users will expect it to just work out of the box and somehow we have to bridge that gap. Your latest efforts may have addressed this.

PandorasFox:
yeah, the need for the 'simple' animation driver to get >60hz frame callbacks from quickshell is still kinda funky to me. As far as I could tell, the [default animation driver](https://doc.qt.io/qt-6/qtquick-visualcanvas-scenegraph.html#what-does-advance-animations-refer-to-in-the-above-diagrams) always runs in 16ms ticks, so >60hz updates weren't possible - which leads to bad motion on higher refresh rate panels, compared to niri's workspace switch animations. if there's another way to get unthrottled frame hints/callback timers, or properly vsynced frame callbacks in the default driver, that'd be great, but I don't think such an interface existed last I checked.

If there's a way to just dynamically change the QSG animation driver for just the wallpaper renderer, when in scrolling mode, that might also be ideal - though, I think that might have weird implications for blurred foreground surfaces (e.g. popouts) that then update at 60hz?

I'm poking at the output/surface re-init after display events and putting this back in draft for a bit, since I think the advent of blur has given me a good bit more to test with this.

PandorasFox:
here's where things are presently:
- all the foreground blurs work happily with the constantly-changing background layer :)
- adding a new window to the last workspace causes an instant shift from niri adding a new workspace after it. when closing those windows, the extra empty workspace isn't popped until after the workspace switch event, leading to some mild resettling
- frame or two of blank void visible after unlock/relock due to resurfacing
  - could perhaps tighten this to only after display reseats
  - could maybe also try to make the new surface before destroying the old, to try and avoid the one frame of void upon unlock?

https://github.com/user-attachments/assets/ef970cfe-c1ef-41f5-92e8-a1ea95ddd1c9

overall it works reasonably well, but the re-surfacing bits to survive output reseats could perhaps be polished a bit more - i recall seeing a release note a while ago about something similar for the bar re-surfacing itself to survive output reseats, so I can look into standardizing this a bit more?

the other note is that wallpaper switching blanks to void while loading rather than doing a transition effect; I figure there's room for bridging that w/ the state the lockscreen uses (+ not killing the old canvas immediately? haven't dug into that one still)

PandorasFox:
Rebased onto all the recenter render-loop changes and the diff is much cleaner now :)

quickshell/Modules/Lock/LockScreenContent.qml
```diff
@@ -201,20 +201,50 @@ Item {
...
+    }
+
+    ShaderEffectSource {
+        id: srcWallpaper
+        sourceItem: wallpaperSource
+        hideSource: true
+        live: false
+    }
+
+    ShaderEffect {
```
bbedward: This shader should be gated in a loader

bbedward (changes requested):
I have some conerns about the animation driver change, but I don't think there's any negative implications from what I can see, potentially on 60Hz or less displays.

I'd like to see if theres ways to make the Qt.callLater workarounds a little less prevant, and generallyl gate off things when the wallpaper isnt in scroll mode.

PandorasFox:
> I have some conerns about the animation driver change, but I don't think there's any negative implications from what I can see, potentially on 60Hz or less displays.
> 
> I'd like to see if theres ways to make the Qt.callLater workarounds a little less prevant, and generallyl gate off things when the wallpaper isnt in scroll mode.

Yeah, the main concern is (I think) battery devices are supposed to throttle to below 60Hz when on battery. Documentation was unclear & all my Linux boxes are desktops/servers now, so I couldn't really test that dimension - defaulting it to 0 and noting it as an option for high refresh rate panels/workstations would be totally fine imo, I just defaulted it to 1 if unset for my lazy testing purposes :)

Thanks! Glad to get this cleaned up and merged finally 😄

Issue #1663, opened 2026-02-13
Language select option


Ali-SZ:
### Feature Description

New option in settings to choose the language either from system locale or another language that is supported by DMS.

### Use Case

It will allow users to use a language for DMS different from their locale and users will be able to change it easily without the env variable change.

### Compositor(s)

- [x] All compositors
- [ ] Niri
- [ ] Hyprland
- [ ] MangoWC (dwl)
- [ ] Sway
- [ ] Other (specify below)

PR #1817, opened 2026-02-24
feat: add setting to change and hotreload locale

Silzinc:
I essentially extended the I18n service to manage the locale, and replaced pretty much every call of `Qt.locale` by `I18n.locale`. This should answer #1663.

### DIFF
--- a/quickshell/Modules/ControlCenter/BuiltinPlugins/CupsWidget.qml
+++ b/quickshell/Modules/ControlCenter/BuiltinPlugins/CupsWidget.qml
@@ -288,7 +288,7 @@ PluginComponent {
                                         StyledText {
                                             text: {
                                                 var date = new Date(modelData.timeCreated);
-                                                return date.toLocaleString(I18n.locale(), Locale.ShortFormat);
+                                                return date.toLocaleString(Qt.locale(), Locale.ShortFormat);
                                             }
                                             font.pixelSize: Theme.fontSizeSmall
                                             color: Theme.surfaceTextMedium
--- a/quickshell/Modules/DankDash/Overview/CalendarOverviewCard.qml
+++ b/quickshell/Modules/DankDash/Overview/CalendarOverviewCard.qml
@@ -15,7 +15,7 @@ Rectangle {
     signal closeDash
 
     function weekStartJs() {
-        return I18n.locale().firstDayOfWeek % 7;
+        return Qt.locale().firstDayOfWeek % 7;
     }
 
     function startOfWeek(dateObj) {
@@ -223,11 +223,10 @@ Rectangle {
             Repeater {
                 model: {
                     const days = [];
-                    const loc = I18n.locale();
-                    const qtFirst = loc.firstDayOfWeek;
+                    const qtFirst = Qt.locale().firstDayOfWeek;
                     for (let i = 0; i < 7; ++i) {
                         const qtDay = ((qtFirst - 1 + i) % 7) + 1;
-                        days.push(loc.dayName(qtDay, Locale.ShortFormat));
+                        days.push(I18n.locale().dayName(qtDay, Locale.ShortFormat));
                     }
                     return days;
                 }
--- a/quickshell/Modules/DankDash/WeatherForecastCard.qml
+++ b/quickshell/Modules/DankDash/WeatherForecastCard.qml
@@ -28,7 +28,7 @@ Rectangle {
         try {
             const date = new Date(root.forecastData.rawTime);
             const format = SettingsData.use24HourClock ? "HH:mm" : "h:mm AP";
-            return date.toLocaleTimeString(I18n.locale(), format);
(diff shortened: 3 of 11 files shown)


Silzinc:
Just realised that it would be nice to store the locale in settings. Right now, it still loads the system's default on startup. I will have a look at it soon enough.

Silzinc:
Locale is now a regular setting. It is `""` by default. If it is `""` and if the system locale is set, then it will be set to the system's locale. Otherwise, or if the setting leads to an invalid/unsupported locale, it is left unchanged and the interface is in English. If it is a valid language, it is loaded, with a priority over the system's locale.

quickshell/translations/en.json
```diff
@@ -149,12 +149,6 @@
     "reference": "Modules/Settings/NotificationsTab.qml:819, Modules/Settings/NotificationsTab.qml:832, Modules/Settings/NotificationsTab.qml:837, Modules/Settings/ClipboardTab.qml:103",
     "comment": ""
   },
-  {
-    "term": "1 device connected",
-    "context": "KDE Connect status single device",
```
bbedward, line 154: Why is this file so modified? What I do for local dev is clone `dms-plugins` into `quickshell` (which is gitignored), since its easier for local development. So all of the terms from official plugins get picked up by the script, but since you haven't it cloned I presume, it deleted all of the terms from 1st party plugins

quickshell/Common/I18n.qml
```diff
@@ -34,8 +35,10 @@ Singleton {
         showDirs: false
         showDotAndDotDot: false
 
-        onStatusChanged: if (status === FolderListModel.Ready)
-            root._pickTranslation()
+        onStatusChanged: if (status === FolderListModel.Ready) {
+            root._loadPresentLocales();
+            root._pickTranslation();
```
bbedward, line 40: This runs on startup now, after loading the files, but since pickTranslation writes to SettingsData, don't that break the auto/system locale case in future startups?

Silzinc: There is certainly redundancy in writing to SettingsData here in some cases. 

The typical behaviour of the component would be, on the first startup (and because there is no preferred locale yet in the settings), to pick the system's locale and write it in the settings as the preferred locale. On subsequent startups, it will try to read the preferred locale before looking at the system's one (see how `_rawLocale` is defined).

The machinery of looking at the system's or the preferred locale could be done in `_pickTranslation` entirely I guess, since this is only done at startup.

quickshell/translations/en.json
Silzinc: Oops I did not take enough care there. I should have noticed that something was wrong with the amount of deletions. I will try to correct this.

Silzinc: I did regenerate the translations after cloning the official plugins like you said. That's still a lot of unecessary changes, I hope I did not break anything.

quickshell/Common/settings/SettingsSpec.js
```diff
@@ -32,6 +32,7 @@ var SPEC = {
     mangoLayoutRadiusOverride: { def: -1, onChange: "updateCompositorLayout" },
     mangoLayoutBorderSize: { def: -1, onChange: "updateCompositorLayout" },
 
+    locale: { def: "", onChange: "updateLocale" },
```
bbedward, line 35: The other thing I guess, that might be valuable, is different overrides for different values. E.g. i may want LANG=en , but keeping LC_TIME (or LC_ALL) to be DE/FR or whatever. Right now this also is applying the same locale everywhere (language, calendar, and so on)

bbedward (changes requested):

quickshell/Common/settings/SettingsSpec.js
Silzinc: Definitely valuable, but it will require more work before merging. I think some settings like unit system or time display format already exist, maybe they could just be put in the `locale` object and their associated settings UIs in the locale tab.

bbedward: At least as a quicker thing I'd prefer the override to only affect language specific things probably, since that's the more common use case I think (e.g. a European or Asian wants to use English language)

Silzinc: If I understand correctly, that's essentially replacing I18n.locale back by Qt.locale in most places where it used, right? That way I18n.tr still uses the chosen language, but the rest uses the system's locale.

bbedward: Yea, I think that makes the most sense, and satisfies the most common use case (wanting a different language but not different locale for everything)

Silzinc: I would still want to use the saved locale for certain things like day names. I have tried to pick places where only digits would be displayed to use `Qt.locale`, but left `I18n.locale` on the rest. It would feel weird to see your UI in English but your month and day names in Hungarian even though you explicitly chose English, imo.

Issue #1607, opened 2026-02-06
Question: Simulating a screensaver


augustocdias:
### Compositor

- [ ] Niri
- [x] Hyprland
- [ ] MangoWC (dwl)
- [ ] Sway
- [ ] Other (specify below)

### Distribution

NixOS

### Description

I'm trying to simulate a screensaver by having `mpv` running a video in fullscreen but I don't know how to make it be on top of the lock screen. Would that be even possible? 

What I tried is basically calling locking the session with `loginctl` and calling a wrapper that opens `mpv` and stops it on any input.

bbedward:
Not possible, I'm not sure if there's another wayland protocol I don't know about or somethiing - but screensaver-while-locked is something that'd have to be integrated into DMS lock screen itself since its the locking surface. 

Need some research to see how it's done on wayland really, maybe some other lock screens do it.

augustocdias:
@bbedward I'm am thinking on trying to implement some sort of overlay on top of the lock screen playing the screensaver. Would that be possible with a plugin? I think no right? If no, is that a feature you'd be interested in the project? I'd wanna try to make a PR in that case.

PR #1819, opened 2026-02-24
feat(lockscreen): enable use of videos as screensaver in the lock screen

augustocdias:
I've asked in #1607 about the possibility to create screensavers and I decided to try out to implement it as an overlay on top of the lockscreen. 

I've created a new section in the lock screen settings page to set it up. When enabled the user can select a video or a directory. There's also a toggle for automatic cycling, which will pick a random video on the folder each time the screensaver is going to be shown. The behavior for this is as follows:

* When video path is a directory: a random video will be picked each time independently if the automatic cycling is enabled or not
* When video path is a video and automatic cycling is disabled: the selected video will be played every time
* When video path is a video and automatic cycling is enabled: a random video will be picked from the same directory as the selected file
* When video path points to a directory with no videos or non existing file: no screensaver is played and lock screen is shown as usual

Settings screen:
image
image

Any keypress or mouse movement cause the video to stop and the login screen to be shown. If the user goes idle in the login screen it will start the screensaver again. 

Question: should I run the script to generate the i18n files or this is ran by you/CI?

Disclaimer on the use of LLMs: I didn't generate the code with it, but I've used it heavily to learn about quickshell (this is my first time implementing something with it so code might not be the best either) and your codebase. I've used it to search for patterns to replicate like the wallpaper cycling logic and replicate in here as well as components to use.

### DIFF
--- a/quickshell/Modules/Lock/VideoScreensaver.qml
+++ b/quickshell/Modules/Lock/VideoScreensaver.qml
@@ -1,7 +1,6 @@
 pragma ComponentBehavior: Bound
 
 import QtQuick
-import QtMultimedia
 import Quickshell.Io
 import qs.Common
 import qs.Services
@@ -15,6 +14,7 @@ Item {
     property bool inputEnabled: false
     property point lastMousePos: Qt.point(-1, -1)
     property bool mouseInitialized: false
+    property var videoPlayer: null
 
     signal dismissed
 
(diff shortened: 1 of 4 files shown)


quickshell/Modules/Lock/VideoScreensaver.qml
```diff
@@ -0,0 +1,168 @@
+pragma ComponentBehavior: Bound
+
+import QtQuick
+import QtMultimedia
```
bbedward: We have to do some creative gymnastics here to keep QtMultimedia as an optional dependency, its done in AudioService - which can be re-used `AudioService.soundsAvailable`

But then additionally this entire component has to be created at runtime to avoid `import QtMultimedia`, since it will break compilation/runtime if qtmultimedia isnt installed.

That's done w ith like `Qt.createQmlObject`, also similar to AudioService

```qml
            powerUnplugSound = Qt.createQmlObject(`
                import QtQuick
                import QtMultimedia
                MediaPlayer {
                    source: "${powerUnplugPath}"
                    audioOutput: AudioOutput {
                        ${deviceProperty}volume: notificationsVolume
                    }
                }
            `, root, "AudioService.PowerUnplugSound");
```

augustocdias: I've created a `MultimediaService` and refactored the `AudioService.soundsAvailable` to bind to it. I hope that's okay.

PR #1828, opened 2026-02-25
matugen/template: Pasterizing neovim.

rPlakama:
The original template start burning my eyes... Anyway, its more an opinion than anything, so check yourself:

- New template:

image
image
image

- Old template (As reference.)

image
image
image

### DIFF
--- a/quickshell/matugen/templates/neovim.lua
+++ b/quickshell/matugen/templates/neovim.lua
@@ -4,25 +4,23 @@ return {
 		priority = 1000,
 		config = function()
 			require('base16-colorscheme').setup({
-				base00 = '{{colors.background.default.hex}}',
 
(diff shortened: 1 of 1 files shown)


bbedward:
It looks too bland and lacking contrast with only matugen colors, which is a problem for  text editors. I think a better approach would be to flip specific color values, like the green position in base16 is probably the one you have the most problem with, etc.

rPlakama:
> It looks too bland and lacking contrast with only matugen colors, which is a problem for text editors. I think a better approach would be to flip specific color values, like the green position in base16 is probably the one you have the most problem with, etc.

Alright, I'll try to use some of dank16's arguments for extra contrast

rPlakama:
I picked the colors from vscode template instead. 
image
image

PR #1842, opened 2026-02-26
feat: Add FIDO2/U2F security key support for lock screen

### LINKED ISSUE #1841: FIDO2/U2F security key support for lock screen
### Feature Description

Add FIDO2/U2F hardware security key (e.g. YubiKey) support for lock screen authentication, with two configurable modes: Alternative (OR) where the key works as a standalone unlock method racing alongside password/fingerprint, and Second Factor (AND) where password or fingerprint is required first followed by a security key touch.

### Use Case

Hardware security keys like YubiKey provide strong phishing-resistant authentication. Users who already have pam_u2f configured and keys enrolled should be able to use them on the DMS lock screen — either as a convenient tap-to-unlock alternative or as a mandatory second factor for higher security.

### Compositor(s)

All compositors

### If Other, please specify

_No response_

### Proposed Solution

- New PamContext for U2F in Pam.qml with state machine (idle → insert → waiting → success/retry)
- Bundled PAM config with support for system-level /etc/pam.d/dankshell-u2f override
- Availability detection for pam_u2f.so and ~/.config/Yubico/u2f_keys
- Settings toggle and mode dropdown in Lock Screen tab
- Lock screen UI feedback: passkey icon, placeholder text, status messages
- Input blocked (enter button, OSK) during U2F pending state in AND mode

SDDM/GDM support PAM-level U2F but with no user-facing feedback. swaylock has no U2F support. This provides visual feedback and configurability in DMS settings.

xPathin:
## Summary

Closes #1841

- Adds FIDO2/U2F hardware security key authentication for the lock screen with two modes: Alternative (OR) and Second Factor (AND)
- Settings UI with toggle, enrollment detection, and mode dropdown in Lock Screen tab
- Lock screen UI feedback: passkey icon, placeholder/status text, input blocking during U2F phase
- Bundled PAM config with `/etc/pam.d/dankshell-u2f` override support
- Fixes `persist: false` properties (e.g. `fprintdAvailable`) being reset to defaults on settings file reload

## Test plan

- [x] Enable U2F in settings with an enrolled YubiKey, verify OR mode unlocks on key touch
- [x] Switch to AND mode, verify password/fingerprint is required first, then key touch
- [x] Remove YubiKey, verify "Insert your security key..." prompt and retry on re-insert
- [x] Verify "Not enrolled" shown when pam_u2f.so or u2f_keys missing
- [x] Verify fingerprint/U2F availability persists after changing any setting
- [x] Verify OSK and enter button are hidden during U2F pending state

### DIFF
--- a/quickshell/Common/settings/Processes.qml
+++ b/quickshell/Common/settings/Processes.qml
@@ -62,7 +62,7 @@ Singleton {
     }
 
     property var u2fDetectionProcess: Process {
-        command: ["sh", "-c", "(test -f /usr/lib/security/pam_u2f.so || test -f /usr/lib64/security/pam_u2f.so) && test -f \"$HOME/.config/Yubico/u2f_keys\""]
+        command: ["sh", "-c", "(test -f /usr/lib/security/pam_u2f.so || test -f /usr/lib64/security/pam_u2f.so) && (test -f /etc/pam.d/dankshell-u2f || test -f \"$HOME/.config/Yubico/u2f_keys\")"]
         running: false
         onExited: function (exitCode) {
             if (!settingsRoot)
--- a/quickshell/Modules/Lock/Pam.qml
+++ b/quickshell/Modules/Lock/Pam.qml
@@ -251,7 +251,7 @@ Scope {
     Process {
         id: u2fAvailProc
 
-        command: ["sh", "-c", "(test -f /usr/lib/security/pam_u2f.so || test -f /usr/lib64/security/pam_u2f.so) && test -f \"$HOME/.config/Yubico/u2f_keys\""]
+        command: ["sh", "-c", "(test -f /usr/lib/security/pam_u2f.so || test -f /usr/lib64/security/pam_u2f.so) && (test -f /etc/pam.d/dankshell-u2f || test -f \"$HOME/.config/Yubico/u2f_keys\")"]
         onExited: code => {
             u2f.available = code === 0;
             u2f.checkAvail();


kulmin:
Screenshots from settings:

Screenshot from 2026-02-26 12-43-46
Screenshot from 2026-02-26 12-44-44

kulmin:
**Custom PAM config support:** Users with non-default key paths, custom origin/appid can create `/etc/pam.d/dankshell-u2f` with their own `pam_u2f.so` configuration. DMS will use it instead of the bundled default and skip the default `~/.config/Yubico/u2f_keys` path check for availability detection.

augustocdias:
Would this work on the greeter as well?

kulmin:
> Would this work on the greeter as well?

Lock screen only, the greeter runs through greetd which owns the PAM session, so there's no way to run a separate PamContext for fingerprint or U2F from the greeter side. If you add pam_u2f.so to /etc/pam.d/greetd it would work at the PAM level, this is what I do on my system btw.

tino376dev:
> > Would this work on the greeter as well?
> 
> Lock screen only, the greeter runs through greetd which owns the PAM session, so there's no way to run a separate PamContext for fingerprint or U2F from the greeter side. If you add pam_u2f.so to /etc/pam.d/greetd it would work at the PAM level, this is what I do on my system btw.

This also works for me, but there is no user feedback in greetd. I simply press enter (prompt asks for a password), and then tough my security key when the greeter says "authorizing". Is there a cleaner solution such that greetd prompts to touch the security key directly?

HirschBerge:
There is something about the detection that is broken. My security key (yubikey fido U2f) was working flawlessly with DMS's lock until this update broke it.
I've tried forcing the enableU2f to true but while it enables the switch, it doesn't actually correctly identify that I have the key.... 

Device works in:

- [X] Hyprlock
- [X]  Sudo
- [X]  tty login
- [X]  pkexec
- [X]  dank greeter
- [ ] DMS lock screen

Additional info that might be useful. Running on NixOS. Does there need to be an additional dependency added to the DMS flake?

augustocdias:
the detection seems to work in a very specific setup. I have nixos for example and the pam_u2f.so file is located somewhere else. Part of the detection is hardcoded to Yubikey and I believe other brands don't create the Yubico config folder. I believe the best course of action would be to have the "detection" be configurable and using the current behavior as a default.

purian23:
Hey everyone, we just pushed a refactor as well to hopefully help some of the woes when using ext devices/fingerprint auth etc in 32d16d067364e3181922bba5f5a5a4f2b0515a7b - we'll consolidate some of the greeter issues to ensure we're on the same page.

Issue #1846, opened 2026-02-26
DankBar calendar widget: add setting to make Monday the first day of the week


fdmarcin:
### Feature Description

I would like to be able to set the calendar widget to show the calendar with Monday as the first day of the week.

It's possible that currently it reads off my locale, which is en-US.

### Use Case

Users in multiple countries prefer using Monday or Saturday as the first day of the week:

<img src="https://vividmaps.com/wp-content/uploads/2023/06/First-Day-of-Week-World-Map.jpg" alt="world map showing preferred first of the week" />

### Compositor(s)

All compositors

### Proposed Solution

Add a setting to Settings to select the first day of the week among Monday, Saturday, and Sunday.

### Alternatives/Existing Solutions

- Windows 10:
  
  <img width="615" height="344" alt="Image" src="https://github.com/user-attachments/assets/7a436cca-1e9b-4b4c-b27c-0b36e3ee890a" />

- GNOME: you have to change the locale or use a GNOME extension

lakapple:
I think this should be closed by now

fdmarcin:
Thanks @Silzinc!

PR #1854, opened 2026-02-27
feat: add setting for first day of the week

Silzinc:
Addresses #1846. The setting `firstDayOfWeek` is an index ranging from 0 (Sunday) to 6 (Saturday), following the JS and QML Date APIs. It is set to `-1` by default. There are two places where it is used so far, where a function `weekStartQt()` is defined to retrieve that index, and if it is not in the valid range, set it using the system's locale. I admit that it does sound weird to lazily set it to the system's locale rather than on startup. I don't know how such a startup routine would be done in DMS.

I have not generated translations this time. Even after freshly cloning dms plugin and syncing my fork, I still have a bunch of changed to the translation files, so I'd rather let this be done by someone else's careful hands.

### DIFF
--- a/quickshell/Modules/DankDash/Overview/CalendarOverviewCard.qml
+++ b/quickshell/Modules/DankDash/Overview/CalendarOverviewCard.qml
@@ -16,7 +16,7 @@ Rectangle {
 
     function weekStartQt() {
         if (SettingsData.firstDayOfWeek >= 7 || SettingsData.firstDayOfWeek < 0) {
-            SettingsData.set("firstDayOfWeek", Qt.locale().firstDayOfWeek);
+            return Qt.locale().firstDayOfWeek;
         }
         return SettingsData.firstDayOfWeek;
     }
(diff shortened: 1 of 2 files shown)


quickshell/Modules/DankDash/Overview/CalendarOverviewCard.qml
```diff
@@ -14,8 +14,15 @@ Rectangle {
 
     signal closeDash
 
+    function weekStartQt() {
+        if (SettingsData.firstDayOfWeek >= 7 || SettingsData.firstDayOfWeek < 0) {
+            SettingsData.set("firstDayOfWeek", Qt.locale().firstDayOfWeek);
```
bbedward: We shouldnt set this here, if its unset or set to a system default  state we should just return the Qt.locale() one. This way would make it so it  only reads the locale once and then never again.

Silzinc: Should be good now I think

PR #1856, opened 2026-02-27
Add GeoClue2 integration as alternative to IP location

SuNNjek:
# Motivation
The weather widget never worked for me, presumably because (for some reason) the URL that is called to determine the user location doesn't load for me.

What does work however is [GeoClue2](https://gitlab.freedesktop.org/geoclue/geoclue/-/wikis/home), a background service that can determine your location using a variety of sources (e.g. Wifi or device GPS) and that can be queried using DBus. To me, this also seems a bit nicer than calling a web service.

# Proposed changes

Determine IP/auto location in for weather/gamma/theme by using GeoClue2. If it's not found/running, it will fall back to the old IP approach. For this I extended the Go backend with a new manager that communicates with GeoClue via DBus and added a new LocationService that is in turn used by the WeatherService in the Quickshell code. I tested it locally and it seems to work fine.


quickshell/Services/WeatherService.qml
```diff
@@ -511,8 +511,8 @@ Singleton {
         cityGeocodeFetcher.running = true;
     }
 
-    function getLocationFromIP() {
-        ipLocationFetcher.running = true;
+    function getLocationFromService() {
+        getLocationFromCoords(LocationService.latitude, LocationService.longitude);
```
bbedward, line 515: We should probably guard this function against 0, 0 as coords so it doesnt make a call with bad location data, otherwise the PR looks good. Right now it guards against NaN but since LocationService is initializing to 0, 0 now it wont guard against that.

Issue #1831, opened 2026-02-25
NixOS Go vendor hash mismatch


SuNNjek:
### Compositor

Hyprland

### Distribution

NixOS

### Select your Installation Method

Source

### Was this your original Installation method?

Yes

### dms doctor -vC

Not applicable here.

### Description

The latest release has an outdated vendor hash for the Go modules, causing the build to fail on NixOS.

### Expected Behavior

The build completes successfully.

### Steps to Reproduce

1. Have a NixOS configuration that uses dms (`github:AvengeMedia/DankMaterialShell/stable`)
2. Update the dms flake input to the latest version
3. Try and build the configuration
4. The build fails with the given error message

### Error Messages/Logs

Output of `nh os switch`:

```
┃ error: Cannot build '/nix/store/bp920yy21i8chqqjyb6l8mrysj7zkdbz-dms-shell-1.4.3+date=2026-02-25_3f24cf3.drv'.
┃        Reason: 1 dependency failed.
┃        Output paths:
┃          /nix/store/a3kl69ff6hfxs7acsmwj1dj6hvvzdinf-dms-shell-1.4.3+date=2026-02-25_3f24cf3
┃ error: hash mismatch in fixed-output derivation '/nix/store/pd8hzvj304gi0fxfjh2s032zbh99sjbs-dms-shell-1.4.3+date=2026-02-25_3f24cf3-go-modules.drv':
┃          specified: sha256-dEk7IOd6aQwaxZruxQclN7TGMyb8EJOl6NBWRsoZ9HQ=
┃             got:    sha256-cVUJXgzYMRSM0od1xzDVkMTdxHu3OIQX2bQ8AJbGQ1Q=
```

LuckShiba:
Fixed in https://github.com/AvengeMedia/DankMaterialShell/commit/47b12d09fc8526f9c231de60848a41b5990b4a37

PR #1864, opened 2026-02-27
nix: add `package` option for `dms-shell`

ilkecan:
... to make it configurable when using nix modules.

As an example, this would allow one to overcome #1831 with:
```nix
  programs.dank-material-shell = {
    enable = true;
    package = inputs.dms.packages.x86_64-linux.dms-shell.overrideAttrs { vendorHash = "sha256-cVUJXgzYMRSM0od1xzDVkMTdxHu3OIQX2bQ8AJbGQ1Q="; };
  }
```

when using [v1.4.3](https://github.com/AvengeMedia/DankMaterialShell/releases/tag/v1.4.3).

https://github.com/ilkecan/config/commit/f1bf1eecfd5b772abbe3d1a65ceaca7336ea5d89

### DIFF
--- a/distro/nix/common.nix
+++ b/distro/nix/common.nix
@@ -2,7 +2,6 @@
   config,
   lib,
   pkgs,
-  dmsPkgs,
   ...
 }:
 let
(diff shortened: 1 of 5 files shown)


LuckShiba:
I tested locally and the changes seem to build just fine :) 
Thanks!

LuckShiba:
Note that the `stable` branch (that tracks the latest released version) already got this fixed while we don't get the 1.4.4 release!

PR #1875, opened 2026-03-01
feat(tailscale): add Tailscale control center widget

Gdetrane:
## Summary

Native Tailscale integration as a builtin control center plugin — search, filter,
and copy IPs or FQDNs of any node in the tailnet without opening the web admin
panel or using the CLI.

## Motivation

Tailscale is a rapidly growing mesh VPN with a vibrant community, now shipping by
default in distros like Bazzite and other Fedora variants. Existing desktop
integrations (e.g., GNOME extensions) are limited and unreliable. This widget
provides quick, native access to tailnet device information directly from the DMS
control center.

### Go backend (core/internal/server/tailscale/)
- **Official Tailscale client** — uses `tailscale.com/client/local` (v1.96.1) for
  automatic socket discovery and typed API responses, with 3s context-based poll
  timeout and 30s poll interval
- **Status conversion** — converts `ipnstate.Status` into lean IPC types with
  peer sorting (online first, alphabetical), DNS-derived hostnames (unique per
  node), user ID resolution, and relative time formatting
- **IPC methods** — tailscale.subscribe (streaming state updates),
  tailscale.getStatus (one-shot), tailscale.refresh (manual poll)
- **Daemon lifecycle handling** — eager init with 30s background retry when
  tailscaled is not running at startup; polls keep running through daemon
  stop/restart cycles for automatic recovery
- **17 unit tests** covering status conversion, peer parsing, sorting, manager
  lifecycle, poll errors, handler routing, and state comparison

### QML frontend
- **TailscaleService** singleton with ref counting and subscription management
  following the CupsService pattern
- **TailscaleWidget** PluginComponent with control center tile (device_hub icon,
  online peer count) and expandable detail view
- **Filter chips** — My Online / Online / All with live device counts; filters
  compose with search (tagged devices correctly included in My Online)
- **Searchable device list** — match on hostname, DNS name, IP, and OS; search
  applies within the active filter
- **Copy-to-clipboard** buttons for Tailscale IPs and DNS names
- **Self device** highlighted with accent tint and This device badge
- **Expandable cards** showing DNS name, tags, owner, relay/direct info
- All strings use I18n.tr() with translator context per CONTRIBUTING.md

### Integration
- Widget registered in WidgetModel.qml with DragDropGrid and DetailHost wiring
- DMSService.qml signal added for tailscale state updates
- DetailHost.qml height set to 350px (matching Bluetooth/VPN)

### Dependencies
- Add `tailscale.com` v1.96.1 (official local API client, ~22 transitive deps, pure Go)
- Bump Go to 1.26.1 (required by tailscale.com)
- Bump golangci-lint hook v2.9.0 → v2.11.3

## Daemon lifecycle handling
- Tailscale running at startup: widget shows immediately
- Tailscale started after DMS: widget appears within 30s (retry loop)
- Tailscale stopped while running: widget shows Disconnected within 30s
- Tailscale restarted: widget recovers automatically within 30s
- Tailscale not installed: widget shows Not available, no errors

## Test plan
- [x] Unit tests pass (`make test`)
- [x] Widget appears when Tailscale is running
- [x] Widget shows Not available when Tailscale is not installed
- [x] Filter chips filter correctly (My Online includes tagged devices)
- [x] Search composes with active filter
- [x] Copy buttons copy correct values to clipboard
- [x] Self device shows with accent highlight and This device badge
- [x] Expanding a card reveals DNS name, tags, owner
- [x] Daemon stop: widget shows Disconnected
- [x] Daemon restart: widget recovers automatically
- [x] Nodes with duplicate OS hostnames show unique DNS-derived names
- [x] DMS started without Tailscale, Tailscale started later: widget appears

**Tested on:**
- CachyOS (Arch), niri compositor, 57-node tailnet


purian23:
Hi there, thanks for the PR. I think this is better suited as an optional plug-in rather a native DMS feature as it's best fit for a select number of users. I will refer to @bbedward for a final accessment.

Gdetrane:
image

----

Let me know if you want to see more screenshots

Gdetrane:
@Purian23 Hi, thanks for jumping in, that's fair. I thought this could be a nice built-in feature, given Tailscale's growth and it being included in some distros, but I totally see your point. 

I'm still new to this repo, but I imagine this would require keeping the core Go logic and moving the QML logic to a plugin package, right? 

Any feedback is welcome

mecattaf:
I would 100% use that!

Gdetrane:
Thanks @mecattaf I really appreciate that :smile: 
I moved everything to Tailscale, it's brilliant and genuinely love it. It's exploding in popularity, I work at Red Hat and I know that in the Fedora world it's starting to get a lot of traction, see https://universal-blue.org/ 
If I'm not mistaken it's pre-installed in all their variants. One day I'd be very happy to see an official DMS flavor of Fedora btw :sunglasses:

mecattaf:
> Thanks @mecattaf I really appreciate that :smile: 
> I moved everything to Tailscale, it's brilliant and genuinely love it. It's exploding in popularity, I work at Red Hat and I know that in the Fedora world it's starting to get a lot of traction, see https://universal-blue.org/ 

Been a ublue user for 5+ years! Specifically have my own bluebuild recipe

> If I'm not mistaken it's pre-installed in all their variants. One day I'd be very happy to see an official DMS flavor of Fedora btw :sunglasses: 

Actually DMS ships with the miracle window manager official distro but that's super niche. Otherwise check out Zirconium (dms + niri)

purian23:
Thanks again for the forward thinking here! The main dev @bbedward is out this week but we'll review whether a plugin would work or by using it natively in a few days!

purian23:
Just another quick note on the plugin option route I mentioned. One already exists for you to see how it works. 
https://danklinux.com/plugins
Tailscale Manager
https://github.com/cglavin50/dms-tailscale

Gdetrane:
@Purian23 No worries at all, we all need time off :+1:  

Just a little note on the plugin route, I think this might be a bit too complex for a plugin if I understand this correctly, given there's some Go code to handle everything via Tailscale's Unix sockets API, which imho is cleaner than relying on the cli as a middleman. 

I'm not sure obviously, I don't exactly have a ton of experience with implementing tools using Unix sockets from my day to day work, but my intuition was in general to use the APIs that are natively provided to integrate with whatever the underlying system is

core/internal/server/tailscale/client.go
```diff
@@ -0,0 +1,208 @@
...
+	"io"
+	"net"
+	"net/http"
+	"sort"
+	"strings"
+	"time"
+)
+
+const (
+	defaultSocketPath = "/var/run/tailscale/tailscaled.sock"
```
bbedward: Is this correct? Shouldn't it have some lookup hierarchy or something, like XDG_RUNTIME_DIR, etc. ?

core/internal/server/tailscale/client.go
```diff
@@ -0,0 +1,208 @@
...
+	statusEndpoint    = "/localapi/v0/status"
+	fetchTimeout      = 10 * time.Second
```
bbedward: Feels much too long for a socket timeout

core/internal/server/server.go
```diff
@@ -63,6 +64,7 @@ var waylandManager *wayland.Manager
 var bluezManager *bluez.Manager
 var appPickerManager *apppicker.Manager
 var cupsManager *cups.Manager
+var tailscaleManager *tailscale.Manager
```
bbedward, line 69: Should handle cleanup/closing client in cleanupManagers

core/internal/server/tailscale/manager.go
```diff
@@ -0,0 +1,201 @@
...
+const pollInterval = 30 * time.Second
+
+// Manager manages Tailscale state polling and subscriber notifications.
+type Manager struct {
+	state       *TailscaleState
+	stateMutex  sync.RWMutex
+	subscribers syncmap.Map[string, chan TailscaleState]
+	stopChan    chan struct{}
+	pollWG      sync.WaitGroup
+	notifierWg  sync.WaitGroup
```
bbedward: Seems unused, nothing is ever added here

core/internal/server/tailscale/manager.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	if old == nil || new == nil {
+		return true
+	}
+
+	oldJSON, err := json.Marshal(old)
+	if err != nil {
+		return true
+	}
+	newJSON, err := json.Marshal(new)
```
bbedward: Not a big deal but json marshal is kinda heavy here in the retry loop, it would be more efficient to just compare bytes or a hash directly for state differences.

bbedward (changes requested):
This looks mostly good, and I think it's fine as a built-in, though its worth noting a plugin could interface with it directly since it is a unix socket API, which can be done in QML/JS

Gdetrane:
Hi @bbedward thank you for the review, I will happily address all the requested changes asap

bbedward:
@Gdetrane  Also consider using the tailscale GO client, it would probably simplify things quite a bit and help compatibility/maintenance long-term.  https://github.com/tailscale/tailscale/blob/main/client/local/local.go

Gdetrane:
@bbedward I noticed the Tailscale go client requires a very recent go version, 1.26+, considering there might be CI involved with 1.25, nix packages I haven't verified availability for, I think I will simply assess your feedback in code without changing the core implementation, but in the future maybe the tailscale go client will be the way to go (pun intended). 
I'll resume work on this ASAP, a bit busy at work this week, but the changes I believe will be minimal, based on your feedback

bbedward:
Bumping GO version is no big deal, CI should be unaffected since it pulls version from go.mod

Gdetrane:
@bbedward Cool, I will dive deeper into the library then, it doesn't seem to have a ton of dependencies either, so win win

Gdetrane:
@bbedward moved to the Tailscale go client and refactored based on it, I made some extra modifications fixing a couple of lingering issues I've noticed in manual testing done on my desktop. I believe your points should be addressed now, but ptal :) 
Rebased on upstream and updated the description to reflect the new changes while at it. 
One thing I should've asked, I'm used to squashing into a single commit and amending, that's a convention from my team at work, but let me know if you would prefer seeing multiple commits in the future

Gdetrane:
Noticed new conflicting changes merged upstream, rebased again

core/internal/server/server.go
```diff
@@ -1334,11 +1404,35 @@ func cleanupManagers() {
...
+	}
 }
 
 func Start(printDocs bool) error {
 	cleanupStaleSockets()
 
+	// Eagerly try to initialize Tailscale manager; if it fails (socket not
+	// found), start a background retry loop so the widget recovers when
+	// tailscaled starts later.
+	if err := InitializeTailscaleManager(); err != nil {
```
bbedward: This looks like it will never be triggered, because tailscale.NewManager() never returns an error.

core/internal/server/tailscale/manager.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+}
+
+// checkAndNotify compares the current state with the last notified state and broadcasts if changed.
+func (m *Manager) checkAndNotify() {
+	m.stateMutex.RLock()
+	current := m.state
+	m.stateMutex.RUnlock()
+
+	if !reflect.DeepEqual(m.lastState, current) {
```
bbedward: lastState RW should also be protected by mutex, or we could probably be more cautious with a notifier in a single goroutine pattern that might be more thread-safe, like some other managers do.

core/internal/server/server.go
```diff
@@ -250,6 +254,19 @@ func InitializeCupsManager() error {
...
 }
 
+func InitializeTailscaleManager() error {
+	manager, err := tailscale.NewManager("")
+	if err != nil {
+		log.Warnf("Failed to initialize tailscale manager: %v", err)
+		return err
+	}
+
+	tailscaleManager = manager
```
bbedward: tailScaleManager probably needs mutex protection, seems like it can be written from multiple goroutines.

core/internal/server/tailscale/handlers.go
```diff
@@ -0,0 +1,70 @@
...
+type TailscaleEvent struct {
+	Type string         `json:"type"`
+	Data TailscaleState `json:"data"`
+}
+
+// HandleRequest routes an IPC request to the appropriate handler.
+func HandleRequest(conn net.Conn, req models.Request, manager *Manager) {
+	switch req.Method {
+	case "tailscale.subscribe":
+		handleSubscribe(conn, req, manager)
```
bbedward: We can remove this case, since the pattern is to subscribe through the main server.go with what we want. Other managers may do it (but its not a pattern thats used)

core/internal/server/tailscale/manager.go
```diff
@@ -0,0 +1,169 @@
...
+	"sync"
+	"time"
+
+	"github.com/AvengeMedia/DankMaterialShell/core/internal/log"
+	"github.com/AvengeMedia/DankMaterialShell/core/pkg/syncmap"
+	"tailscale.com/client/local"
+)
+
+const (
+	pollInterval = 30 * time.Second
```
bbedward: I would really prefer to not do polling here, I think we can use `watch-ipn-bus` : https://github.com/tailscale/tailscale/blob/54606a0a89bc5e90b5295179e5b776f6a2df8cfa/ipn/localapi/localapi.go#L119  , that would be much nicer

bbedward: As a bonus without polling maybe we can avoid all of the `reflect.DeepEqual` usage. Which has a few risks:
1. Very slow and inefficient
2. May evaluate as not equal on fields you dont care about

core/internal/server/server.go
```diff
@@ -80,6 +82,8 @@ const dbusClientID = "dms-dbus-client"
 var capabilitySubscribers syncmap.Map[string, chan ServerInfo]
 var cupsSubscribers syncmap.Map[string, bool]
 var cupsSubscriberCount atomic.Int32
+var tailscaleSubscribers syncmap.Map[string, bool]
```
bbedward: We increment these but I can't tell that they actually matter or are used for anything. Ideally we would stop the retry loop when its 0 at least

quickshell/Services/TailscaleService.qml
```diff
@@ -0,0 +1,182 @@
...
+        }
+    }
+
+    property bool connected: false
+    property string version: ""
+    property string backendState: ""
+    property string magicDnsSuffix: ""
+    property string tailnetName: ""
+    property var selfNode: null
+    property var peers: []
```
bbedward, line 43: The way this is computed probably triggered an inefficient binding chain, but it's not a major deal for now.

bbedward (changes requested):

Gdetrane:
@bbedward taking some time to address all the issues, I'm a little swamped at work right now

HeroGamers:
Just saw this open PR, and wanted to add that I've been having great joy with https://github.com/DeedleFake/trayscale, which works really well - and has a lot of nice features already for exit nodes and such! It's also written in Go, and is also under the MIT License, so for this built-in module it could likely make sense to use it as inspiration for some of the more advanced CLI capabilities - and how to translate them into a UI :)

Gdetrane:
@bbedward sorry for keeping this open for so long, I had most of the changes ready, but some life circumstances didn't give me enough bandwidth to properly resume work, hopefully the context is fresh enough. I addressed your changes and verified no regressions in real usage, obviously rebased on the latest upstream as of now

TwinFox20:
I noticed a small UI inconsistency with the Tailscale widget.

On the first screenshot, the highlighted background color doesn't seem to follow the same color layering rules used throughout the rest of DMS. Compared to the second screenshot (which matches the rest of the UI), the Tailscale widget uses a noticeably brighter background instead of the darker surface color that similar components use.

It would be nice if the widget used the same surface/elevation colors as the other control center components to keep the appearance consistent.

*First screenshot: Tailscale widget*
*Second screenshot: another control center widget with the expected styling*

image

image

L0renz00:
Im having a similar issue as @TwinFox20. Picture attached.
image
As they said it would be nice if the widget used the same colors as the other control center widgets. 

Thanks for the great work on the widget!

bbedward:
The inconsistency should be fixed in 1.5.1+

PR #1878, opened 2026-03-01
fix(icons): apply appIdSubstitutions across all icon resolution paths

odtgit:
## Summary
Follow-up to #1867 and #1877. An audit of all icon resolution paths in the codebase found **7 additional components** that resolve icons via `Quickshell.iconPath()` or `image://icon/` URLs without checking `appIdSubstitutions` first.

This means file path substitutions (e.g. for PWA icons that Qt's icon theme system can't reliably resolve — see background below) only worked in the taskbar (#1867) and launcher (#1877), but not in notifications, dock context menus, or settings UI.

### Background: why substitutions are needed for PWAs

Qt6's `QIcon::fromTheme()` has a limitation where it silently resolves PWA icon names (e.g. `chrome-xxx-Default`) to the **generic browser icon** instead of the actual PWA-specific icon installed in `~/.local/share/icons/hicolor/`. This was confirmed via pixel-level comparison — the rendered pixmap is byte-identical to `google-chrome`. PWA icons that happen to be bundled in the active system theme (e.g. Qogir ships YouTube) work correctly; others don't. This is a Qt upstream issue that can't be fixed at the DMS level.

The `appIdSubstitutions` feature (Settings → Running Apps) lets users map these broken icon names to direct file paths, bypassing Qt's theme lookup entirely. This PR ensures that mapping works consistently everywhere icons are displayed.

### Zero cost for non-users

When no substitutions match, `Paths.moddedAppId()` returns the original icon name unchanged (identity), so the existing `Quickshell.iconPath()` / `image://icon/` path executes exactly as before. There is no behavioral change for users who don't configure substitutions.

### Audit results

| Component | File | Issue |
|-----------|------|-------|
| Notification card (grouped + expanded) | `NotificationCard.qml` | `Quickshell.iconPath(appIcon, true)` without substitution |
| Notification popup | `NotificationPopup.qml` | Same |
| Notification history | `HistoryNotificationCard.qml` | Same |
| Dock context menu | `DockContextMenu.qml` | Same |
| Bar dock context menu | `AppsDockContextMenu.qml` | Same |
| Launcher edit dialog | `LauncherContent.qml` | `"image://icon/" + icon` without substitution |
| Launcher settings | `LauncherTab.qml` (×3) | Same |

All now route through `Paths.moddedAppId()` before falling through to the default icon lookup, matching the pattern in `Paths.getAppIcon()` and `AppIconRenderer.iconPath`.

## Test plan
- Configure a regex appIdSubstitution mapping a PWA app_id to a file path
- Verify PWA icons display correctly in: taskbar, launcher, notifications, dock right-click menu, launcher settings
- Verify non-PWA apps are unaffected (no substitution match → original path unchanged)

### DIFF
--- a/quickshell/Modals/DankLauncherV2/LauncherContent.qml
+++ b/quickshell/Modals/DankLauncherV2/LauncherContent.qml
@@ -789,7 +789,19 @@ FocusScope {
                 Image {
                     width: 40
                     height: 40
-                    source: editingApp?.icon ? "image://icon/" + editingApp.icon : "image://icon/application-x-executable"
+                    source: {
+                        const icon = editingApp?.icon;
+                        if (!icon) return "image://icon/application-x-executable";
+                        const moddedIcon = Paths.moddedAppId(icon);
+                        if (moddedIcon !== icon) {
+                            if (moddedIcon.startsWith("~") || moddedIcon.startsWith("/"))
+                                return Paths.toFileUrl(Paths.expandTilde(moddedIcon));
+                            if (moddedIcon.startsWith("file://"))
+                                return moddedIcon;
+                            return "image://icon/" + moddedIcon;
+                        }
+                        return "image://icon/" + icon;
+                    }
                     sourceSize.width: 40
                     sourceSize.height: 40
                     fillMode: Image.PreserveAspectFit
--- a/quickshell/Modules/DankBar/Widgets/AppsDockContextMenu.qml
+++ b/quickshell/Modules/DankBar/Widgets/AppsDockContextMenu.qml
@@ -273,7 +273,18 @@ PanelWindow {
 
                             IconImage {
                                 anchors.fill: parent
-                                source: modelData.icon ? Quickshell.iconPath(modelData.icon, true) : ""
+                                source: {
+                                    if (!modelData.icon) return "";
+                                    const moddedIcon = Paths.moddedAppId(modelData.icon);
+                                    if (moddedIcon !== modelData.icon) {
+                                        if (moddedIcon.startsWith("~") || moddedIcon.startsWith("/"))
+                                            return Paths.toFileUrl(Paths.expandTilde(moddedIcon));
+                                        if (moddedIcon.startsWith("file://"))
+                                            return moddedIcon;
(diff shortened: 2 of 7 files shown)


odtgit:
### Why not fix Qt's icon theme lookup instead?

We did a deep investigation into why Qt6 can't resolve PWA icons from `~/.local/share/icons/hicolor/`. Here's what we found:

**Qt6's `QIcon::fromTheme()` silently resolves PWA icon names to the generic browser icon** rather than finding the actual PWA-specific icon files. Confirmed via pixel-level MD5 comparison — e.g. `chrome-comkdlimbkhemidbbpchhepidbmjpnhh-Default` renders byte-identical to `google-chrome`.

**What we ruled out:**
- **Not a GTK icon-theme.cache issue** — same behavior with and without the cache file
- **Not file permissions** — tested `644` vs `600`, identical results
- **Not non-deterministic** — consistently reproducible across fresh process invocations
- **Not the `isNull()` gate in quickshell** — `QIcon::fromTheme().isNull()` returns `false` for all PWA icons (Qt6 lazy-loads), so quickshell's check passes. But the icon that actually renders is the wrong one.

**Root cause:** Qt checks the active theme (e.g. Qogir) first. PWA icons that happen to be bundled in Qogir/Papirus (like YouTube) work. PWA icons that only exist in the user-local `~/.local/share/icons/hicolor/` are not reliably found by Qt's `QIconLoader` — it appears to not properly scan user-local hicolor directories when a system theme is active. For unfound names, the lazy `QIcon` object silently falls back to a generic match at render time.

**This is a Qt upstream issue** in `QIconLoader` that can't be fixed at the DMS level. The `appIdSubstitutions` approach bypasses Qt's theme system entirely by resolving directly to icon file paths on disk, which is exactly what this feature was designed for.

odtgit:
Closing in favor of #1880, which supersedes this PR. Instead of duplicating the inline moddedAppId pattern across 8 files, #1880 centralizes the logic into `Paths.resolveIconPath()` and `Paths.resolveIconUrl()` — all the same callsites are covered but with one-line calls instead of repeated 8-line blocks.

PR #1880, opened 2026-03-01
refactor(icons): centralize icon resolution into Paths.resolveIconPath/resolveIconUrl

odtgit:
## Summary
Supersedes #1878. An audit of all icon resolution paths found 8 components that resolve icons via `Quickshell.iconPath()` or `image://icon/` URLs without checking `appIdSubstitutions`. Rather than adding the same 8-line inline pattern to each file (#1878), this PR centralizes the logic into two new functions in `Paths.qml`:

- **`resolveIconPath(iconName)`** — for `Quickshell.iconPath()` callsites, handles substitution + file path detection + `DesktopService.resolveIconPath()` fallback
- **`resolveIconUrl(iconName)`** — for `image://icon/` URL callsites, same substitution logic

All consumer files now use one-line calls instead of duplicated inline blocks.

### Why substitutions are needed

Qt6's `QIcon::fromTheme()` has a limitation where it silently resolves some icon names (e.g. PWA icons like `chrome-xxx-Default`) to the generic browser icon instead of the actual icon installed in `~/.local/share/icons/hicolor/`. This was confirmed via pixel-level comparison. Icons bundled in the active system theme (e.g. Qogir ships YouTube) work; others don't. This is a Qt upstream issue. The `appIdSubstitutions` feature lets users map broken icon names to direct file paths, bypassing Qt's theme lookup.

### Zero cost for non-users

When no substitutions match, `moddedAppId()` returns the original icon name unchanged (identity), so the existing `Quickshell.iconPath()` / `image://icon/` path executes exactly as before.

### Files changed (9)

| File | Change |
|------|--------|
| `Common/Paths.qml` | Add `resolveIconPath()`, `resolveIconUrl()`, simplify `getAppIcon()` |
| `Widgets/AppIconRenderer.qml` | 8 lines inline → `Paths.resolveIconPath()` |
| `Notifications/Center/NotificationCard.qml` (×2) | `Quickshell.iconPath(appIcon, true)` → `Paths.resolveIconPath(appIcon)` |
| `Notifications/Popup/NotificationPopup.qml` | Same |
| `Notifications/Center/HistoryNotificationCard.qml` | Same |
| `Dock/DockContextMenu.qml` | Same |
| `DankBar/Widgets/AppsDockContextMenu.qml` | Same |
| `Modals/DankLauncherV2/LauncherContent.qml` | `"image://icon/" + icon` → `Paths.resolveIconUrl()` |
| `Modules/Settings/LauncherTab.qml` (×3) | Same |

## Test plan
- Configure regex appIdSubstitutions mapping PWA app_ids to file paths
- Verify PWA icons display correctly in: taskbar, launcher, dock right-click menus, launcher settings
- Verify non-PWA apps are unaffected (no substitution → original path unchanged)
- Verify fallback icons still work for apps with no icon

### DIFF
--- a/quickshell/Common/Paths.qml
+++ b/quickshell/Common/Paths.qml
@@ -71,19 +71,40 @@ Singleton {
         return appId;
     }
 
-    function getAppIcon(appId: string, desktopEntry: var): string {
-        if (appId === "org.quickshell") {
-            return Qt.resolvedUrl("../assets/danklogo.svg");
-        }
-
-        const moddedId = moddedAppId(appId);
-        if (moddedId !== appId) {
+    function resolveIconPath(iconName: string): string {
+        if (!iconName) return "";
+        const moddedId = moddedAppId(iconName);
+        if (moddedId !== iconName) {
             if (moddedId.startsWith("~") || moddedId.startsWith("/"))
                 return toFileUrl(expandTilde(moddedId));
             if (moddedId.startsWith("file://"))
                 return moddedId;
             return Quickshell.iconPath(moddedId, true);
         }
+        return Quickshell.iconPath(iconName, true) || DesktopService.resolveIconPath(iconName);
+    }
+
+    function resolveIconUrl(iconName: string): string {
+        if (!iconName) return "";
+        const moddedId = moddedAppId(iconName);
+        if (moddedId !== iconName) {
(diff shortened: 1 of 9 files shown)


purian23:
Thanks!

PR #1892, opened 2026-03-02
Add qtwebsockets to flake

djmaze:
This allows using the [HomeAssistant plugin](https://github.com/xxyangyoulin/dms-plugin-hass).

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -48,6 +48,7 @@
           sonnet
           qtmultimedia
           qtimageformats
+          qtwebsockets
           kimageformats
         ];
     in


LuckShiba:
I'm not sure if adding qtwebsockets to the Flake to all users is a good idea, since not all users will use that functionality and it's also not a dependency in other distros. What about we add in the package an option to pass additional packages, so the user can do the following:

```nix
programs.dankMaterialShell.package = dmsPkgs.dms-shell.override {
  extraQmlPackages = with pkgs; [ qtwebsockets ];
};
```

This would allow users to add other QML dependencies other plugins may require too!

djmaze:
@LuckShiba Sounds like a good idea! I can try to implement that.

LuckShiba:
@djmaze, I opened #1903 addressing that. I called the property `extraQtPackages`. It should solve your use case!  I added an example on the description. Closing this PR.

djmaze:
Thanks!

PR #1900, opened 2026-03-03
Not everyone uses paru or yay on Arch: Support pacman command

merdely:
Allow the System Update widget to work on Arch systems without paru or yay installed

### DIFF
--- a/quickshell/Services/SystemUpdateService.qml
+++ b/quickshell/Services/SystemUpdateService.qml
@@ -185,7 +185,7 @@ Singleton {
 
     Process {
         id: pkgManagerDetection
-        command: ["sh", "-c", "which which paru || which yay || which pacman || which dnf"]
+        command: ["sh", "-c", "which paru || which yay || which pacman || which dnf"]
 
         onExited: exitCode => {
             if (exitCode === 0) {


merdely:
I added code that uses sudo properly when using pacman.  In order to not duplicate the archBasedPMSettings structure, I turned it into a function that returns the structure and accepts an argument for requiresSudo

purian23:
Hi there, this entire system updater widget is set to get an overhaul in v1.5. Everything should be addressed at that time. 

Pacman should go to the end so it's like `paru || yay || pacman` so the code can still iterate the AUR helpers in order. If they are picked up, they tend to update the pacman system apps as well.

merdely:
I made your suggested change.  Though, should we just close this if the overhaul is coming soon?

quickshell/Services/SystemUpdateService.qml
```diff
@@ -182,7 +185,7 @@ Singleton {
 
     Process {
         id: pkgManagerDetection
-        command: ["sh", "-c", "which paru || which yay || which dnf"]
+        command: ["sh", "-c", "which which paru || which yay || which pacman || which dnf"]
```
purian23: `which` is listed twice here.

merdely:
I fixed the error.  Sorry about that

PR #1903, opened 2026-03-04
flake: allow extra QT packages in dms-shell package

LuckShiba:
This allows users to, optionally, add extra QT packages in dms-shell wrapper variables. This is useful for some QT packages some plugins may need, such as `qtwebsockets` in #1892. 

Example usage:
```nix
{
  programs.dank-material-shell.package = dmsPkgs.dms-shell.override {
    extraQtPackages = with pkgs.kdePackages; [ qtwebsockets ];
  };
}

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -72,76 +72,82 @@
             "${cleanVersion}${dateSuffix}${revSuffix}";
         in
         {
-          dms-shell = pkgs.buildGoModule (
-            let
-              rootSrc = ./.;
-            in
+          dms-shell = pkgs.lib.makeOverridable (
             {
-              inherit version;
-              pname = "dms-shell";
-              src = ./core;
-              vendorHash = "sha256-dEk7IOd6aQwaxZruxQclN7TGMyb8EJOl6NBWRsoZ9HQ=";
+              extraQtPackages ? [ ],
+            }:
+            pkgs.buildGoModule (
+              let
+                rootSrc = ./.;
+                qtPackages = (qmlPkgs pkgs) ++ extraQtPackages;
+              in
+              {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by djmaze** [comment]
Mhh.. Somehow with this approach, the HomeAssistant plugin again reports "Missing dependency". Not sure if this because I was using the stable version before.

**UPDATE**: No, was using the same DMS / Quickshell version of course. Will have a look what I am doing wrong.

**Comment by LuckShiba** [comment]
@djmaze, have you used the `kdePackages.qtwebsockets` and restarted your DMS after rebuilding? I just tested on my system, and adding qtwebsockets the same way as in the example in the description here got rid of the "Missing dependency" warning 😕 

Run `grep qtwebsockets $(which dms)` and tell me if it has the qtwebsockets paths on it, it should have.

**Comment by djmaze** [comment]
Yeah, it is in there, I already checked that before. Of course I also restarted dms. When switching back to my branch, it works again.

I realized though that with the setup from my branch it is qtwebsockets-6.10.2 while with the config I am using your branch with it is qtwebsockets-6.10.1. Though I tried to use nixos-unstable same as in the dms flake.

But I have to admit I am not a NixOS expert, so trying to find my way through this. My config inside looks like this (I have a flake-based NixOS setup):

```nix
  programs.dank-material-shell = {
    enable = true;
    enableSystemMonitoring = true;
    package = dms.packages.x86_64-linux.dms-shell.override {
      extraQtPackages = with nixos-unstable.kdePackages; [ qtwebsockets ];
    };
    dgop.package = pkgs-unstable.dgop;
  };
```

I probably need to find a way to fix this myself.

PR #1924, opened 2026-03-06
dock: restore Hyprland special workspace windows on click

lpv11:
It's a patch for the Dock in hyprland that comes with an on/off setting: when you click a window that is on a special workspace, restore the window to the working workspace instead of moving the screen to the special workspace.

### DIFF
--- a/quickshell/Common/settings/SettingsSpec.js
+++ b/quickshell/Common/settings/SettingsSpec.js
@@ -294,6 +294,7 @@ var SPEC = {
     dockAutoHide: { def: false },
     dockSmartAutoHide: { def: false },
     dockGroupByApp: { def: false },
+    dockRestoreSpecialWorkspaceOnClick: { def: false },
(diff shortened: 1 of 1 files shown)


quickshell/Common/SettingsData.qml
```diff
@@ -479,6 +479,7 @@ Singleton {
     property bool dockAutoHide: false
     property bool dockSmartAutoHide: false
     property bool dockGroupByApp: false
+    property bool dockRestoreSpecialWorkspaceOnClick: false
```
bbedward, line 482: This should also be set in SettingsSpec.js

lpv11: Added it. You can make any changes to it you want obviously. I'm just suggesting that this should exist at least. In my hyprland setup I use special workspaces a lot so I don't close windows and don't have to reopen them - the main apps at least - so it's very helpful to me :)

PR #1954, opened 2026-03-10
feat: Add Zed editor theming support

Adarsh219:


quickshell/matugen/configs/zed.toml
```diff
@@ -0,0 +1,3 @@
+[templates.dmszed]
+input_path = "SHELL_DIR/matugen/templates/dank-zed.json"
+output_path = "~/.config/zed/themes/dank-zed-theme.json"
```
bbedward: This should be `CONFIG_DIR/zed/themes/dank-zed-theme.json`

bbedward (changes requested):

quickshell/matugen/configs/zed.toml
Adarsh219: Fixed it

cemoktra:
For Rust in Zed, the generated "Everforest" theme is unusable using this configuration. Most keywords like `let`, `impl` are hardly to distinguish from the background.

PR #1957, opened 2026-03-10
feat: highlight active workspace app tiles

0xdeafbeef:
image

image

Adds outline for active column. I can make a follow-up to make it toggleable  from settings

### DIFF
--- a/quickshell/Common/SettingsData.qml
+++ b/quickshell/Common/SettingsData.qml
@@ -280,6 +280,7 @@ Singleton {
     property bool showOccupiedWorkspacesOnly: false
     property bool reverseScrolling: false
     property bool dwlShowAllTags: false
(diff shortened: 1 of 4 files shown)


purian23:
Hi there, thanks for the PR, but we definitely don't want this enabled and default only options. A toggle would have to be implemented here to not alter our current defaults.

0xdeafbeef:
image

0xdeafbeef:
@Purian23 done

0xdeafbeef:
Also i've made it because icons weren't highlighted without icon pack. So mb we should require icon packs

purian23:
The only concern I have is with dropping theme.primary and Theme.primarySelected. I guess it inherits the border color though. I'll let @bbedward do a quick review though.

Issue #1959, opened 2026-03-11
Copy Notification Text to Clipboard


oly-nittka:
### Feature Description

Often, I would like to be able to copy (part of) the notification text (or title) to the clipboard.
Ideally by using keyboard shortcuts or ipc commands.

### Use Case

Receiving a passcode or similar in a notification and need to enter it in a web form.

### Compositor(s)

All compositors

### Proposed Solution

I'd love to do a proof-of-concept if someone could give me a pointer into the code on where to start ;-)

nick-linux8:
You would add it here DankMaterialShell/core/internal/notify/notify.go I can see some possibilities but sounds like you want to work it up so this is the file you would be able to implement this in

oly-nittka:
Well actually I'm a novice in go (expert in C++ and Python, though), so any input is of course appreciated :-)
I just thought I'd do my part as a member of the open source community and help wherever I can.

Is dms supposed to run on X11 as well as Wayland? I think we may need to call an external application to actually copy to the clipboard (or does go provide something useful?) and this may differ depending on the $XDG_SESSION_TYPE.

nick-linux8:
No DMS is Wayland only it ties into Wayland's IPC setup which is what makes it work it would never work on X11 but I was thinking of adding a copy button onto the notifications and then add into the listener function and pass to DMS's builtin clipboard I can work on this and test locally and then setup a PR as an enhancement

oly-nittka:
Oh, wow!
That definitely sounds much better than what I had in mind ;-)

nick-linux8:
Got this done now the Devs can review and see what the want to do but I did use all native DMS tools so there was no added dependencies to make it work attached a screenshot of the success

PR #1966, opened 2026-03-11
feat(notify): add 'Copy Text' action to notifications

nick-linux8:
Description

This PR adds a "Copy Text" action button to all system notifications generated by DMS. This allows users to quickly copy notification body text (such as error messages, status updates, OTP Codes, or logs) directly to the Wayland clipboard without needing to highlight or manually re-type the content. Requested On Issue #1959 

Key Technical Improvements

    Internal CLI Integration: Instead of relying on external binaries like wl-copy, this implementation utilizes the built-in dms cl copy command. This ensures the feature works out-of-the-box on minimal distributions like NixOS where external clipboard utilities may not be in the user's $PATH.

    Native History Support: Because it hooks into the DMS clipboard manager, copied text is automatically captured in the bbolt history database and is accessible via the clipboard manager UI/CLI.

    Decoupled Listener: Leverages the existing notify-action-generic background listener to handle DBus ActionInvoked signals, maintaining a small memory footprint.

Changes

    internal/notify/notify.go:

        Updated Send() to include the "Copy Text" action and pass the notification body to the action listener.

        Extended spawnActionListener and RunActionListener to handle passing and parsing the notification body text.

        Implemented handleAction logic to execute dms cl copy via os.Executable() for reliable path resolution.

Testing Performed (NixOS / Hyprland)

    Compiled with go build -o dms ./core/cmd/dms.

    Triggered test notification: ./dms notify "Test" "Sample Text".

    Verified "Copy Text" button appears in notification.

    Verified clipboard content via ./dms cl paste -> Output: Sample Text.

    Verified history persistence via ./dms cl history -> Entry found with correct ID and timestamp.

### DIFF
--- a/core/internal/notify/notify.go
+++ b/core/internal/notify/notify.go
@@ -50,14 +50,18 @@ func Send(n Notification) error {
 	}
 
 	var actions []string
+	        actions = []string{"copy", "Copy Text"}
+
 	if n.FilePath != "" {
-		actions = []string{
-			"open", "Open",
-			"folder", "Open Folder",
-		}
+		actions = append(actions, "open", "Open", "folder", "Open Folder")
 	}
 
 	hints := map[string]dbus.Variant{}
+	if n.FilePath != "" {
+		hints["image_path"] = dbus.MakeVariant(n.FilePath)
+	}
+
+	hints = map[string]dbus.Variant{}
 	if n.FilePath != "" {
 		hints["image_path"] = dbus.MakeVariant(n.FilePath)
 	}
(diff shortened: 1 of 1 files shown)


bbedward:
This is good for DMS-originated notifications, that comes from things like IE screenshot tool or KDE connect. But I think the original issue wanted it on all notifications from any app (which I'm not sure about adding an action to everything, since it kinda clutters the buttons) - but it'd be a separate thing to do similar to the Dismiss action in QML which is basically just hardcoded for every notification

nick-linux8:
That's all 100% true this is just on the Go backend and I hear you about clutter I'll propose an idea for you to consider utilizing the QML frontend logic adding just the standard copy icon and it only shows upon the mouse entering the notification bubble?

Issue #1955, opened 2026-03-10
Freeze when mouse is not moving


arjan-s:
### Compositor

Niri

### Distribution

NixOS

### Select your Installation Method

Distro Packaging

### Was this your original Installation method?

Yes

### dms doctor -vC

Can't execute this because DMS freezes immediately as soon as the mouse doesn't move

### Description

When logging in, DMS starts successfully and shows the first start screen. As long as I keep moving the mouse, it keeps working well, and I even managed to close the first start screen and click some buttons. But when I stop moving the mouse, DMS/Niri freezes completely.

I found out that some input is still accepted while not displaying anything: so I pressed Ctrl+Alt+F2 to switch to the console, which didn't do anything. But when closing and reopening the laptop lid, I'm on the console. This is where I made the log export using `journalctl --user --boot -u niri -u dms`.

I'd love some help analyzing the log to find out why this happens. Apparently Niri is pausing the session, whatever that means, causing it to freeze.

### Expected Behavior

DMS/Niri doesn't freeze when I stop moving the mouse.

### Steps to Reproduce

Don't move the mouse when logging in.

### Error Messages/Logs

mrt 10 13:03:42 laptop systemd[3227]: Starting A scrollable-tiling Wayland compositor...
mrt 10 13:03:42 laptop niri[3338]: 2026-03-10T12:03:42.131800Z  INFO niri: starting version 25.11 (Nixpkgs)
mrt 10 13:03:42 laptop niri[3338]: 2026-03-10T12:03:42.140235Z DEBUG niri_config: loaded config from "/home/user/.config/niri/config.kdl"
mrt 10 13:03:42 laptop niri[3338]: 2026-03-10T12:03:42.201771Z  INFO niri::backend::tty: using as the render node: "/dev/dri/renderD128"
mrt 10 13:03:42 laptop niri[3338]: 2026-03-10T12:03:42.249186Z DEBUG niri::backend::tty: adding device: 57857 "/dev/dri/card1"
mrt 10 13:03:42 laptop niri[3338]: 2026-03-10T12:03:42.249236Z DEBUG niri::backend::tty: this is the primary node
mrt 10 13:03:42 laptop niri[3338]: 2026-03-10T12:03:42.424040Z DEBUG niri::backend::tty: got render node: renderD128
mrt 10 13:03:42 laptop niri[3338]: 2026-03-10T12:03:42.424156Z DEBUG niri::backend::tty: initializing the primary renderer
[...]

PR #1980, opened 2026-03-13
Compositor Pause Hardening

nick-linux8:
Description:
This PR is to work on #1955 and introduces a retry and backoff mechanism to the main Wayland eventDispatcher in wlcontext.go to prevent the shell from permanently freezing during compositor hardware resets or DPMS power-state transitions.

The Problem:
Currently, if ctx.Dispatch() encounters an error that isn't a standard timeout, the dispatcher logs the error and immediately hits return. This permanently kills the shell's event loop. We discovered that certain hardware configurations (specifically AMD GPUs running the amdgpu driver) emit spurious udev change events when shifting idle power states. Strict compositors (like Niri) react to these events by defensively pausing the entire Wayland session. When the session pauses, the socket becomes unresponsive, ctx.Dispatch() throws an error, and DMS exits the loop—causing a permanent UI freeze even after the compositor resumes.

The Solution:

    Added a consecutiveErrors counter to track consecutive dispatch failures.

    Implemented a progressive time.Sleep backoff to prevent CPU spin-looping while the compositor is unresponsive.

    The shell will now patiently wait out temporary compositor pauses (up to a reasonable threshold) and immediately resume normal operations once ctx.Dispatch() succeeds again.

    If the compositor is genuinely dead (hits maxConsecutiveErrors), the shell will exit gracefully as it did before.

### DIFF
--- a/core/internal/server/wlcontext/context.go
+++ b/core/internal/server/wlcontext/context.go
@@ -2,10 +2,10 @@ package wlcontext
 
 import (
 	"fmt"
+	"golang.org/x/sys/unix"
 	"os"
 	"sync"
-
-	"golang.org/x/sys/unix"
+	"time"
 
 	"github.com/AvengeMedia/DankMaterialShell/core/internal/errdefs"
 	"github.com/AvengeMedia/DankMaterialShell/core/internal/log"
@@ -123,6 +123,9 @@ func (sc *SharedContext) eventDispatcher() {
 		{Fd: int32(sc.wakeR), Events: unix.POLLIN},
 	}
 
+	consecutiveErrors := 0
+	const maxConsecutiveErrors = 20
+
 	for {
 		sc.drainCmdQueue()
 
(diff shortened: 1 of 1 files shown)


core/internal/server/wlcontext/context.go
```diff
@@ -153,9 +156,19 @@ func (sc *SharedContext) eventDispatcher() {
 		}
 
 		if err := ctx.Dispatch(); err != nil && !os.IsTimeout(err) {
-			log.Errorf("Wayland connection error: %v", err)
-			return
+			consecutiveErrors++
+			log.Warnf("Wayland connection error (%d/%d): %v", consecutiveErrors, maxConsecutiveErrors, err)
+
+			if consecutiveErrors >= maxConsecutiveErrors {
+				log.Errorf("Fatal: Wayland connection unrecoverable after %d attempts. Exiting dispatcher.", maxConsecutiveErrors)
```
bbedward, line 163: We should probably really propagate this up, so the shell can clean everything up and exit with an error code. Some of the other goroutines do that, including probably this one if it actually returns an error instead of just killing the loop.

PR #1985, opened 2026-03-13
New neovim theme engine

Silzinc:
I've been working on this during the past weeks, and here it is. Have a demo:

https://github.com/user-attachments/assets/c2a649be-50dd-4ad5-858e-399cb37b7d0c

Advantages compared to the previous one:
- No dependency on `lazy.nvim`. The user is free to install the plugin in whichever way he/she prefers.
- More idiomatic neovim colorscheme definition, and more graceful handling of errors (although it's not perfect).
- Dozens of pretty cohesive base themes are available to derive the colorscheme `dms`. They are essentially NvChad's excellent themes, reworked to integrate with any neovim config as a standalone plugin, and to be customized dynamically by dms.
- The user can choose how much the base theme should be tinted towards the source color of the current DMS theme.

A few problems still:
- Maybe the settings UI will feel cluttered if we start doing what I did here for other editors or matugen themes (I imagine VSCode users would also like an improvement on the dms plugin). 
- The different new settings all point to the same key, which makes them harder to fuzzy find in the settings window. I am not sure as to how this can be solved simply.

### DIFF
--- a/quickshell/Modules/Settings/ThemeColorsTab.qml
+++ b/quickshell/Modules/Settings/ThemeColorsTab.qml
@@ -2578,7 +2578,7 @@ Item {
                     tags: ["matugen", "neovim", "terminal", "template"]
                     settingKey: "matugenTemplateNeovim"
                     text: "neovim"
-                    description: getTemplateDescription("nvim", I18n.tr("Required plugin: ") + "https://github.com/Silzinc/base46")
+                    description: getTemplateDescription("nvim", I18n.tr("Required plugin: ") + "https://github.com/AvengeMedia/base46")
                     descriptionColor: getTemplateDescriptionColor("nvim")
                     visible: SettingsData.runDmsMatugenTemplates
                     checked: SettingsData.matugenTemplateNeovim
--- a/quickshell/matugen/templates/neovim.lua
+++ b/quickshell/matugen/templates/neovim.lua
@@ -1,7 +1,7 @@
 local present, base46 = pcall(require, "base46")
-if not present or base46._AUTHOR ~= "Silzinc" then
+if not present or not base46._DMS_SUPPORT then
 	vim.notify(
-		"base46 plugin not found or incorrect, make sure to install Silzinc/base46",
+		"base46 plugin not found or incorrect, make sure to install AvengeMedia/base46",
 		vim.log.levels.ERROR,
 		{ title = "dms integration" }
 	)


Silzinc:
- Very good point on the dependency to a personal fork. Well, there was a dependency on lazy.nvim before, which is even harder in my opinion as it is hard to integrate in many neovim configs, although it is a well known project with many contributors. My base46 plugin is very much self-contained and config-agnostic, and the template does not forcibly install it but rather asks the user to do it if they want to use the engine. 
  And there also was base16-nvim which is not too far from my fork in nature, although it does have a lot more users. I do see an advantage in the fact that the person in charge of the themeing engine plugin uses DMS, I can tailor it to fit its needs (just like I did up until now).
- I forgot that matugen 4 was breaking things apart. I am still using matugen 3. I guess I will have to update before that PR gets a chance to be merged.
- The `_AUTHOR` field is just my way of checking that it's not one of the many `base46` forks existing around (or even just the NvChad plugin which is quite popular), but I admit it is hacky. I can perfectly envisage changing this to a `_DMS_SUPPORT = true` field, that would exist both on my fork and on any subfork if I were to abandon it (I cannot promise that I won't), or even on a fork by AvengeMedia.

Also, I am myself a NixOS user, although on stable branch, and I haven't updated my flakes in a while. Maybe this is why it works on my machine and not on yours out of the box.

bbedward:
I'd like input from more neovim users on the base46 part, but I'd like to migrate it AvengeMedia org, you can retain maintainer access on it though.

Silzinc:
Migrating to AvengeMedia org is perfectly fine by me, as long as the plugin remains available to non-dms users. That'd probably give better visibility to the plugin, and I don't feel comfy hosting a dependency of DMS on my account anyways. 

Idk how to get input from neovim users... I guess they could just test it once merged, but it looks a bit tiresome to make both engines coexist to avoid unwanted breaks.

bbedward:
Can you do the transfer request, then we can update the refs and just merge it in and see what people say about it

Adarsh219:
> I'd like input from more neovim users on the base46 part, but I'd like to migrate it AvengeMedia org, you can retain maintainer access on it though.

It's just a thought, and I'm not sure how viable it is, but what do you think about expanding Dank16 to Dank46 and utilising it for all code editors?

bbedward:
> > I'd like input from more neovim users on the base46 part, but I'd like to migrate it AvengeMedia org, you can retain maintainer access on it though.
> 
> It's just a thought, and I'm not sure how viable it is, but what do you think about expanding Dank16 to Dank46 and utilising it for all code editors?

Yea it would be awesome to have these options available for VScode, IMO - and zed, etc.

Silzinc:
I suppose base46 can serve as a ground for improving theming engines for lots of other pieces of software, maybe even not just code editors. Siduck (the creator of NvChad) really did an amazing job with this design. 

The idea of shifting the hues of a popular theme is rather basic and applicable in many contexts.

Also @bbedward I just opened a transfer request to you.

bbedward:
I transfered and gave you write access to the repo

Silzinc:
Also, since this is the first time I transfer a repo, is there anything I have to care about with licenses? There are 3 licenses on base46 now that require to be present in any fork of the code (the original base16 one, that of siduck and mine). I dunno if a fourth one is necessary now that the org owns it.

hnsylzg:
In light mode, some of the text/font colors are too faint and hard to read against the light background.

Silzinc:
> In light mode, some of the text/font colors are too faint and hard to read against the light background.

Can you give an example? (Base theme, harmony level, background color, source color)
All info can be found either in ~/.config/nvim/colors/dms.lua or in the DMS settings.

hnsylzg:
Screenshot from 2026-03-16 19-57-57
Screenshot from 2026-03-16 19-58-15


I would like to point out that I have tested all three modes for the Base theme settings ，and the issue persists in all of them. In light mode, the text contrast is consistently too low.Everything else is set to default.
Base theme: It defaults to github_light .
Harmony level: 0.5 (default fallback).

dms.lua

```
local present, base46 = pcall(require, "base46")
if not present or not base46._DMS_SUPPORT then
	vim.notify(
		"base46 plugin not found or incorrect, make sure to install AvengeMedia/base46",
		vim.log.levels.ERROR,
		{ title = "dms integration" }
	)
	return
end

local config_home = vim.env.XDG_CONFIG_HOME
if config_home == nil or #config_home == 0 then
	config_home = vim.fs.joinpath(vim.env.HOME, ".config")
end
local settings_file_path = vim.fs.joinpath(config_home, "DankMaterialShell", "settings.json")
local settings_file = io.open(settings_file_path, "r")
if settings_file == nil then
	vim.notify(
		"cannnot read dms settings file at '" .. settings_file_path .. "'",
		vim.log.levels.ERROR,
		{ title = "dms integration" }
	)
	return
end
local settings = vim.json.decode(settings_file:read("*a"))
settings_file:close()

local function deepGet(t, k)
	for _, s in ipairs(k) do
		if type(t) ~= "table" then
			return
		end
		t = t[s]
	end
	return t
end

local current_file_path = debug.getinfo(1, "S").source:sub(2)
local theme_base = deepGet(settings, { "matugenTemplateNeovimSettings", vim.o.background, "baseTheme" })
	or ("github_" .. vim.o.background)
local harmony = deepGet(settings, { "matugenTemplateNeovimSettings", vim.o.background, "harmony" }) or 0.5
local theme_name = "dms"

if not _G._matugen_theme_watcher then
	local uv = vim.uv or vim.loop
	_G._matugen_theme_watcher = { uv.new_fs_event(), uv.new_fs_event(), reload_timer = uv.new_timer() }

	local debounce_time = 100 -- ms
	local function handler()
		_G._matugen_theme_watcher.reload_timer:stop()
		_G._matugen_theme_watcher.reload_timer:start(
			debounce_time,
			0,
			vim.schedule_wrap(function()
				base46.theme_tables[theme_name] = nil
				if vim.g.colors_name == theme_name then
					vim.cmd.colorscheme(theme_name)
					vim.notify("Theme reload", vim.log.levels.INFO, { title = "dms integration" })
				end
				-- NOTE: contrary to what the documentation says, uv fs events usually do not manage to react to more than one edit.
				-- I understand that this is not intended: some edit processes in a typical system (e.g. the one neovim uses with
				-- multiple renames and changes) make things hard to follow for libuv. Therefore, a restart is the best option.
				_G._matugen_theme_watcher[1]:stop()
				_G._matugen_theme_watcher[2]:stop()
				_G._matugen_theme_watcher[1]:start(current_file_path, {}, handler)
				_G._matugen_theme_watcher[2]:start(settings_file_path, {}, handler)
			end)
		)
	end
	_G._matugen_theme_watcher[1]:start(current_file_path, {}, handler)
	_G._matugen_theme_watcher[2]:start(settings_file_path, {}, handler)
end

if not base46.theme_tables[theme_name] or base46.theme_tables[theme_name].type ~= vim.o.background then
	local builtin = vim.deepcopy(assert(base46.get_builtin_theme(theme_base)))
	local harmonized = base46.theme_harmonize(builtin, "#1976d2", harmony)
	harmonized = base46.theme_set_bg(harmonized, "#f7f9ff")

	base46.theme_tables[theme_name] = harmonized
end

base46.load(theme_name)
vim.g.colors_name = theme_name
```

Silzinc:
The pictures I see do actually look a lot like github_dark. What do you see when running `=vim.o.bg`?

hnsylzg:
I checked it, and =vim.o.background is indeed dark.

Silzinc:
I have been careless. On my system, neovim updates `vim.o.bg` automatically and reloads the colorscheme when it detects that the background of my terminal becomes white, and I have a hook on DMS that changes my kitty theme to make that happen.

I'll change the template to set `vim.o.bg` from the DMS mode. For the time being, you can just use `set bg=light` when opening neovim.

hnsylzg:
Thanks for the explanation! That makes sense. I don't have that terminal hook, so Neovim stays in dark mode by default.

Using :set bg=light manually works perfectly for now and restores the contrast. Looking forward to the template update that syncs vim.o.bg with the DMS mode!

Silzinc:
Note that I know very few themes that set `background` manually actually. It's really made to sync neovim's apparence with that of the terminal, hence that special and rather unique colorscheme reload behaviour that is built into neovim.

NvChad sets `vim.o.bg` and I kept that behaviour even though I am not sure if it is idiomatic. For a DMS integration, I guess it can make sense since I can just get the theme with an IPC call. Well, if you want to have a dark terminal and a white neovim I guess that's the best way out.

PR #1997, opened 2026-03-14
fix(nix/greeter): skip invalid customThemeFile in preStart

z1-0:
Avoid attempting to copy a null/empty/missing customThemeFile path by validating the jq result and file existence before cp.

The default implementation of wrap systemd script in Nix is "set -e", so the code is interrupted here.

### DIFF
--- a/distro/nix/greeter.nix
+++ b/distro/nix/greeter.nix
@@ -196,7 +196,7 @@ in
 
       if [ -f settings.json ]; then
           theme_file="$(${jq} -r '.customThemeFile // empty' settings.json)"
-          if [ -r "$theme_file" ]; then
+          if [ -f "$theme_file" ] && [ -r "$theme_file" ]; then
               cp "$theme_file" custom-theme.json
               mv settings.json settings.orig.json
               ${jq} '.customThemeFile = "${cacheDir}/custom-theme.json"' settings.orig.json > settings.json


distro/nix/greeter.nix
```diff
@@ -195,9 +195,12 @@ in
       fi
 
       if [ -f settings.json ]; then
-          if cp "$(${jq} -r '.customThemeFile' settings.json)" custom-theme.json; then
-              mv settings.json settings.orig.json
-              ${jq} '.customThemeFile = "${cacheDir}/custom-theme.json"' settings.orig.json > settings.json
+          theme_file="$(${jq} -r '.customThemeFile // empty' settings.json)"
+          if [ -n "$theme_file" ] && [ -f "$theme_file" ]; then
```
LuckShiba: I think these three conditions (including L200) could be simplified for a simple `if [ -r "$theme_file" ]`, as it will validate both if theme_file is valid (the `-n` here is redundant) and it's readable.

distro/nix/greeter.nix
```diff
@@ -195,9 +195,12 @@ in
...
+              if cp "$theme_file" custom-theme.json; then
+                  mv settings.json settings.orig.json
+                  ${jq} '.customThemeFile = "${cacheDir}/custom-theme.json"' settings.orig.json > settings.json
+              fi
```
LuckShiba:
```suggestion
          theme_file="$(${jq} -r '.customThemeFile // empty' settings.json)"
          if [ -r "$theme_file" ]; then
              cp "$theme_file" custom-theme.json
              mv settings.json settings.orig.json
              ${jq} '.customThemeFile = "${cacheDir}/custom-theme.json"' settings.orig.json > settings.json
          fi
```

purian23:
Happy to merge when you all think it's good to go

z1-0:
Changes addressed, PTAL

If this PR is accepted, I will submit the same changes to the dms module in the nixpkgs repo

LuckShiba:
It seems to work in my system. I added back the `-f` once I realized `-r` may return true if it's a directory. LGTM!

purian23:
@LuckShiba, please me sure I didn't break anything via 1661d3264181f39ed2027796d7ff56c342185eb8

LuckShiba:
Re-testing now!

LuckShiba:
Seems to work with https://github.com/AvengeMedia/DankMaterialShell/commit/1661d3264181f39ed2027796d7ff56c342185eb8.

PR #1998, opened 2026-03-14
feat(nix/niri): add new includes for dms 1.4

NikSneMC:
I've read the startup code a little bit and it seems like those are being generated at init

### DIFF
--- a/distro/nix/niri.nix
+++ b/distro/nix/niri.nix
@@ -2,11 +2,9 @@
   config,
   lib,
   ...
-}:
-let
+}: let
   cfg = config.programs.dank-material-shell;
-in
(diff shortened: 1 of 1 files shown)


LuckShiba:
Thanks :)

purian23:
My only concern was that we already include cursor in the "includes" file. But if this works, all good.

LuckShiba:
> My only concern was that we already include cursor in the "includes" file. But if this works, all good.

The "includes" file?

purian23:
> > My only concern was that we already include cursor in the "includes" file. But if this works, all good.
> 
> The "includes" file?

Yes

PR #1999, opened 2026-03-14
feat: add color picker to launcher

ronharel02:
Adds a built-in `Color Picker` launcher entry and wires its action to open the color picker popout via `ipc:color-picker`.


purian23:
Nice, thanks!

Issue #1992, opened 2026-03-13
Greeter ignores theme accent color (other than default)


Useekaw:
### Compositor

Niri

### Distribution

Arch Linux

### Select your Installation Method

Distro Packaging

### Was this your original Installation method?

Yes

### dms doctor -vC

<details>
<summary>Click to expand</summary>
## DMS Doctor Report

**System**
- [ok] Operating System: CachyOS
  - ID: cachyos, Version: rolling, Arch: amd64
- [ok] Architecture: amd64
- [ok] Display Server: Wayland
  - WAYLAND_DISPLAY=wayland-1

**Versions**
- [ok] DMS CLI: dms v1.4.0-224-ge86227f0
  - /usr/bin/dms
- [ok] Quickshell: quickshell 0.2.1, revision dacfa9de829ac7cb173825f593236bf2c21f637e, distributed by: CachyOS Linux
  - /usr/bin/qs
- [ok] DMS Shell: v1.5-beta
  - /usr/share/quickshell/dms

**Installation**
- [ok] DMS Configuration: Found
  - /usr/share/quickshell/dms
- [ok] shell.qml: Present
  - /usr/share/quickshell/dms/shell.qml
- [info] Install Type: System package
  - /usr/share/quickshell/dms

**Compositor**
- [ok] niri: 25.11
  - /usr/bin/niri
- [info] Active: niri

**Quickshell Features**
- [ok] Polkit: Available
  - Authentication prompts
- [ok] IdleMonitor: Available
  - Idle detection
- [ok] IdleInhibitor: Available
  - Prevent idle/sleep
- [ok] ShortcutInhibitor: Available
  - Allow shortcut management (niri)

**Optional Features**
- [ok] accountsservice: Available
  - User accounts
- [ok] power-profiles-daemon: Available
  - Power profile management
- [ok] logind: Available
  - Session management
- [warn] cups-pk-helper: Not available (install cups-pk-helper)
  - Printer management
- [info] I2C/DDC: No monitors detected
  - External monitor brightness control
- [ok] qt6-imageformats: Installed (1 formats)
  - Formats: GIF (/usr/lib/qt6/plugins/imageformats:/usr/lib64/qt6/plugins/imageformats)
- [ok] kimageformats: Installed (4 formats)
  - Formats: AVIF, HEIF, JXL, EXR (/usr/lib/qt6/plugins/imageformats:/usr/lib64/qt6/plugins/imageformats)
[...]

PR #2000, opened 2026-03-15
Change Greetd logic to pull accent color

nick-linux8:
Greeter wasn't using the accent color from the users set theme. For Issue #1992 

Root cause: Theme.loadCustomTheme was calling 
SettingsData.getRegistryThemeMultiVariant in all contexts including 
greeter mode. In greeter mode SettingsData does not have the user's 
registryThemeVariants populated, so it always fell back to 
modeDefaults.accent from the theme file. GreetdSettings was also 
not reading registryThemeVariants from the synced settings file.

Files Changes:
- GreetdSettings.qml: add registryThemeVariants property and read it 
  in parseSettings
- Theme.qml: loadCustomTheme now uses GreetdSettings.registryThemeVariants 
  in greeter mode instead of SettingsData

Confirmed dms greeter sync already copies the full settings.json 
(commands_greeter.go:1507) so registryThemeVariants is available 
in the greeter cache after sync — no Go changes needed.

Note: Unable to test end-to-end as I don't run greetd. Tested logic 
by code path analysis. Needs verification from someone running greetd 
with a multi-variant theme and non-default accent.

### DIFF
--- a/quickshell/Common/Theme.qml
+++ b/quickshell/Common/Theme.qml
@@ -1248,7 +1248,11 @@ Singleton {
             if (themeData.variants.type === "multi" && themeData.variants.flavors && themeData.variants.accents) {
                 const defaults = themeData.variants.defaults || {};
                 const modeDefaults = defaults[colorMode] || defaults.dark || {};
-                const stored = typeof SettingsData !== "undefined" ? SettingsData.getRegistryThemeMultiVariant(themeId, modeDefaults, colorMode) : modeDefaults;
+                const isGreeterMode = typeof SessionData !== "undefined" && SessionData.isGreeterMode;
+		const stored = isGreeterMode ?
+		      (GreetdSettings.registryThemeVariants[themeId]?.[colorMode] || modeDefaults) :
+		      (typeof SettingsData !== "undefined" ?
+		       SettingsData.getRegistryThemeMultiVariant(themeId, modeDefaults, colorMode) : modeDefaults);
                 var flavorId = stored.flavor || modeDefaults.flavor || "";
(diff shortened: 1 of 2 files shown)


purian23:
Thanks! Also Hmm my accent color works fine as-is. Let me test with this soon.

purian23:
I treat Nix like such an Outlaw :rofl:

lakapple:
I'm using Catppuccin on CachyOS and also have the same issue

This PR doesnt fix the issue for me tho

nick-linux8:
Thanks for testing @lakapple  A couple of questions before I had back to the drawing board on the issue — did you run `dms greeter sync` after running this change? Also could you check if `registryThemeVariants` is present in `/var/cache/dms-greeter/settings.json` after syncing? Also check for `registryThemeVariants` in `~/.config/DankMaterialShell/settings.json` to rule out the sync not having anything to pull

lakapple:
My theme does appear in `registryThemeVariants`, I ran `dms gretter sync` and even rebooted.

nick-linux8:
Which one exists the one in /var or ~/.config? Or both because that will tell me what direction I need to go also What theme are you using and from where? Also @Purian23 would you mind answering that as well for you as a reference point?

lakapple:
Both exists

purian23:
Thanks for the PR here again, I'm merging this in and added a small section to it. I found a regression in the auto theme accent that I'm bundling in to fix both at once!

lakapple:
The problem still exists on my machine after this PR

purian23:
Right, expected for now. Follow up to it is in the works.

PR #2009, opened 2026-03-16
added non-flake nix compatibility with flake-compat

sunworms:
This PR aims to add compatibility for non-flake setups by adding the input `NixOS/flake-compat` and creating a `default.nix` and a `shell.nix` that provide the same outputs as the ones provided by the flake.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,11 @@
+(import (
+  let
+    lock = builtins.fromJSON (builtins.readFile ./flake.lock);
+  in
+  fetchTarball {
+    url =
+      lock.nodes.flake-compat.locked.url
+        or "https://github.com/NixOS/flake-compat/archive/${lock.nodes.flake-compat.locked.rev}.tar.gz";
+    sha256 = lock.nodes.flake-compat.locked.narHash;
+  }
+) { src = ./.; }).defaultNix
--- a/distro/nix/default.nix
+++ b/distro/nix/default.nix
@@ -1,6 +1,6 @@
 (import (
   let
-    lock = builtins.fromJSON (builtins.readFile ./flake.lock);
+    lock = builtins.fromJSON (builtins.readFile ../../flake.lock);
   in
   fetchTarball {
     url =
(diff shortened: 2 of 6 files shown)


marcusramberg (approved):
looks ok.

LuckShiba (commented):
The addition looks okay, but I'm thinking about how necessary that is. Users that doesn't want to use the flake should in a best-case scenario use the Nixpkgs module, but it's still not in stable so I guess that is okay.

The only problem here is that it'll add two new top-level Nix files, which is okay but not ideal. @sunworms, I don't know much about non-flake usage, would it be viable having these files inside distro/nix or not at all?

LuckShiba:
By testing here it seems to work by moving the files to `distro/nix`, changing the paths to reference `../../` and running `nix-shell distro/nix` or `nix-build distro/nix`, couldn't that be better?

purian23:
@sunworms have you looked into the last comment by LuckShiba? I would like to get this in or cleared out if we can.

sunworms:
> By testing here it seems to work by moving the files to `distro/nix`, changing the paths to reference `../../` and running `nix-shell distro/nix` or `nix-build distro/nix`, couldn't that be better?

That could work too, I kept them top-level because it'll be easier for people who use something like `npins` or `nvfetcher`. For example people who use npins can do something like this easily:
```nix
let
  sources = import ./npins; # or wherever their npins directory is
  dms = import sources.dms {};
in
{
  imports = [
    dms.nixosModules.default
  ];
  
  programs.dank-material-shell.enable = true;
}
```

LuckShiba:
Since it's more of an edge case (most people use flakes!) I've decided to change them to `distro/nix` to avoid having to put more nix-only files on top-level. Tested with `nix-build ./distro/nix` and `nix-shell ./distro/nix`.

It also shouldn't be basically any much different for, as in your example, npins users:
```nix
let
  sources = import ./npins;
  dms = import (sources.dms + "/distro/nix") {};
in
{
  imports = [
    dms.nixosModules.default
  ];

  programs.dank-material-shell.enable = true;
}
```

Thank you for your contribution!

PR #2069, opened 2026-03-23
feat: add FullscreenToplevel detection

kanghengliu:
For animating bar for fullscreen events.

https://github.com/user-attachments/assets/88d831a7-d7ab-4b18-8ee8-b8824ae6aa86

### DIFF
--- a/quickshell/Modules/DankBar/DankBarWindow.qml
+++ b/quickshell/Modules/DankBar/DankBarWindow.qml
@@ -683,12 +683,12 @@ PanelWindow {
         }
 
         property bool reveal: {
-            if (barWindow.hasFullscreenToplevel)
-                return false;
-
             const inOverviewWithShow = CompositorService.isNiri && NiriService.inOverview && (barConfig?.openOnOverview ?? false);
             if (inOverviewWithShow)
                 return true;
+
+            if (barWindow.hasFullscreenToplevel)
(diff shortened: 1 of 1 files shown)


benabraham:
This checks fullscreen on any window in a workspace but in niri you can have fullscreen window in different column than is currently visible/focused. So this hides bars when they should be visible. Should I start an issue?

kanghengliu:
@benabraham Good point, Yes it is an issue with niri. I wonder if there is a non-hacky way to determine if there's still a full-screen window visible in niri so that we don't occlude content when unneeded.

kanghengliu:
Hotfix #2089 to align with the previous show/hide strategy. In the future it's probably good to have intelligent auto-hide for the bar as well.

benabraham:
The activated property on toplevels already tracks whether a window is currently focused. WallpaperCyclingService.qml uses exactly this pattern:
```
if (toplevel.fullscreen && toplevel.activated)
```
In Niri, a fullscreen window in a scrolled-off column is still on the active workspace but isn't activated. Adding `&& filtered[i]?.activated` to the check in `_updateHasFullscreenToplevel()` might fix it for Niri (and is correct for other compositors too).


_disclaimer: I let AI research this for me and I haven't got deep understanding of DMS but sounds plausible enough…_

PR #2077, opened 2026-03-25
feat: FileBrowser video thumbnail

kanghengliu:
This PR adds support for video files' thumbnail generation with `ffmpegthumbnailer`

## Features
- thumbnail lookup and generation
  - find thumbnails in XDG_CACHE_HOME first, generate with `Proc.runCommand` when not found.
- fallback to video icon placeholder if `ffmpegthumbnailer` not found in PATH. 

## Demo

https://github.com/user-attachments/assets/d20d6a5a-2f4f-4e17-b7e8-f2ece2e8ddd0

## Notes
This PR is mainly for support with [dms-mpvpaper](https://github.com/kanghengliu/dms-mpvpaper) plugin; It reuses the filebrowser for wallpaper selector.

### DIFF
--- a/quickshell/Common/Paths.qml
+++ b/quickshell/Common/Paths.qml
@@ -10,6 +10,7 @@ Singleton {
 
     readonly property url home: StandardPaths.standardLocations(StandardPaths.HomeLocation)[0]
     readonly property url pictures: StandardPaths.standardLocations(StandardPaths.PicturesLocation)[0]
+    readonly property url xdgCache: StandardPaths.standardLocations(StandardPaths.GenericCacheLocation)[0]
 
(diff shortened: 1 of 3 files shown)


bbedward:
I wonder if there's some QML plugin or something that can get video thumbnail previews in a more native way, thought it may not be possible with QML only.

kanghengliu:
@bbedward That's a good point. With limited knowledge on the Qt apps landscape, I dug around at how Dolphin from KDE handles thumbnails, and they use a C-backed [KIO](https://invent.kde.org/frameworks/kio/-/tree/master?ref_type=heads) framework to handle the thumbnail generation. 

In the future we can consider building/integrating a similar toolset in the go backend to handle thumbnails for more filetypes.

PR #2078, opened 2026-03-25
Neovim template enhancements

Silzinc:
- Add a matugen template to support the popular neovim `lualine` plugin automatically.
- Add an option to follow the DMS background color in neovim (was forced into true before).
- Neovim's dark/light mode will try to follow that of DMS.

### DIFF
--- a/quickshell/translations/en.json
+++ b/quickshell/translations/en.json
@@ -10490,7 +10490,7 @@
   {
     "term": "Search...",
     "context": "Search...",
-    "reference": "Widgets/DankDropdown.qml:301, Modals/Settings/SettingsSidebar.qml:641, Modules/ProcessList/ProcessListPopout.qml:185",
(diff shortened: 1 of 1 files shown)


ajensen1234:
Hey, this PR breaks startup for users on existing installs. The new `matugenTemplateNeovimSetBackground` key is missing from any `settings.json` that predates this commit, causing:

```
Unable to parse settings.json
Cannot assign [undefined] to bool
```

This caused DMS to return to the default settings after restarting.

I fixed this manually by adding `matugenTemplateNeovimSetBackground` to `./config/DankMaterialShell/settings.json`. I think the main issue is that `SettingsSpec.json` isn't used as a fallback for an existing config.

Unfortunately, I'm not well-versed enough with the codebase to be able to create such a change. I hope this information helps.

Silzinc:
I do not really understand what I can change to solve this, I already added the right field to SettingsSpec.js. I think I did not see this because I had set that field manually in my settings.json on the first attempt. Did something change in the way DMS updates the settings scheme?

Silzinc:
Mb I'm an idiot, I set it to `true` and not to `{ def: true }`. Opening another PR right away.

PR #2081, opened 2026-03-25
fix: set default value for matugenTemplateNeovimSetBackground

Silzinc:
Answers startup issues encountered after #2078 was merged. My apologies for the silly mistake.


ajensen1234:
Thanks for the quick fix. Tested it locally and everything worked.

PR #2085, opened 2026-03-25
fix: pywalfox output name is wrong

mochouaaaaa:
Resolves errors caused by inconsistencies between the names `output_path` and `post_hook`.

### DIFF
--- a/quickshell/matugen/configs/pywalfox.toml
+++ b/quickshell/matugen/configs/pywalfox.toml
@@ -1,4 +1,4 @@
 [templates.dmspywalfox]
 input_path = 'SHELL_DIR/matugen/templates/pywalfox-colors.json'
-output_path = 'CACHE_DIR/wal/dank-pywalfox.json'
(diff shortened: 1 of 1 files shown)


quickshell/matugen/configs/pywalfox.toml
```diff
@@ -1,4 +1,4 @@
 [templates.dmspywalfox]
 input_path = 'SHELL_DIR/matugen/templates/pywalfox-colors.json'
-output_path = 'CACHE_DIR/wal/dank-pywalfox.json'
```
bbedward, line 3: The idea, I think at least, is to make it so they had to be explicitly enabled https://danklinux.com/docs/dankmaterialshell/application-themes#option-2-pywalfox

Although I'm ok with this approach but we should make pywalfox default to false in SettingsSpec/SettingsData

mochouaaaaa: Okay, then I'll close this PR, because the steps in the documentation are correct.

PR #2086, opened 2026-03-26
feat(danklauncher): add launcher history

### LINKED ISSUE #2055: Remember last search query in Dank Launcher
### Feature Description

I'd like to see an option that allows Dank Launcher to remember the last typed search query, like what's found in KDE Plasma's KRunner for example. Currently, only the last category (All, Apps, Files or Plugins) is saved, so this feels like a logical continuation.

### Use Case

A lot of times I find myself to launch the same thing again, or at least something very similar in name such that I don't have to type everything from scratch again. This can save a bit of time on my part.

### Compositor(s)

All compositors

### If Other, please specify

_No response_

### Proposed Solution

N/A but it probably wouldn't be very hard to implement given the last search category is already saved.

_No response_

lakapple:
Closes #2055

### DIFF
--- a/quickshell/Modals/DankLauncherV2/Controller.qml
+++ b/quickshell/Modals/DankLauncherV2/Controller.qml
@@ -45,9 +45,7 @@ Item {
 
     onActiveChanged: {
         if (!active) {
-            if (searchQuery && searchQuery.trim().length > 0) {
-                SessionData.addLauncherHistory(searchQuery);
-            }
+            SessionData.addLauncherHistory(searchQuery);
 
             sections = [];
             flatModel = [];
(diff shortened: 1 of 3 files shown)


lakapple:
https://github.com/user-attachments/assets/e75f761a-22a5-4077-aa26-a7d2487cacb4

kanghengliu:
Could this be triggered via a keyboard shortcut? \<C-r\> or Uparrow for example to populate the last query, and by default leave empty. I vaguely remember walker does the same.

lakapple:
Good idea. I didnt think of it. Thank you.

lakapple:
@bbedward appDrawerLastMode is separated from launcherLastMode, should that be the same for appDrawerLastQuery? Or just unify them under launcherLastMode and launcherLastQuery?

bbedward:
> @bbedward appDrawerLastMode is separated from launcherLastMode, should that be the same for appDrawerLastQuery? Or just unify them under launcherLastMode and launcherLastQuery?

Probably doesn't have to be for this case - that one I like to be separate since I like the overview one to always default to apps while the IPC one defaults to all

lakapple:
https://github.com/user-attachments/assets/c7690937-5748-49d9-8f0f-3445b9fb19ba

Navigate history with ctrl + arrow up and down
Added option for autofilling last query on launcher openning.

Issue #2088, opened 2026-03-26
Bar is hidden when scrolling away from a fullscreen window


kanghengliu:
### Compositor

Niri

### Distribution

Arch Linux

### Select your Installation Method

Distro Packaging

### Was this your original Installation method?

Yes

### dms doctor -vC

<details>
<summary>Click to expand</summary>
..

</details>

### Description
Regression. 

PR #2069 [comment](https://github.com/AvengeMedia/DankMaterialShell/pull/2069#issuecomment-4131808895)
> in niri you can have fullscreen window in different column than is currently visible/focused. So this hides bars when they should be visible. 

### Steps to Reproduce

1. fullscreen a window 
2. scroll away

avivbintangaringga:
same issue here

cwelsys:
Adding my experience and a way to reproduce (on niri):

- apply toggle-windowed-fullscreen to a window of your choice
- switch focus to another window
- watch as the bar turtles

I applied #2089 locally and it does address this, in that the bar doesn't hide when switching away from the 'fullscreen' window, it now hides when the 'fullscreen' window is in focus. This is likely the desired behavior for most cases. 

Ironically, given the test case, I don't think you would want the top bar hidden when setting a window to [fake fullscreen](https://github.com/niri-wm/niri/wiki/Fullscreen-and-Maximize#windowed-fullscreen). This behavior was introduced in #2069 with no opt-out.

kanghengliu:
Hmm. The fake fullscreen window is picked up by quickshell.wayland as a fullscreen toplevel. I'm not versed in niri to know if it's a deliberate design. Either that or when we get some proper Niri support in Quickshell.

cwelsys:
There is an open PR over at niri-wm to add fullscreen states to the ipc, if that gets merged then DMS's NiriService.qml could be extended to distinguish between those states. Though I'm not even sure that all is needed for this case, as niri already renders fullscreen windows above everything else (including the bar). I opened a PR which just drops the niri check in _updateHasFullscreenToplevel() 

with that, #2089 isn't needed. It will still work on Hyprland.

PR #2091, opened 2026-03-26
fix(bar): exclude niri from fullscreen toplevel detection

cwelsys:
Niri already renders fullscreen windows above the top layer-shell layer (see render_above_top_layer() in niri's scrolling.rs). Aside from redundancy, this check also hides the bar for windowed-fullscreen windows (aka fake fullscreen), since the Wayland protocol reports identical state for both.

### DIFF
--- a/quickshell/Modules/DankBar/DankBarWindow.qml
+++ b/quickshell/Modules/DankBar/DankBarWindow.qml
@@ -181,7 +181,7 @@ PanelWindow {
     }
 
     function _updateHasFullscreenToplevel() {
-        if (!CompositorService.isHyprland && !CompositorService.isNiri) {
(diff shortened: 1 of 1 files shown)


kanghengliu:
This basically excludes Niri from fullscreen-event triggered reveal, preventing reveal/hide animations from taking place. As discussed in issue #2088 , It's difficult to have best of both worlds w/o some level of support from Niri Ipc. 

Also, If this ends up being preferred, It's better to properly rewrite the fullscreen detection logic since Hyprland exposes the fakefullscreen state.

The real trade-off from this is providing support for Niri's [windowed fullscreen](https://github.com/niri-wm/niri/wiki/Fullscreen-and-Maximize#windowed-fullscreen)

Issue #2092, opened 2026-03-27
Niri empty workspace name results in unexpected behavior


SturgeonInc:
### Compositor

Niri

### Distribution

Fedora

### Select your Installation Method

Distro Packaging

### Was this your original Installation method?

Yes

### dms doctor -vC

<details>
<summary>Click to expand</summary>

## DMS Doctor Report

**System**
- [ok] Operating System: Fedora Linux 43 (Sway)
  - ID: fedora, Version: 43, Arch: amd64
- [ok] Architecture: amd64
- [ok] Display Server: Wayland
  - WAYLAND_DISPLAY=wayland-1

**Versions**
- [ok] DMS CLI: dms v1.4.4
  - /usr/bin/dms
- [ok] Quickshell: quickshell 0.2.1, revision 26531fc46ef17e9365b03770edd3fb9206fcb460, distributed by: Fedora COPR (avengemedia/quickshell)
  - /usr/bin/qs
- [ok] DMS Shell: 1.4.4
  - /usr/share/quickshell/dms

**Installation**
- [ok] DMS Configuration: Found
  - /usr/share/quickshell/dms
- [ok] shell.qml: Present
  - /usr/share/quickshell/dms/shell.qml
- [info] Install Type: System package
  - /usr/share/quickshell/dms

**Compositor**
- [ok] niri: 25.11
  - /usr/bin/niri
- [ok] Sway: 1.11
  - /usr/bin/sway
- [info] Active: niri

**Quickshell Features**
- [ok] Polkit: Available
  - Authentication prompts
- [ok] IdleMonitor: Available
  - Idle detection
- [ok] IdleInhibitor: Available
  - Prevent idle/sleep
- [ok] ShortcutInhibitor: Available
  - Allow shortcut management (niri)

**Optional Features**
- [ok] accountsservice: Available
  - User accounts
- [warn] power-profiles-daemon: Not available
  - Power profile management
- [ok] logind: Available
  - Session management
- [ok] cups-pk-helper: Available
  - Printer management
- [info] I2C/DDC: No monitors detected
  - External monitor brightness control
- [ok] qt6-imageformats: Installed (1 formats)
  - Formats: GIF (/usr/lib64/qt6/plugins/imageformats)
- [ok] kimageformats: Installed (4 formats)
  - Formats: AVIF, HEIF, JXL, EXR (/usr/lib64/qt6/plugins/imageformats)
- [ok] Terminal: alacritty
[...]

PR #2094, opened 2026-03-28
fix: use UnsetWorkspaceName to fix  empty input in workspace rename

Tobi-maru:
Problem : previously if we open up workspace renaming dialoag box and press enter without typing anything it would rename the workspace to an empty string . So if we have to remove the name of two workspaces we cant do that because it would set the name of two workspaces to empty strings and duplicate workspace names are not allowed .

Fix : Used unsetWorkspaceName  to unset the workspace name instead of setting it to empty string .


Tobi-maru:
Closes #2092

PR #2105, opened 2026-03-30
chore(niri): use satty screenshot tool as default

tulilirockz:
Swappy seems to not be quite as well maintained as satty, latest commit (https://github.com/jtheoof/swappy/commit/ff7d641b8c0d461b8a90448a5893e4aa3a0533b1) is a translation, before that the latest commit was in august 2025. Latest release is also from Aug 27 2025.

Satty however has had commits quite recently and releases as well.
- https://github.com/Satty-org/Satty/commit/590253c8bba01f5c559808dd432cbb80bc4bd7c8 - This was from yesterday
- https://github.com/Satty-org/Satty/releases/tag/v0.20.1 - Febuary 6th


purian23:
Bit of a bummer it isn't packaged on Fedora atm., but it's fairly easy to build. I do like its tooling over Swappy though.

PR #2117, opened 2026-03-31
fix: tray icons duplicating and disappearing after session restore

Dimariqe:
## Problem

After resume from suspend (or forced session restore via Hyprland),
some tray icons disappear. Apps that don't detect a StatusNotifierWatcher
restart don't re-register their SNI items, leaving the tray empty.

Additionally, a `TrayRecoveryService` that scans DBus for unregistered
SNI items after resume could produce **duplicate icons** because apps
may register their SNI item either via their well-known name
(`com.example.AppName/StatusNotifierItem`) or via their connection ID
(`:1.49/StatusNotifierItem`), and the old dedup logic didn't account
for this equivalence.

## Root causes

1. **Directional blind spot** — the well-known name scan and the
   connection-ID scan each only checked their own form against the
   registered items list. If an app registered as `:1.49`, the
   well-known-name section would still try to register
   `com.example.AppName`, and vice versa.

2. **Case-sensitive SNI Id comparison** — some apps report their SNI
   `Id` property in a different case than their well-known name
   (e.g. `"exampleapp"` vs `"com.example.ExampleApp/..."`).
   `grep -qF` never matched, so the fallback dedup check was silently
   skipped every time.

## Solution

Add `TrayRecoveryService` — a singleton that fires 3 seconds after
resume and scans DBus for SNI items that need re-registration.

Key design decisions:

- **Unified `REGISTERED_CONN_IDS` set** built upfront by resolving
  every currently registered item (regardless of whether it was
  registered as a well-known name or a connection ID) to its canonical
  `:1.xxx` connection ID. This is the single source of truth used by
  both scan sections.
- **Well-known name section** first checks `REGISTERED_CONN_IDS` via
  the name's connection ID before probing — skips apps already covered
  by a connection-ID registration.
- **After registering** via well-known name, `REGISTERED_CONN_IDS` is
  updated immediately so the connection-ID section running next won't
  probe the same app again in the same recovery pass.
- **Connection-ID section** uses `REGISTERED_CONN_IDS` as its primary
  gate, with a case-insensitive SNI `Id` string match as a secondary
  fallback inside each parallel subshell.
- **Single `busctl` snapshot** (`BUSCTL_OUT`) is taken once and reused
  across both sections, replacing two redundant `dbus-send ListNames`
  calls.

### DIFF
--- a/quickshell/DMSShell.qml
+++ b/quickshell/DMSShell.qml
@@ -27,6 +27,8 @@ import qs.Services
 Item {
     id: root
 
+    readonly property var _trayRecovery: TrayRecoveryService
+
     Instantiator {
         id: daemonPluginInstantiator
         asynchronous: true
--- a/quickshell/Services/TrayRecoveryService.qml
+++ b/quickshell/Services/TrayRecoveryService.qml
@@ -1,0 +1,146 @@
+pragma Singleton
+pragma ComponentBehavior: Bound
+
+import QtQuick
+import Quickshell
+import Quickshell.Io
+import Quickshell.Services.SystemTray
+
+Singleton {
+    id: root
+
+    // Re-run after resume from suspend
+    Connections {
+        target: SessionService
+        function onSessionResumed() {
+            resumeTimer.restart();
+        }
+    }
+
+    Timer {
+        id: resumeTimer
+        interval: 3000
+        repeat: false
+        running: false
+        onTriggered: root.recoverTrayItems()
+    }
+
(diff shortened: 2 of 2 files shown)


bbedward:
This is nice, though I think we could just integrate it into DMS server, and not need any additional QML for it. That might be a better approach since it's already hooked in to everything.

Dimariqe:
I'll be back with a Go implementation.

PR #2137, opened 2026-04-03
fix: add TrayRecoveryService with bidirectional SNI dedup (Go server)

Dimariqe:
## Go server: add TrayRecoveryService with bidirectional SNI dedup

Go-side implementation of the TrayRecoveryService introduced in #2117.

### What

New `trayrecovery` package (`core/internal/server/trayrecovery/`) that
re-registers lost system tray icons after resume from suspend using
native DBus calls (`godbus/dbus/v5`) instead of shelling out to bash.

### How it works

1. On startup (3 s delay) **and** on `PrepareForSleep true→false`
   transition from loginctl, the service scans the session bus for
   unregistered SNI items.
2. Every registered item (well-known name or `:1.xxx`) is resolved to a
   canonical connection ID, building a unified `registeredConnIDs` set.
   This prevents duplicates in both directions — same logic as the QML
   version from #2117.
3. **Well-known names** (DinoX, nm-applet, etc.) are probed sequentially
   at `/StatusNotifierItem` and `/org/ayatana/NotificationItem/$SHORT`
   with a 300 ms timeout.
4. **Connection IDs** (Vesktop, Electron apps) are probed in parallel
   (batches of 30, 150 ms timeout) with case-insensitive SNI Id dedup.
5. Found items are re-registered via
   `org.kde.StatusNotifierWatcher.RegisterStatusNotifierItem`.

### Why startup scan is needed

The DMS systemd unit is `Type=dbus` with
`BusName=org.freedesktop.Notifications`. During suspend the bus name can
briefly disappear, causing systemd to kill and restart the process. The
fresh process never sees the `PrepareForSleep true→false` transition, so
the loginctl watcher alone is not enough. The startup scan covers this
case and is harmless on a normal boot — it finds all items already
registered and exits early.

### Files changed

| File | Description |
|------|-------------|
| `core/internal/server/trayrecovery/manager.go` | Manager lifecycle, loginctl watcher, startup scan |
| `core/internal/server/trayrecovery/recovery.go` | DBus scanning and registration logic |
| `core/internal/server/server.go` | Integration: init, cleanup, loginctl bridge |

### DIFF
--- a/core/internal/server/trayrecovery/recovery.go
+++ b/core/internal/server/trayrecovery/recovery.go
@@ -145,7 +145,7 @@ func (m *Manager) scanWellKnownNames(allNames []string, registeredItems []string
 
 		for _, objPath := range objectPaths {
 			if m.probeSNI(name, objPath, probeTimeout) {
-				m.registerSNI(name, objPath)
+				m.registerSNI(name)
 				// Update set so the connection-ID section won't double-register this app
 				if connForName != "" {
 					registeredConnIDs[connForName] = true
@@ -190,7 +190,7 @@ func (m *Manager) scanConnectionIDs(allNames []string, registeredItems []string,
 				return
 			}
 
-			m.registerSNI(conn, "/StatusNotifierItem")
+			m.registerSNI(conn)
 			log.Infof("TrayRecovery: re-registered %s (Id: %s)", conn, sniID)
 		}(name)
 	}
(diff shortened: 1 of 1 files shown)


core/internal/server/trayrecovery/recovery.go
```diff
@@ -0,0 +1,262 @@
...
+	err := obj.CallWithContext(ctx, propsIface+".Get", 0, sniItemIface, "Id").Store(&variant)
+	if err != nil {
+		return ""
+	}
+
+	id, _ := variant.Value().(string)
+	return id
+}
+
+func (m *Manager) registerSNI(name, objPath string) {
```
bbedward: I think the objPath should be removed, because it isnt used here besides the log.

PR #2141, opened 2026-04-03
fix: use prefer-light and dconf for GTK4 color-scheme refresh (niri compatibility)

### LINKED ISSUE #2140: GTK4 color-scheme refresh fails on niri: use prefer-light and dconf instead of default/gsettings
## Problem

Experienced this on my local setup (DMS + Niri), the GTK4 color-scheme refresh introduced in #1643 does not work correctly.

### `default` as intermediate toggle value

`refreshGTK4()` currently toggles through `default` when the current scheme is `prefer-dark`:

```go
if current == "prefer-dark" {
    toggle = "default"  // ← problematic
} else {
    toggle = "prefer-dark"
}
```

It seems on niri, `default` is not treated the same as on GNOME — it may not trigger the GTK4 style reload that the toggle is meant to force. Using `prefer-light` as the intermediate value is explicit and unambiguous, and produces the same toggling effect.

`dconf write` is more universal: it writes directly to the DConf database without needing schema definitions, and is available on virtually any system that runs GTK apps.

This was already noted in the comments on #1643:

> Now that I think about it, we should use `dconf write /org/gnome/desktop/interface/color-scheme '"prefer-light"'` instead of gsettings, because it is an "universal" way of doing it. For example, in NixOS it is a pain to setup gsettings schemas, and some minimalistic distro won't include it, however dconf probably is universal, or at least much more common.

## Proposed Fix

- Replace `"default"` with `"prefer-light"` as the intermediate toggle value in `refreshGTK4()` and as the scheme value in `syncColorScheme()` for light mode.
- Add a `DconfSet` utility that tries `dconf write` first (falling back to `gsettings`), and use it for all `color-scheme` writes.

A pull request with the fix is attached.

auricom:
Fixes #2140

## Summary

- Replace `"default"` with `"prefer-light"` as the intermediate toggle value in `refreshGTK4()` — avoids ambiguity on compositors like niri where `default` may not trigger a style reload
- Replace `"default"` with `"prefer-light"` in `syncColorScheme()` for light mode — explicit value is safer across environments
- Add `utils.DconfSet()` which tries `dconf write` first (falling back to `gsettings`), and use it for all `color-scheme` writes — `dconf` is more universal and does not require GSettings schema setup (niri, NixOS, minimal distros)

## Test plan

- [ ] Apply theme in dark mode on niri — GTK4 apps should reload their color scheme
- [ ] Apply theme in light mode on niri — GTK4 apps should reload their color scheme
- [ ] Verify behavior is unchanged on a standard GNOME session
- [ ] Verify `dconf` is tried before `gsettings` for `color-scheme` writes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/core/internal/matugen/matugen.go
+++ b/core/internal/matugen/matugen.go
@@ -810,17 +810,17 @@ func refreshGTK4() {
 
 	var toggle string
 	if current == "prefer-dark" {
-		toggle = "default"
+		toggle = "prefer-light"
 	} else {
 		toggle = "prefer-dark"
 	}
 
-	if err := utils.GsettingsSet("org.gnome.desktop.interface", "color-scheme", toggle); err != nil {
+	if err := utils.DconfSet("org.gnome.desktop.interface", "color-scheme", toggle); err != nil {
 		log.Warnf("Failed to toggle color-scheme for GTK4 refresh: %v", err)
 		return
 	}
 	time.Sleep(50 * time.Millisecond)
-	if err := utils.GsettingsSet("org.gnome.desktop.interface", "color-scheme", current); err != nil {
+	if err := utils.DconfSet("org.gnome.desktop.interface", "color-scheme", current); err != nil {
 		log.Warnf("Failed to restore color-scheme for GTK4 refresh: %v", err)
 	}
 }
@@ -867,10 +867,10 @@ func signalByName(name string, sig syscall.Signal) {
 func syncColorScheme(mode ColorMode) {
 	scheme := "prefer-dark"
 	if mode == ColorModeLight {
-		scheme = "default"
+		scheme = "prefer-light"
 	}
 
-	if err := utils.GsettingsSet("org.gnome.desktop.interface", "color-scheme", scheme); err != nil {
+	if err := utils.DconfSet("org.gnome.desktop.interface", "color-scheme", scheme); err != nil {
 		log.Warnf("Failed to sync color-scheme: %v", err)
 	}
 }
--- a/core/internal/utils/gsettings.go
+++ b/core/internal/utils/gsettings.go
@@ -29,3 +29,13 @@ func GsettingsSet(schema, key, value string) error {
 	}
 	return exec.Command("dconf", "write", dconfPath(schema, key), "'"+value+"'").Run()
 }
+
+// DconfSet writes a value using dconf write, falling back to gsettings.
+// Prefer this for settings like color-scheme where dconf is more universally
+// available (e.g. on niri, NixOS) and does not require GSettings schema setup.
+func DconfSet(schema, key, value string) error {
+	if err := exec.Command("dconf", "write", dconfPath(schema, key), "'"+value+"'").Run(); err == nil {
+		return nil
+	}
+	return exec.Command("gsettings", "set", schema, key, value).Run()
+}


bbedward:
I was under the impression that prefer-light was more of a accessibility thing and the preferred preference for light mode was default, at least thats  what I gained from yalter comments awhile back when we used to use prefer-light

auricom:
I'm using Automatic Color Mode based on Location on DMS, and it works well. Every night, DMS and system applications switch to dark mode, and it switches to light mode during the day as expected.


However, my Flatpak applications (Zen, Vestkop, Zed, Vscode, Slack, etc...), all configured with themes synchronized with system, are staying in dark mode everytime.

The fix I found is to manually launch `dconf write /org/gnome/desktop/interface/color-scheme '"prefer-dark"'` or `dconf write /org/gnome/desktop/interface/color-scheme '"prefer-light"'` based on what time of the day it is ; because 'default' equals to 'prefer-dark' on my system.

I thought reading that comment that using `default` was not optimal compared to `prefer-light`, but I apologize if that's not the case.

Do you have an idea how I could manually trigger a script when DMS changes from dark to light ? I could fix this locally if upstream should stay like this.
Many thanks

JEB-Bem:
Hi, I think I’m encountering a related issue that might be connected to how DMS currently writes the intermediate value `"default"` to `org.gnome.desktop.interface color-scheme`.

On my system, when DMS switches wallpapers, it updates the color scheme twice in sequence:

Example observed via:

```
gsettings monitor org.gnome.desktop.interface
```

Output:

```
gtk-theme: ''
gtk-theme: 'adw-gtk3-dark'
color-scheme: 'default'
color-scheme: 'prefer-dark'
```

Because Chromium/Chrome on Wayland follows `org.freedesktop.appearance color-scheme` via xdg-desktop-portal, this causes websites using `prefers-color-scheme` to switch themes twice during wallpaper changes.

Observed behavior:

* When currently in **light mode**:
  wallpaper change triggers:

  ```
  light → dark → light
  ```

  visible as a flash

* When currently in **dark mode**:
  wallpaper change triggers:

  ```
  dark → light → dark
  ```

So effectively Chrome is reacting correctly to two system theme changes, but the intermediate `"default"` value causes an unintended transition state.

Interestingly, some websites do not visibly flicker (for example GitHub). My assumption is that those sites do not dynamically react to runtime prefers-color-scheme changes, while others (e.g. ChatGPT or some personal blog frameworks) do.

My question:

Is this behavior caused by DMS writing `"default"` before applying the final value (`prefer-dark` / `prefer-light`)?

If yes, is there a recommended way to avoid the intermediate `"default"` transition while still keeping Chrome synced with the system theme automatically?

I would like Chrome to continue following the system appearance via xdg-desktop-portal, but without the intermediate flicker.

System environment:

Compositor:
Hyprland 0.54.3

Portal:
xdg-desktop-portal (GTK backend)

Shell:
DankMaterialShell v1.5-beta
Quickshell 0.2.1

OS:
Arch Linux

Browser:
Google Chrome 146.0.7680.177 (Wayland)

The portal updates can be confirmed with:

```
dbus-monitor "interface='org.freedesktop.portal.Settings'"
```

Thanks!

Spoch-dev:
Closing

bbedward:
The behavior of writing twice was a hack to try to get GTK apps to refresh in real time, it was removed in 1.5. Or made disabled by default.

Closing this as `default` and `prefer-dark` is the correct path, though

PR #2155, opened 2026-04-05
feature: add login sound functionality and settings entry

### LINKED ISSUE #1838: [Enchancement] Add the option to play log-in sounds from the sound theme
### Feature Description

A toggle (or custom field) that will play a login sound upon startup of the shell after the greeter

### Compositor(s)

All compositors

### Proposed Solution

Add a new event in quickshell/services/AudioService and update the corresponding sound path in getSoundPath along with the functions needed (function playLoginSound, and accompanying change in createSoundPlayers) and a hook to play that sound on login

_No response_

IrisHormoviti:
Closes: https://github.com/AvengeMedia/DankMaterialShell/issues/1838 

The login sound will play half a second after DMS starts. It worked with smaller delays, but I'm just being safe.
I have added a new settings entry, but not any translations for it.

I've also added a now-deprecated desktop-login sound from the XDG sound theme as a fallback. I want feedback of whether this should be included or not.


purian23:
Thanks, this looks good overall! One request, keep it off by default. 
Probably by `property bool soundLogin: false` in SettingsData.

IrisHormoviti:
> Thanks, this looks good overall! One request, keep it off by default. Probably by `property bool soundLogin: false` in SettingsData.

Alright, I can do that

quickshell/DMSShell.qml
```diff
@@ -221,10 +221,23 @@ Item {
...
+            AudioService.loginSoundRequested = true;
+            AudioService.playLoginSoundIfEnabled();
+        }
+    }
+
     Component.onCompleted: {
         dockRecreateDebounce.start();
         // Force PolkitService singleton to initialize
         PolkitService.polkitAvailable;
+        loginSoundTimer.start();
```
bbedward, line 239: The problem with this and the reason I didn't do this before is that it becomes a "dms started sound" instead of a "login/session started sound"

I think there's 2 potential options

1. Play login sound as a oneshot systemd user unit, although that makes it not easy to configure from a DMS GUI
2. Create a tmp file to indicate the sound was already played and not play it again

2 is probably a better solution, that would look like:

1. Startup DMS, check for login sound enabled + existence of XDG_RUNTIME_DIR/dms.login-sound-$XDG_SESSION_ID
2. If file doesnt exist and sound is enabled, create the file and play sound

That would guard against a standard dms restart playing the sound.

bbedward (changes requested):

quickshell/DMSShell.qml
IrisHormoviti: 1 sounds a bit convoluted, but I could implement 2.

IrisHormoviti: other files by dms in XDG_RUNTIME_DIR start with `danklinux.`, perhaps i should use that instead

IrisHormoviti:
I have implemented the second solution. An empty file in now created when the sound plays in the runtime dir, and the sound doesn't play if it already exists.

jasonm23:
This seems to be missing from settings, and translations.  Can open an issue if needed.

Purian23:
> This seems to be missing from settings, and translations. Can open an issue if needed.

This is live in DMS git, I can't recall if we shipped it to stable yet. But it will be in the next one if we missed it. 

https://github.com/AvengeMedia/DankMaterialShell/pull/2155/changes#diff-234a486a57b1af3b9e09cd6d14293efc05357a449c3e0cffb2d5c292d2e7e275

jasonm23:
Thanks, I'll switch to -git release.

Purian23:
> Thanks, I'll switch to -git release.

There's a ton of extras and new features in dms-git over the latest stable release. Enjoy!

jasonm23:
> > Thanks, I'll switch to -git release.
> 
> There's a ton of extras and new features in dms-git over the latest stable release. Enjoy!

Looking forward to it, nice work!

PR #2217, opened 2026-04-13
Fix ddc brightness not applying because process exits before debounce timer runs

DavutHaxor:
The debounce timer waits 200ms before applying the brightness. But on cli process exits after `SetBrightnessWithExponent` returns, leading to i2c write never happening.

When i wrote a simple debug log to see what is the problem i saw that
❯ DMS_LOG_LEVEL=debug ./bin/dms brightness set ddc:i2c-3 50 --ddc
...
 DEBUG  go: !!! Attempting to set brightness VIA SYSFS for device: ddc:i2c-3
 DEBUG  go: sysfs.SetBrightness failed: <nil>
 DEBUG  go: initialized I2C-3 with brightness 100/100
 DEBUG  go: found DDC device on i2c-3
...
 DEBUG  go: !!! Attempting to set brightness VIA DDC for device: ddc:i2c-3
Set ddc:i2c-3 to 50%

But after running this it didn't do any change to brightness

❯ DMS_LOG_LEVEL=debug ./bin/dms brightness get ddc:i2c-3 --ddc
...
 DEBUG  go: initialized I2C-3 with brightness 100/100
 DEBUG  go: found DDC device on i2c-3
...
ddc:i2c-3: 100% (100/100)
 

Creating a WaitPending function in core/internal/server/brightness/ddc.go to use before process return fixed it when i use from cli. The issue doesn't persist if i just use brightness slider.

### DIFF
--- a/core/internal/server/brightness/ddc.go
+++ b/core/internal/server/brightness/ddc.go
@@ -218,7 +218,9 @@ func (b *DDCBackend) SetBrightnessWithExponent(id string, value int, exponential
 	if timer, exists := b.debounceTimers[id]; exists {
 		timer.Reset(200 * time.Millisecond)
 	} else {
+		b.debounceWg.Add(1)
 		b.debounceTimers[id] = time.AfterFunc(200*time.Millisecond, func() {
+			defer b.debounceWg.Done()
 			b.debounceMutex.Lock()
 			pending, exists := b.debouncePending[id]
 			if exists {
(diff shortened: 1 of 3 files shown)


core/internal/server/brightness/ddc.go
```diff
@@ -490,5 +490,32 @@ func (b *DDCBackend) valueToPercent(value int, max int, exponential bool) int {
 	return percent
 }
 
+func (b *DDCBackend) WaitPending() {
```
bbedward, line 495: This needs a timeout, if i2c hangs or something it will block forever.

core/internal/server/brightness/ddc.go
```diff
@@ -490,5 +490,32 @@ func (b *DDCBackend) valueToPercent(value int, max int, exponential bool) int {
...
+    b.debounceMutex.Lock()
+    hasPending := len(b.debouncePending) > 0
+    b.debounceMutex.Unlock()
+
+    if !hasPending {
+        return
+    }
+
+    // Poll until all pending sets are flushed
+    go func() {
```
bbedward: Might be a better pattern to use a waitgroup here, add a `sync.WaitGroup` to the `DDCBackend` struct and then in the SetBrightnessWithExponent:

```
      b.debounceWg.Add(1)
      b.debounceTimers[id] = time.AfterFunc(200*time.Millisecond, func() {
          defer b.debounceWg.Done()

          ....
 .....

And then here

```go
  func (b *DDCBackend) WaitPending() {
        done := make(chan struct{})
        go func() {
                b.debounceWg.Wait()
                close(done)
        }()

        select {
        case <-done:
        case <-time.After(5 * time.Second):
                log.Debug("WaitPending timed out waiting for DDC writes")
        }
  }
```

bbedward (changes requested):

DavutHaxor:
Done

core/internal/server/brightness/ddc.go
```diff
@@ -490,5 +492,19 @@ func (b *DDCBackend) valueToPercent(value int, max int, exponential bool) int {
...
+    done := make(chan struct{})
+    go func() {
+        b.debounceWg.Wait()
+        close(done)
+    }()
+
+	select {
+	case <- done:
+	case <- time.After(5 * time.Second):
+			log.Debug("WaitPending timed out waiting for DDC writes")
```
bbedward: Overall would be better if we sent errors back over the channel so the response could be a timeout error instead of reporting success , but not a big deal overall.

bbedward (approved):

PR #2241, opened 2026-04-18
feat(system-tray): add optional monochrome icons setting

rollecode:
## Summary

Adds an opt-in "Monochrome System Tray Icons" toggle under Settings → Dank Bar → Settings (right after "Maximize Detection"). When enabled, every system tray icon is desaturated via MultiEffect { saturation: -1 }, producing a uniform monochrome bar that matches minimal / one-color themes.

Default is false, so existing users see no change.

## Changes

- quickshell/Common/settings/SettingsSpec.js - add systemTrayMonochromeIcons: { def: false }
- quickshell/Common/SettingsData.qml - add property bool systemTrayMonochromeIcons: false
- quickshell/Modules/DankBar/Widgets/SystemTrayBar.qml - wire layer.enabled + MultiEffect with saturation: -1 on the four IconImage blocks (horizontal bar, vertical bar, overflow menu item, inline expansion icon)
- quickshell/Modules/Settings/DankBarTab.qml - new SettingsToggleCard after "Maximize Detection"

## Test plan

-  [x] Toggle in Settings → Dank Bar appears after "Maximize Detection"
-  [x] Enabling the toggle desaturates all tray icons (Signal, 1Password, Spotify, Element, etc.) to grayscale
-  [x] Disabling restores color
-  [ ] Overflow menu icons also respect the setting

Disabled (default):

Enabled (toggled):

Icon: https://fonts.google.com/icons?selected=Material+Symbols+Rounded:filter_b_and_w

### DIFF
--- a/quickshell/Common/SettingsData.qml
+++ b/quickshell/Common/SettingsData.qml
@@ -211,6 +211,7 @@ Singleton {
     property int selectedGpuIndex: 0
     property var enabledGpuPciIds: []
     property bool showSystemTray: true
+    property bool systemTrayMonochromeIcons: false
     property bool showClock: true
     property bool showNotificationButton: true
     property bool showBattery: true
--- a/quickshell/Common/settings/SettingsSpec.js
+++ b/quickshell/Common/settings/SettingsSpec.js
@@ -79,6 +79,7 @@ var SPEC = {
     selectedGpuIndex: { def: 0 },
     enabledGpuPciIds: { def: [] },
     showSystemTray: { def: true },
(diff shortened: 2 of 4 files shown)


arora-archit:
i think apart from monochrome, having an option to choose filter from primary, secondary, surface colours etc. will be nice to have

rollecode:
> i think apart from monochrome, having an option to choose filter from primary, secondary, surface colours etc. will be nice to have

Good point. It seems this landed. Thanks @bbedward, working great :heart:

PR #2244, opened 2026-04-19
flake: let module callers supply pkgs so overlays reach the build

amaanq:
The nixosModule/homeModule path previously called `buildDmsPkgs pkgs` but internally referenced `self.packages.${system}.default`, which was instantiated via `nixpkgs.legacyPackages`, an unoverlayed pkgs. That meant downstream flakes couldn't reach through their own overlays to the dms-shell build (e.g. to swap `kdePackages.sonnet` or trim perl out of the aspell closure).

Extract the derivation as `mkDmsShell = pkgs: ...` at the top-level `let`, and call it from both `packages.${system}.dms-shell` (for direct consumers of the flake) and `buildDmsPkgs pkgs` (for module consumers, which now pass in the system's overlayed pkgs).

Also re-checks overrideAttrs / .override still work: `mkDmsShell pkgs` is the same `pkgs.lib.makeOverridable` wrapper as before, just parameterized on the caller's pkgs instance.

Thanks!


flake.nix
```diff
@@ -73,10 +69,18 @@
...
+
+      # Extracted so both `packages.<system>.dms-shell` (for direct consumers:
+      # `nix build .#`) and `buildDmsPkgs pkgs` (for NixOS/home-manager modules)
+      # build the derivation with a caller-supplied `pkgs`. Previously the
+      # module path reused the pre-built `self.packages.<system>.default`,
+      # which was instantiated from `nixpkgs.legacyPackages` — an unoverlayed
+      # pkgs instance. That made it impossible for downstream flakes to reach
+      # through their own overlays to the build (e.g. to swap kdePackages.sonnet
+      # or trim perl out of aspell). With this refactor, the module's `pkgs`
+      # arg (which carries the user's overlays) is used for the actual build.
```
LuckShiba:
```suggestion
      # Allows downstream modules to provide their own 'pkgs' (with overlays)
      # instead of being forced to use the flake's locked nixpkgs.
```

LuckShiba (changes requested):
The changes make sense! I just think the comment you added is a bit way too verbose. What do you think about this?

flake.nix
amaanq: agreed, that is way too verbose. thanks!

LuckShiba:
oops, accidentally closed 😅

LuckShiba (approved):

LuckShiba:
Thanks!

PR #2265, opened 2026-04-25
feat(system-tray): add icon tinting

arora-archit:
Added Primary and Secondary Tints along with monochrome with a saturation and strength slider

image


arora-archit:
why did settings reorg had to happen at the same time
i'll resolve conflicts and update in a few hours

PR #2275, opened 2026-04-26
feat(display): Fix and implement display auto-switching with JSON profile storage

DK0280705:
The display auto-configuration feature was previously broken and explicitly disabled throughout the codebase — the UI toggle was hidden, the IPC toggleAuto command returned a hardcoded error, and the feature never ran at startup. This PR fixes all of that and delivers a fully working auto-switching system backed by a new compositor-neutral JSON profile format.

### Changes:

- Profiles are now stored as compositor-neutral JSON. Compositor-specific config files (Hyprland `.conf`, Niri `.kdl`, DWL `.conf`) are generated from this JSON at apply time, unifying profile logic across compositors.
- `DisplayConfigState` is force-initialized at shell startup (`DMSShell.qml`) so auto-config runs on login.
- IPC `toggleAuto` fixed. Previously returned a hardcoded error, now properly toggles `SettingsData.displayProfileAutoSelect`, persists it, and calls `autoSelectProfile()`.
- IPC `status` fixed. It was hardcoded to report auto: "off"; now reflects the real setting.
- The auto-select toggle is visible and functional, wired to `SettingsData.displayProfileAutoSelect` and triggering `applyAutoConfig()` on enable.
- Profile controls disabled during auto-select. The new/delete/edit buttons and profile dropdown are disabled while auto-select is on to prevent conflicts.
- "Edit Monitors" button. New toolbar button opens a per-profile monitor inclusion dialog via `getProfileMonitorInclusion()`.
- Dropdown shows "Auto"  when auto-select is active, the profile dropdown displays "Auto" via a reactive Binding.
- Added `canDisableOutput()` to prevent disabling the last active monitor. 
- Profiles are validated to ensure at least one output is enabled. If all outputs are somehow disabled, the first is re-enabled automatically.


bbedward:
Have you tested this on niri? With hot plugging displays. The reason it was temp disabled mainly was because of that, there would be crashes or display freezes on resume I think from some race between niri and dms writing display configurations.

DK0280705:
> Have you tested this on niri? With hot plugging displays. The reason it was temp disabled mainly was because of that, there would be crashes or display freezes on resume I think from some race between niri and dms writing display configurations.

Yes, actually i only used niri to test automatic display configuration, so i don't know if this going to work on hyprland or dwl/mango too. 

I'm using this configuration for a while and there are no problems so far. Probably only when i suddenly unplug my external monitor from my laptop, some electron apps like vscode or discord freeze then crash, but that's not from the dms side as they also crash when i test with [kanshi](https://sr.ht/~emersion/kanshi/). 

I don't think the problem comes from the race between niri and dms(?)
Iirc the previous code creates callback loop between niri state and display config state. Also `DisplayConfigState` was not initialized at startup, so auto profile changes won't work until you open settings.

quickshell/Modules/Settings/DisplayConfig/DisplayConfigState.qml
```diff
@@ -71,224 +77,535 @@ Singleton {
...
+                return;
             }
-            validatedProfiles = validated;
-            matchedProfile = findMatchingProfile();
+            try {
+                monitorsCache = JSON.parse(content);
+                if (!Array.isArray(monitorsCache.configurations))
+                    monitorsCache.configurations = [];
+            } catch (e) {
+                console.warn("Failed to parse monitors.json, using empty config", e);
```
bbedward: In master there is a new scoped log service, I plan to add a pre-commit check for it but can we ensure that all console.log/info/warn/error use it?

Goes like

```qml
readonly property var log: Log.scoped("DisplayConfig")

....

log.warn("Failed to parse monitors.json, using empty config")
```

bbedward (changes requested):

quickshell/Modules/Settings/DisplayConfig/DisplayConfigState.qml
DK0280705: sure one

DK0280705 (commented):
Changed all console logs with the new log service.

PR #2276, opened 2026-04-26
Fix: close notification center after clicking action buttons

nficca:
Closes #2178

Close the notification center when clicking notification action buttons, matching the existing behavior of the popup toast, keyboard navigation (`NotificationKeyboardController`), and control center widgets. This lets the relevant app receive focus after the action is invoked, rather than being blocked by the layer-shell surface.

Also sync `notificationHistoryVisible` from `onShouldBeVisibleChanged` so that programmatic `close()` calls (like from `PopoutService.closeNotificationCenter()`) keep the property consistent. Without this, the bell toggle requires two presses to reopen after a programmatic close.

### Before

https://github.com/user-attachments/assets/63ecffec-30de-40c7-85f4-56e6d8f65325

### After

https://github.com/user-attachments/assets/0f2d5a88-3bed-4f94-925f-d567672736c1

### DIFF
--- a/quickshell/Modules/Notifications/Center/NotificationCard.qml
+++ b/quickshell/Modules/Notifications/Center/NotificationCard.qml
@@ -727,8 +727,10 @@ Rectangle {
                                                     onEntered: parent.isHovered = true
                                                     onExited: parent.isHovered = false
                                                     onClicked: {
-                                                        if (modelData && modelData.invoke)
+                                                        if (modelData && modelData.invoke) {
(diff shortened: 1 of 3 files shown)


kanghengliu:
how does this handle the 'dismiss' button? Normally you wouldn't want to dismiss the whole inbox when clicking 'dismiss' on a single notification.

bbedward:
@kanghengliu  dismiss button doesnt close the center, so good there.

PR #2277, opened 2026-04-26
dock: add trash bin button

kanghengliu:
# Planned
- [x] icon reflects content- filled/empty
- [x] multiple file manager support with nautilus as default, no-op as fallback
- [x] settings at dock tab
- [x] trash icon context menu
- [x] custom open file manager command

# Out of scope
- Support for built-in file browser
  - trash logic according to xdg specs
  - clear trash button
  - entry context menu
  
# Demo
https://github.com/user-attachments/assets/4f61beb7-9587-41f0-9d7f-e86613205805

### DIFF
--- a/quickshell/Common/SettingsData.qml
+++ b/quickshell/Common/SettingsData.qml
@@ -547,6 +547,7 @@ Singleton {
     property bool dockShowOverflowBadge: true
     property bool dockShowTrash: false
     property string dockTrashFileManager: "nautilus"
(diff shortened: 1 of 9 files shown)


kanghengliu:
Okay, i'm happy with where I'm at now. Had to drop support for the built-in file chooser, as that lack the proper implementation for xdg trash spec and probably is not fitting to be left long running as it's part of the main shell. (Imagine if the dumpster/archive had a fire and collapsed)

PR #2312, opened 2026-04-29
Fix VPN UI for active transient entries

### LINKED ISSUE #2311: VPN UI ignores active connections without importable profiles
## Summary

The VPN UI currently assumes that visible VPN entries always come from importable VPN profiles.

If the backend reports an active VPN or tunnel connection that does not have a corresponding importable profile entry, the popout and settings UI can still show an empty state such as "No VPN profiles", even though a connection is active.

## Steps to reproduce

1. Have DMS report an active VPN or tunnel connection through the network service.
2. Ensure that active connection is not present in `network.vpn.profiles`.
3. Open the VPN popout or the VPN section in Settings.

## Actual behavior

The UI shows the empty "No VPN profiles" state even while an active VPN or tunnel connection exists.

## Expected behavior

Active VPN or tunnel connections should still be shown in the UI even when they are not backed by an importable profile.

Entries that are not importable or manageable profiles should also avoid showing edit, delete, or expand actions intended for imported VPN configs.

## Notes

This appears to affect both the popout VPN list and the Settings VPN list, since both currently render from the profiles model only.

louzt:
## Summary

Fixes the VPN UI empty-state behavior when the backend reports active VPN or tunnel connections that are not represented as importable profiles.

Fixes #2311

## What changed

- Merge active VPN entries that are missing from the profiles list into the visible VPN model.
- Mark those entries as transient so the UI can show them without treating them like imported configs.
- Hide expand and delete actions for transient entries in both the control center popout and Settings.
- Preserve existing behavior for regular importable VPN profiles.

## Validation

- QML diagnostics reported no errors on the touched files.
- Repository lint bootstrap succeeded, but the local `lint-qml` target could not complete in this environment because its `qmllint` wrapper resolves to a missing Qt5 path.

### DIFF
--- a/quickshell/Modules/Settings/NetworkTab.qml
+++ b/quickshell/Modules/Settings/NetworkTab.qml
@@ -1697,8 +1697,11 @@ Item {
                                 required property int index
 
                                 readonly property bool isActive: DMSNetworkService.isActiveUuid(modelData.uuid)
+                                readonly property bool isTransient: !!modelData.transient
+                                readonly property bool canExpand: modelData.canExpand !== false
+                                readonly property bool canDelete: modelData.canDelete !== false
                                 readonly property bool isExpanded: networkTab.expandedVpnUuid === modelData.uuid
-                                readonly property var configData: isExpanded ? VPNService.editConfig : null
+                                readonly property var configData: (!isTransient && isExpanded) ? VPNService.editConfig : null
 
                                 width: parent.width
                                 height: isExpanded ? 56 + vpnExpandedContent.height : 56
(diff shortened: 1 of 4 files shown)


louzt:
I noticed this bug while validating a self-hosted VPN workflow outside the normal UI import path.

In that setup, the backend state detection was already working correctly and the connection showed up as active, but the VPN UI still rendered the empty "No VPN profiles" state because it only reflected importable profile entries.

That made it easier to separate the problem into two parts:
- backend detection was already correct
- the UI model was treating importable profiles as the only visible source of truth

This PR is meant to address that UI-side gap without tying the fix to any specific provider or private setup.

PR #2321, opened 2026-04-30
support QT_QPA_PLATFORMTHEME=kde

fishman:
dms already supports Kcolorscheme. QT_QPA_PLATFORMTHEME=kde support allows you to dynamically switch light and dark mode of QT apps without needing a restart.


core/internal/matugen/matugen.go
```diff
@@ -182,6 +181,8 @@ func Run(opts Options) error {
 		syncColorScheme(opts.Mode)
 	}
 
+	applyKDEColorScheme(opts.Mode)
```
bbedward, line 185: The GTK version is gated on whether or not the user is actually using the DMS colors or not, I think we need something equivalent. Like we wouldn't wanna apply the scheme if they're using something besides DankMatugen

bbedward (changes requested):

core/internal/matugen/matugen.go
microquake-buildbot: But shouldn't that be the kde color sync? Otherwise what's the point in syncing kde colors ?

bbedward: The matugen is running unconditionally on theme changes I mean. If someone has Breeze configured won't it change it to DankMatugen even if they didn't want that? Some users want dms without all the themes it provides

fishman: Well the color scheme works fine with Breeze that's specifically what Kcolorscheme is designed for. But maybe I should check if QT_QPA_PLATFORMTHEME is set to kde with QT_STYLE_OVERRIDE Breeze. I imagine it might cause problems with kvantum. But the people that use kvantum seem to set QT_QPA_PLATFORMTHEME to qt6ct

bbedward: I guess I mean color scheme, if they are using a different color scheme like Breeze Dark and they want to keep it, this path will unconditionally change it to DankMatugen. So we should add a gate for current color scheme, if it's a DMS one then we can flip it to dark/light

Spoch-dev:
Closing

bbedward:
I rebased in a feature branch and added the conditions in 61510c2186eeddc9cd06a891d1b1f8b287199d9e

PR #2322, opened 2026-05-01
Add inline sliders to Control Center toggle buttons

elweday1:
## Summary
Add inline sliders to Control Center toggle buttons for audio output, audio input, and brightness.

## Behavior
- **audioOutput/audioInput**: Slider shows current volume, adjusts on change
- **brightnessSlider**: Uses same device resolution logic as original `BrightnessSliderRow` component
  - Respects screen-to-device pins (`SettingsData.brightnessDevicePins`)
  - Falls back to widget deviceName, then currentDevice, then first available device
  - Icons change dynamically based on brightness level (low/medium/high)
  - No title (primaryText empty) so slider centers like original component

## Implementation
- `CompoundPill`: Added `hasSlider`, `sliderValue`, `sliderMin`, `sliderMax`, `sliderAction` properties
- `DragDropGrid`: Added slider logic for widget types + icon/primaryText cases for brightnessSlider

## Motivation
This approach is **composable** - slider behavior added via properties to existing CompoundPill, no new widget types needed.

## Screenshots
Before:
ss-20260501-084910

After:
ss-20260501-085820

### DIFF
--- a/quickshell/Modules/ControlCenter/Components/DragDropGrid.qml
+++ b/quickshell/Modules/ControlCenter/Components/DragDropGrid.qml
@@ -150,13 +150,11 @@ Column {
                                 return builtinPluginWidgetComponent;
                             } else if (id.startsWith("plugin_")) {
                                 return pluginWidgetComponent;
-                            } else if (id === "wifi" || id === "bluetooth" || id === "audioOutput" || id === "audioInput") {
+} else if (id === "wifi" || id === "bluetooth" || id === "audioOutput" || id === "audioInput" || id === "brightnessSlider") {
                                 return compoundPillComponent;
                             } else if (id === "volumeSlider") {
                                 return audioSliderComponent;
                             } else if (id === "brightnessSlider") {
-                                return brightnessSliderComponent;
-                            } else if (id === "inputVolumeSlider") {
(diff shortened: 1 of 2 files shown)


bbedward:
The style is ok, but the brightness one is broken - the detail menu doesnt open that lets you choose brightness device

bbedward:
Also I think it'd be better if it was just new widgets available from the customize menu, rather than overriding any existing ones.

elweday1:
> Also I think it'd be better if it was just new widgets available from the customize menu, rather than overriding any existing ones.

Thank you for the reply @bbedward, 
let me elaborate on my thoughts with this was to first consolidate some of the currently existing ControlCenter items/toggles visually and implementation wise.

so later if it is planned for the plugin system to extend new ControlCenter items, like #1875, it wouldnt be needed to patch the core, rather add something new to the [dms-plugin-registry](https://github.com/AvengeMedia/dms-plugin-registry).

Purian23:
Hi there, thanks for the PR but we're going to pass at this time. It changes our defaults which are Material 3 inspired. This isn't what we'd want to ship without an option to change them. I believe this as a plugin would be the best fit. There are plans to update some of the plugins core functions to allow more broad implementations in the future.

PR #2339, opened 2026-05-03
lock: respect notification privacy mode

arora-archit:
Lock screen showed notification description irrespective of the privacy mode which isn't expected because lock screen is meant to keep your desktop private. Maybe a separate setting can be added to toggle privacy only for lock screen


bbedward:
There is an independent setting for lock screen privacy, basically with the display modes.

image

arora-archit:
Oh sorry! My bad, missed it completely. Thank you.

Issue #2348, opened 2026-05-04
[regression] bar/dock auto-hide on niri windowed-fullscreen since 31e60a3


benabraham:
### Compositor

Niri

### Distribution

NixOS

### Select your Installation Method

DankInstaller

### Was this your original Installation method?

Yes

### If no, specify

Distro Packaging (NixOS home-manager flake module from upstream); rebuilt from local fork for testing

### dms doctor -vC

<details>
<summary>Click to expand</summary>

```
## DMS Doctor Report

**System**
- [ok] Operating System: NixOS 25.11 (Xantusia)
  - Supported for runtime (install via NixOS module or Flake)
- [ok] Architecture: amd64
- [ok] Display Server: Wayland
  - WAYLAND_DISPLAY=wayland-1

**Versions**
- [ok] DMS CLI: dms 1.5-beta+date=2026-05-04_e807939
  - /nix/store/hxhl27dhbkl1sirigpfd6799x0pxz6gx-dms-shell-1.5-beta+date=2026-05-04_e807939/bin/.dms-wrapped
- [ok] Quickshell: Quickshell 0.2.1 (revision 783c953987dc56ff0601abe6845ed96f1d00495a, distributed by Official-Nix-Flake)
  - /nix/store/5mcnw8q9knsd6cgihpcp7jl6m7cy5219-quickshell-wrapped-0.2.1/bin/qs
- [ok] DMS Shell: 1.5-beta+date=2026-05-04_e807939
  - /nix/store/hxhl27dhbkl1sirigpfd6799x0pxz6gx-dms-shell-1.5-beta+date=2026-05-04_e807939/share/quickshell/dms

**Installation**
- [ok] DMS Configuration: Found
  - /nix/store/hxhl27dhbkl1sirigpfd6799x0pxz6gx-dms-shell-1.5-beta+date=2026-05-04_e807939/share/quickshell/dms
- [ok] shell.qml: Present
  - /nix/store/hxhl27dhbkl1sirigpfd6799x0pxz6gx-dms-shell-1.5-beta+date=2026-05-04_e807939/share/quickshell/dms/shell.qml
- [info] Install Type: Nix store
  - /nix/store/hxhl27dhbkl1sirigpfd6799x0pxz6gx-dms-shell-1.5-beta+date=2026-05-04_e807939/share/quickshell/dms

**Compositor**
[...]

PR #2349, opened 2026-05-04
feat(bar/dock): add 'Hide When Fullscreen' opt-out toggles

cwelsys:
Adds per-bar 'fullscreenDetection' and global 'dockHideOnFullscreen' settings to gate CompositorService.hasFullscreenToplevelOnScreen()


Purian23:
Hey @cwelsys, thanks for the PR! I had a fix inbound that may conflict with this from yesterday's release of DMS FrameMode as a follow-up. I will let you know what we land on, this may work very well too!

cwelsys:
my earlier PR #2091 excluded niri from this detection for two reasons: niri already renders real fullscreen above the top layer-shell layer (so the hide is redundant), and the Wayland protocol reports identical state for true fullscreen and windowed-fullscreen on niri (so the same check also hides the bar/dock for fake-fullscreen apps).

31e60a3d reintroduced niri detection inside `CompositorService.hasFullscreenToplevelOnScreen()`. The added activated check doesn't disambiguate windowed-fullscreen, which still reports `fullscreen: true` and `activated: true`.

I figured a good middle ground would be opt-out toggles instead of reverting, hence this PR.

A user-configurable per-app show/hide list would be the logical conclusion, independent of fullscreen detection. Happy to draft something to that extent if there is interest.

cwelsys:
> Hey @cwelsys, thanks for the PR! I had a fix inbound that may conflict with this from yesterday's release of DMS FrameMode as a follow-up. I will let you know what we land on, this may work very well too!

Thanks for the reply, and no worries, im happy to drop this if the inbound fix is better. Cheers :)

Purian23:
Thanks for the explanation, I was going to roll in an update to fix #2348 alongside. I only noticed the dbar would resurface upon the other monitor refocus shift, but your PR may very well handle it more explicitly in that direction and they quoted it haha. What do you think?

cwelsys:
> Thanks for the explanation, I was going to roll in an update to fix #2348 alongside. I only noticed the dbar would resurface upon the other monitor refocus shift, but your PR may very well handle it more explicitly in that direction and they quoted it haha. What do you think?

I think they can work in tandem. Question that stands out to me is the default, which may need to be compositor specific. 

Currently its true across the board. I can flip this PR so it's false on niri. I only use niri so its tough for me to comment on where this might be useful for other compositors.

Issue #2382, opened 2026-05-10
Battery Icon Design & Charging Indicator Visibility Improvement on Bar


faisalrbk:
### Feature Description

The current battery icon design  is not very clear or visually appealing. The battery icon is vertical, but a horizontal battery style would look much cleaner and more modern in the panel.

Also, the charging indicator/icon on the battery is extremely small. When the laptop is charging, there is almost no visible difference between charging and non-charging states. Literally, you have to look very carefully to notice it.

### with charging
<img width="501" height="323" alt="Image" src="https://github.com/user-attachments/assets/d3924b04-421c-4fd2-ae4b-6781a7acb428" />
### without charging
<img width="514" height="316" alt="Image" src="https://github.com/user-attachments/assets/a3a513ff-9f58-40e9-aada-b9f1e8e526bb" />

### Use Case

Battery status is something users constantly look at throughout the day. Right now, it is difficult to quickly identify whether the laptop is charging or not because the charging icon is too small and unclear.

A cleaner horizontal design with a more visible charging indicator would improve usability and accessibility.

Why is this feature important?
Better visibility
Easier to identify charging state instantly
More modern and polished UI appearance
Improved user experience, especially on smaller screens or high-resolution displays

### Compositor(s)

Niri

### Proposed Solution

Use a horizontal battery icon similar to modern desktop/mobile UIs and redesign the charging overlay/icon so it is clearly visible at a glance.

The charging state could also include:

a bold lightning icon
animated charging effect
different accent color while charging

### Alternatives/Existing Solutions

Many desktop environments and mobile operating systems use horizontal battery icons with clear charging indicators, including:

Android
GNOME extensions/widgets
Material Design battery icons
KDE Plasma battery indicators

PR #2385, opened 2026-05-11
feat(bar): horizontal battery layout with darker animated chargist bo…

Turbocyborg:
"Fixes issue #2382. Rotates the battery horizontally and adds an animated, darker charging bolt layered inside for better visibility."

### DIFF
--- a/quickshell/Modules/DankBar/Widgets/Battery.qml
+++ b/quickshell/Modules/DankBar/Widgets/Battery.qml
@@ -1,5 +1,4 @@
 import QtQuick
-import Quickshell.Services.UPower
 import qs.Common
 import qs.Modules.Plugins
 import qs.Services
@@ -7,12 +6,9 @@ import qs.Widgets
 
 BasePill {
     id: battery
-    readonly property var log: Log.scoped("Battery")
 
     property bool batteryPopupVisible: false
     property var popoutTarget: null
-
-    property real touchpadAccumulator: 0
 
     readonly property int barPosition: {
         switch (axis?.edge) {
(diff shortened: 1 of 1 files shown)


bbedward:
I'm not a big fan of this particular design, I think it'd be better to go for the modern android style although I couldnt see how to do it very easily with the latest material icons

prateekky:
Thanks for the feedback! I've made some changes that you might like!

Purian23:
Hi there, thanks for your PR. Unfortunately this will not work either. This would force users to install a qt5Compat package and would leave them broken in the meantime. We've went out of our way to avoid the need for QT5 anything in DMS. 

I think the best solution is to update this once more and make it a toggle option that keeps the current in place as we like it, and introduce the new variant. The next alternative is to leave it to our team and we will get a variant in place.

prateekky:
Good point on the Qt5 dependency! I've removed it entirely and rewrote the Android Expressive style using pure QML Rectangles (100% Qt6 safe).

I also added the toggle option you suggested. It checks for expressiveBatteryStyle and defaults to false, so the original design stays exactly as it is for everyone, but users can now easily opt-in to the new style. Let me know what you think!

Purian23:
We appreciate the efforts, but that isn't the direction we'd want to go either. 

- The PR doesn't load as you haven't rebased on the latest master DisplayConfig fix
- The user toggle needs added in 3 other source files to be persistent and user facing
- We have a different design in mind

Thank you again, don't give up on future efforts. At this time we're going to close this up and implement one from our side.

PR #2401, opened 2026-05-12
feat: add default value marker to DankSlider

hthienloc:
### What
This PR adds an optional visual marker for default values to the `DankSlider` component.

### Why
Providing a visual indicator for default values helps users see the recommended or initial state of a setting.

### How it works
- **Property**: `defaultValue` (int) sets the value where the marker will appear.
- **Toggle**: `showDefaultMarker` (bool) must be set to `true` to display the marker. It is `false` by default to maintain compatibility.
- **Aesthetics**: The marker is a 3px wide vertical line using `Theme.surfaceText` with 0.4 opacity.

### Visual Proof
image

### How tested
- Verified positioning logic for both standard and centered sliders.
- Verified visibility toggle.

### DIFF
--- a/quickshell/Modules/Plugins/SliderSetting.qml
+++ b/quickshell/Modules/Plugins/SliderSetting.qml
@@ -69,6 +69,7 @@ Column {
         value: root.value
         minimum: root.minimum
         maximum: root.maximum
+        defaultValue: root.defaultValue
         leftIcon: root.leftIcon
         rightIcon: root.rightIcon
         unit: root.unit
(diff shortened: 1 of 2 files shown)


Purian23:
Hi there, thank you for your PR efforts, but we're going to pass at this time. This setting goes against Material 3 settings already in place to reset defaults and becomes redundant. We appreciate your efforts though. 

image

hthienloc:
@Purian23 Thank you for letting me know, I appreciate it.

Issue #2358, opened 2026-05-06
Enterprise (802.1X) WiFi connect silently no-ops after "interactive: true" detection: no credential dialog shown


domenkozar:
### Compositor

Niri

### Distribution

NixOS

### Select your Installation Method

Distro Packaging

### Was this your original Installation method?

Yes

### dms doctor -vC

<details>
<summary>Click to expand</summary>
## DMS Doctor Report

**System**
- [ok] Operating System: NixOS 25.11 (Xantusia)
  - Supported for runtime (install via NixOS module or Flake)
- [ok] Architecture: amd64
- [ok] Display Server: Wayland
  - WAYLAND_DISPLAY=wayland-1

**Versions**
- [ok] DMS CLI: dms 1.4.6+date=2026-04-29_eb5afcd
  - /nix/store/pw4q5jqpjcrqnxpv2xby7ma075fphci6-dms-shell-1.4.6+date=2026-04-29_eb5afcd/bin/.dms-wrapped
- [ok] Quickshell: Quickshell 0.2.1 (revision 783c953987dc56ff0601abe6845ed96f1d00495a, distributed by Official-Nix-Flake)
  - /run/current-system/sw/bin/qs
- [ok] DMS Shell: 1.4.6+date=2026-04-29_eb5afcd
  - /nix/store/pw4q5jqpjcrqnxpv2xby7ma075fphci6-dms-shell-1.4.6+date=2026-04-29_eb5afcd/share/quickshell/dms

**Installation**
- [ok] DMS Configuration: Found
  - /nix/store/pw4q5jqpjcrqnxpv2xby7ma075fphci6-dms-shell-1.4.6+date=2026-04-29_eb5afcd/share/quickshell/dms
- [ok] shell.qml: Present
  - /nix/store/pw4q5jqpjcrqnxpv2xby7ma075fphci6-dms-shell-1.4.6+date=2026-04-29_eb5afcd/share/quickshell/dms/shell.qml
- [info] Install Type: Nix store
  - /nix/store/pw4q5jqpjcrqnxpv2xby7ma075fphci6-dms-shell-1.4.6+date=2026-04-29_eb5afcd/share/quickshell/dms

**Compositor**
- [ok] niri: 25.11
  - /run/current-system/sw/bin/niri
- [info] Active: niri
- [warn] Background Blur: Unsupported
  - Compositor does not support ext-background-effect-v1

**Quickshell Features**
- [ok] Polkit: Available
[...]

Issue #1520, opened 2026-01-27
Enterprise wifi keeps asking for credentials


willem640:
### Compositor

- [x] Niri
- [ ] Hyprland
- [ ] MangoWC (dwl)
- [ ] Sway

### Distribution

- [ ] Arch Linux
- [ ] CachyOS
- [x] Fedora
- [ ] NixOS
- [ ] Debian
- [ ] Ubuntu
- [ ] Gentoo
- [ ] OpenSUSE
- [ ] Other (specify below)

### dms doctor -v

                                  
██████╗  █████╗ ███╗   ██╗██╗  ██╗
██╔══██╗██╔══██╗████╗  ██║██║ ██╔╝
██║  ██║███████║██╔██╗ ██║█████╔╝ 
██║  ██║██╔══██║██║╚██╗██║██╔═██╗ 
██████╔╝██║  ██║██║ ╚████║██║  ██╗
╚═════╝ ╚═╝  ╚═╝╚═╝  ╚═══╝╚═╝  ╚═╝

 System Health Check
                    
──────────────────────────────────────

  System
    ● Operating System ····· Fedora Linux 43 (Workstation Edition)
      └─ ID: fedora, Version: 43, Arch: amd64
    ● Architecture ········· amd64
    ● Display Server ······· Wayland
      └─ WAYLAND_DISPLAY=wayland-1

  Versions
    ● DMS CLI ·············· dms v1.2.3
      └─ /usr/bin/dms
    ● Quickshell ··········· quickshell 0.2.1, revision 191085a8821b35680bba16ce5411fc9dbe912237, distributed by: Fedora COPR (avengemedia/quickshell)
      └─ /usr/bin/qs
    ● DMS Shell ············ 1.2.3
      └─ /usr/share/quickshell/dms

  Installation
    ● DMS Configuration ···· Found
      └─ /usr/share/quickshell/dms
    ● shell.qml ············ Present
      └─ /usr/share/quickshell/dms/shell.qml
    ○ Install Type ········· System package
      └─ /usr/share/quickshell/dms

  Compositor
    ● niri ················· 25.11
      └─ /usr/bin/niri
    ● mangowc ·············· 0.11.0
      └─ /usr/bin/mango
    ○ Active ··············· niri

  Quickshell Features
    ● Polkit ··············· Available
      └─ Authentication prompts
    ● IdleMonitor ·········· Available
      └─ Idle detection
    ● IdleInhibitor ········ Available
[...]

PR #2402, opened 2026-05-12
feat: retrieve WiFi/VPN secrets from D-Bus Secret Service with unlock support

### LINKED ISSUE #899: wifi not reconnecting on boot
## Compositor

- [ x ] niri
- [ ] Hyprland
- [ ] dwl (MangoWC)
- [ ] sway
- [ ] Other (specify)

## Distribution

Arch

## Description

Wifi won't reconnect on reboot. I have to use nmtui to connect to my wifi on each boot. The gui sometimes works, but the password prompt often doesn't pop up.

## Steps to Reproduce

1.
2.
3.

## Error Messages/Logs

```
Paste error messages or logs here
```

Mikilio:
#### Problem
Saved WiFi/VPN passwords in the keyring were never consulted — the app always prompted the user to re-enter them, even when GNOME Keyring had the correct secret.
#### Changes
- Secret Service client (secret_service.go): connects to `org.freedesktop.Secret`, searches items by connection UUID, and if locked triggers the system unlock dialog via `Prompt.Prompt()` before retrieving the secret
- Agent integration (agent_networkmanager.go): Phase 4 tries the keyring before showing the app's password prompt, skips cached VPN creds when REQUEST_NEW is set, guards cancellation against deleting pre-existing connection profiles
- QR code fallback (backend_networkmanager_wifi.go): reads PSK from `secret service` when `conn.GetSecrets` returns access-denied
#### Current state
Working. Keyring unlock dialog appears when needed, secrets are retrieved afterward, and the app prompt is bypassed.
#### Testing
Manual with locked/unlocked GNOME Keyring. Unit tests for ONLY_SYSTEM/NO_INTERACTION flags. Build & vet clean.

#### Review

Should more Unit tests be required to secure the correct behavior of this? Should the Architecture be improved by generalizing the Secret Service helper?

This may fix #899 #1520

I ask for a thorough review to deal with the above questions.

### DIFF
--- a/core/internal/server/network/secret_service.go
+++ b/core/internal/server/network/secret_service.go
@@ -1,0 +1,255 @@
+package network
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/AvengeMedia/DankMaterialShell/core/internal/log"
+	"github.com/godbus/dbus/v5"
+)
+
+const (
+	secretServiceBusName = "org.freedesktop.secrets"
+	secretServicePath    = "/org/freedesktop/secrets"
+	secretServiceIface   = "org.freedesktop.Secret.Service"
+	secretItemIface      = "org.freedesktop.Secret.Item"
+	secretPromptIface    = "org.freedesktop.Secret.Prompt"
+)
+
+type secretServiceSession struct {
+	conn        *dbus.Conn
+	svc         dbus.BusObject
+	sessionPath dbus.ObjectPath
+}
+
+func openSecretService() (*secretServiceSession, error) {
+	c, err := dbus.ConnectSessionBus()
+	if err != nil {
+		return nil, err
+	}
+
+	svc := c.Object(secretServiceBusName, dbus.ObjectPath(secretServicePath))
+
+	var sessionPath dbus.ObjectPath
+	call := svc.Call(secretServiceIface+".OpenSession", 0, "plain", dbus.MakeVariant(""))
+	if call.Err != nil {
+		c.Close()
+		return nil, call.Err
+	}
+	if err := call.Store(new(dbus.Variant), &sessionPath); err != nil {
+		c.Close()
+		return nil, err
+	}
+
+	return &secretServiceSession{
+		conn:        c,
+		svc:         svc,
+		sessionPath: sessionPath,
+	}, nil
+}
+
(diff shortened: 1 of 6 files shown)


core/internal/server/network/secret_service.go
```diff
@@ -0,0 +1,270 @@
...
+			close(done)
+		}
+		completedMu.Unlock()
+		return err
+	}
+
+	select {
+	case <-done:
+		return nil
+	case <-time.After(120 * time.Second):
```
bbedward: This should close the done channel, but I think this entire function can probably be simplified quite a bit without a watcher.

core/internal/server/network/secret_service.go
```diff
@@ -0,0 +1,270 @@
...
+	done := make(chan struct{})
+	ch := make(chan *dbus.Signal, 10)
+	s.conn.Signal(ch)
+
+	var completedMu sync.Mutex
+	completedOnce := false
+
+	go func() {
+		defer s.conn.RemoveSignal(ch)
+		defer func() { _ = recover() }()
```
bbedward: I think this is potentially suppressing the panic source, which would only be operating on an already closed channel. Can probably avoid that by re-working the general structure of that funciton.

bbedward (changes requested):

Mikilio:
I restructured the function. Sorry if the code style is weird. I am new to go and don't have much of an own opinion on stuff yet when it comes to language features. But I think that is a cleaner way to do it.

domenkozar:
I've checked and it doesn't fix https://github.com/AvengeMedia/DankMaterialShell/issues/2358, because it never reached the point of asking for credentials in the first place.

Mikilio:
@domenkozar That's unfortunate. It seems it was unrelated. The good news is that I already have a fix locally that I am going to submit once this gets merged. It also includes some mocked tests, but I don't actually have a eap network to connect to for a real integration test. I'd appreciate it if you could test that one.

bbedward:
802.11x I haven't been able to figure out the workflow since I dont have a way to run an 802.11x AP, but the GUI and things are mostly there once it does get figured out.

domenkozar:
> 802.11x I haven't been able to figure out the workflow since I dont have a way to run an 802.11x AP, but the GUI and things are mostly there once it does get figured out.

I'll give it a shot using claude code

PR #2407, opened 2026-05-12
refactor(ClipboardHistory): Sync toggle button between saved and recent tabs

hthienloc:
## Summary
- Change the Saved (pin) button to toggle between saved and recents tabs
- Remove the redundant History button for a cleaner UI
- Update tooltip text dynamically based on current tab state

## Problem
When viewing the clipboard history:
1. The Saved button always switched to saved tab but clicking again didn't return to recents
2. Two separate buttons (Saved + History) were redundant since the Saved button can toggle

## Solution
1. Changed Saved button to toggle between tabs:
```qml
// Before
onClicked: tabChanged("saved")

// After
onClicked: tabChanged(header.activeTab === "saved" ? "recents" : "saved")
```

2. Dynamic tooltip:
```qml
// Before
tooltipText: I18n.tr("Saved")

// After  
tooltipText: header.activeTab === "saved" ? I18n.tr("Recent") : I18n.tr("Saved")
```

3. Removed redundant History button

## Testing
- [x] Click Saved button -> shows pinned entries, tooltip shows "Recent"
- [x] Click Saved button again -> returns to recents tab, tooltip shows "Saved"

### DIFF
--- a/quickshell/Modals/Clipboard/ClipboardHeader.qml
+++ b/quickshell/Modals/Clipboard/ClipboardHeader.qml
@@ -49,8 +49,8 @@ Item {
             iconSize: Theme.iconSize - 4
             iconColor: header.activeTab === "saved" ? Theme.primary : Theme.surfaceText
             visible: header.pinnedCount > 0
-            tooltipText: I18n.tr("Saved")
-            onClicked: tabChanged("saved")
+            tooltipText: header.activeTab === "saved" ? I18n.tr("Recent") : I18n.tr("Saved")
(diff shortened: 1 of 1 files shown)


Purian23:
Hi there, this looks good. I wouldn't call it a (fix) as nothing was broken. I renamed the title a bit, thanks for the PR. Feel free to reach out before creating some of these if they may be questionable to our vision for DMS. I don't want anyone to waste their time. I may open up the Discussions board for this reason.

PR #2410, opened 2026-05-13
feat: add Display Profiles builtin control center widget

dvjn:
Adds a new built-in widget that lets users switch between configured display profiles directly from the Control Center, without opening Settings.

## Changes
- New `DisplayProfilesWidget.qml` builtin using `DisplayConfigState` and `SettingsData` singletons — no shell subprocess needed
- Detail panel shows all profiles as a button group for direct selection
- Clicking the pill in the bar cycles to the next profile
- Follows the existing loader/instance pattern of VPN, CUPS, and Tailscale widgets

## Usage
Add the **Display Profiles** widget in Settings → Control Center.

### DIFF
--- a/quickshell/Modules/ControlCenter/BuiltinPlugins/DisplayProfilesWidget.qml
+++ b/quickshell/Modules/ControlCenter/BuiltinPlugins/DisplayProfilesWidget.qml
@@ -1,5 +1,6 @@
 import QtQuick
 import qs.Common
+import qs.Services
 import qs.Widgets
 import qs.Modules.Plugins
 import qs.Modules.Settings.DisplayConfig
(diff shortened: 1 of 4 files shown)


Mikilio (commented):
When Auto profile switch is on. Some of the code needs to be treated differently. Auto-profiles have no name. The easiest would probably be to display something like "Auto".
Of course, you can find more creative ways to deal with it.

dvjn:
Hey Mikilio, thanks for the review.

Pushed changes with Auto mode support. Tried to keep the UX similar to what is there in the settings, ie, Auto mode disables manual profile selection.

Some screenshots for reference:
|    |   |
|---|---|
| Auto mode widget | Manual mode widget |
| Auto mode detail | Manual mode detail |

PR #2415, opened 2026-05-13
fix: prompt for 802.1X credentials on enterprise WiFi connect

Mikilio:
- Fixes enterprise WiFi connection to prompt for 802.1X credentials (password-flags set to AgentOwned) when Interactive: true, and set no password-flags otherwise
- Adds comprehensive test coverage for both interactive and non-interactive enterprise WiFi flows, verifying the correct 802-1x settings are applied
Note: The tests in this PR were written with AI assistance.

Fixes [#2414](https://github.com/AvengeMedia/DankMaterialShell/pull/2414)

### DIFF
--- a/core/internal/server/network/backend_networkmanager_wifi.go
+++ b/core/internal/server/network/backend_networkmanager_wifi.go
@@ -745,10 +745,14 @@ func (b *NetworkManagerBackend) createAndConnectWiFiOnDevice(req ConnectionReque
 				useSystemCACerts = *req.UseSystemCACerts
 			}
 
+			passwordFlags := uint32(0)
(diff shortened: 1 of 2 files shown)


Mikilio:
@domenkozar Could you test this?

Purian23:
The user already opened their own PR: https://github.com/AvengeMedia/DankMaterialShell/pull/2414 so let's figure out which works so we're not creating conflicts.

domenkozar:
This one doesn't fix it for me, while I can confirm that #2414 works

Purian23:
Thank you both! We'll get the only reviewed.

Issue #2290, opened 2026-04-27
Somewhat broken on hyprland with lua configs (workspace switcher, configs)


rose-the-dev:
### Compositor

Hyprland

### Distribution

NixOS

### Select your Installation Method

Distro Packaging

### Was this your original Installation method?

Yes

### dms doctor -vC

<details>
<summary>Click to expand</summary>
## DMS Doctor Report

**System**
- [ok] Operating System: NixOS 26.05 (Yarara)
  - Supported for runtime (install via NixOS module or Flake)
- [ok] Architecture: amd64
- [ok] Display Server: Wayland
  - WAYLAND_DISPLAY=wayland-1

**Versions**
- [ok] DMS CLI: dms v1.4.4.1
  - /nix/store/4fbp9rxkfil7l7zrfcd23iv0rjyn5sd9-dms-shell-1.4.4.1/bin/.dms-wrapped
- [warn] Quickshell: quickshell 0.2.1, revision tag-v0.2.1, distributed by: Nixpkgs
  - /run/current-system/sw/bin/qs
- [ok] DMS Shell: v1.4.4
  - /nix/store/4fbp9rxkfil7l7zrfcd23iv0rjyn5sd9-dms-shell-1.4.4.1/share/quickshell/dms

**Installation**
- [ok] DMS Configuration: Found
  - /nix/store/4fbp9rxkfil7l7zrfcd23iv0rjyn5sd9-dms-shell-1.4.4.1/share/quickshell/dms
- [ok] shell.qml: Present
  - /nix/store/4fbp9rxkfil7l7zrfcd23iv0rjyn5sd9-dms-shell-1.4.4.1/share/quickshell/dms/shell.qml
- [info] Install Type: Nix store
  - /nix/store/4fbp9rxkfil7l7zrfcd23iv0rjyn5sd9-dms-shell-1.4.4.1/share/quickshell/dms

**Compositor**
- [ok] Hyprland: 0.54.0 (git)
  - /run/current-system/sw/bin/hyprland
- [info] Active: Hyprland

**Quickshell Features**
- [info] Polkit: Not available
  - Authentication prompts
- [info] IdleMonitor: Not available
  - Idle detection
- [info] IdleInhibitor: Not available
  - Prevent idle/sleep
- [info] ShortcutInhibitor: Not available
  - Allow shortcut management (niri)

**Optional Features**
- [ok] accountsservice: Available
  - User accounts
- [ok] power-profiles-daemon: Available
[...]

PR #2419, opened 2026-05-14
Support Hyprland lua dispatching

borisbabic:
This consolidates hyprland dispatching, and adds support for the new hyprland lua dispatching while keeping support for the old style.

It solves the workspace switching related issues in #2290


borisbabic:
Those using the flake on nixos can get this right away by changing the flake url:

```
      url = "github:AvengeMedia/DankMaterialShell/pull/2419/head";
```

PR #2428, opened 2026-05-15
feat(power): lower refresh rate on battery

### LINKED ISSUE #1203: Refresh rate change on power source
## Feature Description

An option in the power settings to pick a refresh rate based on battery or AC power. 

## Use Case

When on battery a laptop could switch to 60hz to save power and when on AC power it could switch to its HFR mode automatically. 

## Compositor

- [ X] All compositors
- [ ] niri
- [ ] Hyprland
- [ ] dwl (MangoWC)
- [ ] sway

## Proposed Solution
 
Implement a drop-down menu in the section that lets you choose power profile based on battery or AC

## Alternatives/Existing Solutions

If it can't be done natively, maybe a system like KDE has where it can run a script on power source change. In my case using KDE I have one to switch to 60hz one to switch to 165hz. It detects the change and automatically runs a script of your choosing.

atimofeev:
Add a Power & Sleep setting that lowers eligible displays to 60 Hz on battery and restores their previous mode on AC power.

Skip displays without an available 60 Hz mode, displays with a single refresh-rate option, current modes at or below 60 Hz, and VRR-enabled outputs. 

Apply niri changes at runtime through output commands and use wlr-output-management as fallback.

Closes #1203


atimofeev:
Need help with additional testing, as I only have a single device with multiple refresh rates display (240/60).

Also, changes to DisplayService.qml might be overengineered.

atimofeev:
Here's the first issue:

- When you restart dms being on battery, it thinks that configured refresh rate is 60hz and it never goes back to the preferred 240hz

Probably can be resolved with this logic:

- Get active monitor profile refresh rate and use it as baseline
- If monitor profiles are not configured, fall back to monitor preferred refresh rate
- Re-check these steps on profile or power source change

atimofeev:
Implemented stateless transitions between 60hz and higher refresh rates.
Also improved logging to be more human-readable.

atimofeev:
update after latest changes: functionality looks better, but I’m still hitting one nasty crash.

I can reproduce with this flow:                                                                                                                                               
plug AC -> connect external display over type-c dp -> quickshell crashes

dumps end in same path:
Qt::endPropertyUpdateGroup -> WindowsetManager::onServerCommit -> WorkspaceManager::ext_workspace_manager_v1_done (SIGSEGV)

so this looks more like quickshell workspace/output commit race on hotplug/reconfigure.
refresh-rate logic still seems to be trigger surface (mode apply happens right before crash), but probably not root cause by itself.

atimofeev:
Resolved.

Added minimal 300ms debounce around DisplayService refresh-rate sync, and BatteryService now routes startup/power/profile/output refresh triggers through requestSync(...). This coalesces AC + external-display event bursts and avoids repeated mode applies that were triggering quickshell crash.

Spoch-dev:
Closing

atimofeev:
Although I'm no longer daily-driving Hyprland, this feature seems to work there after a quick test.

quickshell/Services/DisplayService.qml
```diff
@@ -56,6 +77,373 @@ Singleton {
     property int gammaLowTemp: gammaState?.config?.LowTemp ?? 0
     property int gammaHighTemp: gammaState?.config?.HighTemp ?? 0
 
+    function syncRefreshRates(isPluggedIn, reason) {
+        if (isPluggedIn) {
```
bbedward: We should guard this entire function so it doesnt run unconditionally

```js
if (!SettingsData.lowerDisplayRefreshRateOnBattery) {
   return
}
if (!isPluggedIn) {
  applyBatteryTargets...
  return
}
applyACTargets...
````

File: quickshell/Services/DisplayService.qml [hunk 2]
```diff
@@ -56,6 +77,373 @@ Singleton {
...
+        }
+
+        if (!bestMode || bestDiff > batteryRefreshRateTolerance)
+            return null;
+
+        return bestMode;
+    }
+
+    function computeTargetMode(outputName, output, currentMode, backend) {
+        const profileMode = findActiveProfileMode(outputName, output);
```
bbedward, line 489: A lot of this machinery might just be a lot easier/deletable if we persisted the previousMode , instead of needing to try to figure out what the previous mode was across shell restarts.

quickshell/Services/DisplayService.qml
```diff
@@ -56,6 +77,373 @@ Singleton {
...
+        const modes = output.modes || [];
+        const sameResolutionModes = modes.filter(m => getModeWidth(m) === getModeWidth(currentMode) && getModeHeight(m) === getModeHeight(currentMode));
+
+        let preferred = null;
+        let highest = null;
+        let highestRefresh = 0;
+
+        for (const mode of sameResolutionModes) {
+            const refresh = getModeRefresh(mode);
+            if (mode.preferred === true)
```
bbedward: This wont catch on niri probably, since its key is `is_preferred`, but if we just persist the previous mode its probably a moot point.

quickshell/Services/BatteryService.qml
```diff
@@ -213,9 +216,40 @@ Singleton {
...
+    }
+
+    Connections {
+        target: NiriService
+        function onOutputsChanged() {
+            DisplayService.requestSync("output-change");
+        }
+    }
+
+    Connections {
```
bbedward: The listeners should be in DisplayService, not BatteryService

quickshell/Common/SettingsData.qml
```diff
@@ -963,6 +964,7 @@ Singleton {
     property var hyprlandOutputSettings: ({})
     property var displayProfiles: ({})
     property var activeDisplayProfile: ({})
+    property var activeDisplayProfileModes: ({})
```
bbedward, line 967: Missing from SettingsSpec

quickshell/Services/DisplayService.qml
```diff
@@ -56,6 +77,373 @@ Singleton {
...
+        syncDebounce.restart();
+    }
+
+    function _runSync() {
+        syncRefreshRates(BatteryService.isPluggedIn, _pendingReason || "sync");
+        _pendingReason = "";
+    }
+
+    function applyConfiguredTargets(context, reason) {
+        if (CompositorService.isNiri) {
```
bbedward, line 210: It might make sense to have different paths for Hyprland/Mango since we manage them with c onfigurations too and not just with wlr-output

quickshell/Modules/Settings/DisplayConfig/DisplayConfigState.qml
```diff
@@ -2242,6 +2275,15 @@ Singleton {
             outputs: outputConfigs
         };
 
+        if (profileId) {
```
bbedward: I think this should be moved into the callback, after the match check. the `readMonitorsJson` callback

quickshell/Modules/Settings/PowerSleepTab.qml
```diff
@@ -187,6 +187,16 @@ Item {
                     }
                 }
 
+                SettingsToggleRow {
+                    settingKey: "lowerDisplayRefreshRateOnBattery"
+                    tags: ["power", "battery", "display", "refresh", "rate", "60hz", "hz"]
+                    text: I18n.tr("Lower display refresh rate on battery")
```
bbedward: Add translator context, second argument. `I18n.tr("Lower display refresh rate on battery", "setting title under power & sleep tab")`

bbedward (changes requested):

PR #2461, opened 2026-05-21
feat(notifications): user-configurable font size for notification summary and body

klesh:
swappy-20260521_173201

Adds settings and UI to allow users to pick font size for notification summary and htmlBody. Updates NotificationPopup to use these settings.


quickshell/Modules/Notifications/Popup/NotificationPopup.qml
```diff
@@ -895,10 +895,11 @@ PanelWindow {
                         }
                     }
 
+
                     StyledText {
                         text: notificationData ? (notificationData.summary || "") : ""
                         color: Theme.surfaceText
-                        font.pixelSize: Theme.fontSizeMedium
+                        font.pixelSize: SettingsData.notificationSummaryFontSize
```
bbedward: I think these should default to unset, and fallback to the constants Theme.fontSizeMedium. Instead of being separate value defaults, would make it easier to change defaults in the future.

bbedward (changes requested):

quickshell/Modules/Notifications/Popup/NotificationPopup.qml
klesh: @bbedward Thanks for taking time to review the PR. I just updated the code and added the Unset option. Please kindly take another look when you find time. Thx~

Purian23:
Thanks, Klesh!

PR #2464, opened 2026-05-21
feat(Clipboard-Bar-Hist): Add search/filter to saved clipboard entries & animation states

Bodify:
DankBar: Enable searching in saved clipboard entries and changes the popup title between "Clipboard History" and "Clipboard Saved" when toggling the Saved/Recent button.

Also fixes search/filter issues for DankBar's Clipboard History similar to https://github.com/AvengeMedia/DankMaterialShell/pull/2315

### DIFF
--- a/quickshell/Common/Theme.qml
+++ b/quickshell/Common/Theme.qml
@@ -970,6 +970,7 @@ Singleton {
 
     readonly property int shorterDuration: (typeof SettingsData !== "undefined" && SettingsData.animationSpeed === SettingsData.AnimationSpeed.Custom) ? SettingsData.customAnimationDuration : currentDurations.shorter
     readonly property int shortDuration: (typeof SettingsData !== "undefined" && SettingsData.animationSpeed === SettingsData.AnimationSpeed.Custom) ? SettingsData.customAnimationDuration : currentDurations.short
+    readonly property bool snapListModelChanges: shortDuration <= 0
     readonly property int mediumDuration: (typeof SettingsData !== "undefined" && SettingsData.animationSpeed === SettingsData.AnimationSpeed.Custom) ? SettingsData.customAnimationDuration : currentDurations.medium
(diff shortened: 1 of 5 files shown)


purian23 (changes requested):
Thanks for the PR! I tested search on saved/pinned entries and it works great. The ClipboardService filter fix and tab-specific header count look good to me.

Couple of small items:

Please revert the manual `en.json` changes. That file is auto-generated on its own. 

Let's drop the add/remove/displaced/move: null changes, they were intentional. I actually like the list animating while filtering, it feels nicer than an instant snap.

Happy to approve once those are addressed, we appreciate it.

Bodify:
> Please revert the manual en.json changes. That file is auto-generated on its own.

Done, my bad.

> Let's drop the add/remove/displaced/move: null changes, they were intentional. I actually like the list animating while filtering, it feels nicer than an instant snap.

I wouldn't mind some short animations but I can't even say I notice them the way it is now? I only notice the bugs. I'm not entirely sure what the root cause behind it is but keeping the transitions messes everything up same as in https://github.com/AvengeMedia/DankMaterialShell/pull/2315. Removing them fixes it. It will probably require a fix elsewhere that is out of my comfort zone and revert the mentioned PR after a proper fix is in place.

This is what happens for me if I don't null the animations:

https://github.com/user-attachments/assets/f1821989-af05-4d3c-b2a2-292c511f3102

Bodify:
I also made the Pinned/Saved icon stay highlighted when selected to make it more consistent and obvious.

Let me know if I should remove the animation nulls or not or what happens there.

Bodify:
After digging a bit more I found the reason: I have Animation Duration set to None/0ms in the Typography & Motion settings. It doesn't like that.

I'm not sure of the best way fix to this? I removed the nulling of the animations but it will require a fix for people with None/0ms animations set. That fix is then likely also wanted for this previous change in https://github.com/AvengeMedia/DankMaterialShell/pull/2315

Purian23:
Nice work on this one @Bodify! When you first presented what appeared to be a bug, I was baffled, but with animations turned off it made sense. I pushed an update to allow both states. When animations are off, they will now use the null settings, when they are above zero they will use the animations we put in place. Thanks again!

purian23 (approved):

PR #2465, opened 2026-05-21
fix(DankBar): RamMonitor, CpuMonitor, DiskUsage modules fix paddings

Bogdan-Torkhov:


Bogdan-Torkhov:
https://github.com/AvengeMedia/DankMaterialShell/issues/2466 fixing these issue

Bogdan-Torkhov:
Also fixed ram and cpu module paddings

Bogdan-Torkhov (commented):
I think all okay

Bogdan-Torkhov (commented):
I think it's all okay

bbedward:
The rationale of minimum widths is do they bar widgets don't shift around as they change. I admit maybe disk usage shouldn't have that constraint, but for the more frequent changing values like CPU and Ram it's pretty jarring to have adjacent widgets shift around so much.

Bogdan-Torkhov:
> The rationale of minimum widths is do they bar widgets don't shift around as they change. I admit maybe disk usage shouldn't have that constraint, but for the more frequent changing values like CPU and Ram it's pretty jarring to have adjacent widgets shift around so much.

I am using these versions of modules and they change size dynamically, even disk usage, and works very well

Purian23:
I think you're missing the point a bit. You're attempting to "fix" something that was not technically broken. We make it to where padding could be dynamic or static as is. You ran into the static variant and pushed updates. All you had to do was adjust the setting within widgets on your own config, (which is now broken with your edits). 

image

Bogdan-Torkhov:
> I think you're missing the point a bit. You're attempting to "fix" something that was not technically broken. We make it to where padding could be dynamic or static as is. You ran into the static variant and pushed updates. All you had to do was adjust the setting within widgets on your own config, (which is now broken with your edits). 
> 
> image
> 

I will test again, but how I remember disk usage don't have static or dynamic width. Trying to "fix" cpu and ram widgets maybe really was the error, I will recheck all and maybe write dynamic width button for disk usage settings

PR #2470, opened 2026-05-22
feat: unify media controls dropdown interactions, hover behavior and cycle controls

hthienloc:
### What

* **Hover-to-show**: Adds hover-to-show and hover-to-hide behavior to the Media Players and Audio Output Devices dropdowns, matching the Volume dropdown behavior.
* **Click-to-cycle**: Clicking the Media Players or Audio Output Devices button while expanded now cycles to the next available media player or audio output device, instead of closing the dropdown.
* **Dynamic Bindings**: Replaced static argument passing with dynamic `Qt.binding()` to MPRIS controller singletons, fixing a display bug where player lists became stale/stalled upon switching.
* **Artwork Cache Reset**: Added a listener to reset the `lastValidArtUrl` buffer in `DankAlbumArt` when the active player changes, resolving a bug where a previous player's artwork would incorrectly persist on players that have no artwork.
* **Visual Consistency**: Set the Audio Output Devices button icon to always display `"speaker"` (rather than toggling to `"expand_less"` when open), ensuring a uniform design across all media control buttons.
* **Tooltip Cleanup**: Removed redundant helper tooltips ("Output Device" and "Media Players") to declutter the UI.
* **Audio Metadata Decoupling (New)**: Fixed a binding typo in `MediaPlayerTab` where the artist label incorrectly pointed to `trackTitle`, causing the song title and artist name to be duplicated when displaying media info.
* **Minimalist Player Subtitle (New)**: Cleaned up the player selection subtitle to display purely the artist name if available, or `"Unknown Artist"` if absent, aligning with other standard media info displays.
* **Audio Output Volume Indicators (New)**: Replaced the redundant "Active / Available" text in the Audio Output Devices list with the device's actual volume percentage (e.g. `80%`) or `Muted` status, providing highly useful real-time information.
* **Keyboard Shortcuts (New)**: Added a robust set of intuitive keyboard controls to the Media popout when active:
  - `0` - `9`: Seek to the corresponding percentage of the track (`0%` - `90%`).
  - `Left` / `Right` Arrow keys: Seek backward / forward by 5 seconds.
  - `Up` / `Down` Arrow keys: Increase / decrease player volume by 5% and **automatically open the volume dropdown overlay** (auto-hides after inactivity) to give real-time visual feedback.
  - `Spacebar`: Toggle play/pause playback.
  - `M` key: Toggle volume mute/unmute and automatically trigger the volume dropdown overlay.
* **Browser Seek Reset Fix (New)**: Clamped the minimum seek position (via keyboard/drag actions) to `0.1s` rather than exactly `0s`. This prevents the MPRIS bridge of browser-based players (like YouTube on Firefox/Chrome) from thinking the stream is uninitialized or needs a full reload, which temporarily reset the track duration metadata to `1s`.

### Why

Unifies the interaction model and improves the user experience by resolving inconsistencies in hover/click behaviors, dynamic list rendering, and icon states in the Control Center media widgets. Additionally, fixes track metadata duplication, provides useful device volume levels, and adds keyboard shortcuts for convenient media popout controls.

### How tested

Verified locally in the QuickShell environment by hovering, clicking, and cycling through active media players and audio devices, confirming smooth transitions, correct artwork clearing, and real-time list updates. Also tested keyboard shortcuts (0-9 seeking, arrow keys, spacebar) and volume percentage displays across multiple simulated players and audio sinks.

https://github.com/user-attachments/assets/1fb8de7e-3366-4816-92fd-76b3487bc919

### DIFF
--- a/quickshell/Modules/DankDash/DankDashPopout.qml
+++ b/quickshell/Modules/DankDash/DankDashPopout.qml
@@ -230,6 +230,13 @@ DankPopout {
                     return;
                 }
 
+                if (root.currentTabIndex === 1 && mediaLoader.item?.handleKeyEvent) {
+                    if (mediaLoader.item.handleKeyEvent(event)) {
+                        event.accepted = true;
+                        return;
+                    }
+                }
+
                 if (root.currentTabIndex === 2 && wallpaperLoader.item?.handleKeyEvent) {
                     if (wallpaperLoader.item.handleKeyEvent(event)) {
                         event.accepted = true;
--- a/quickshell/Modules/DankDash/DankDashPopout.qml
+++ b/quickshell/Modules/DankDash/DankDashPopout.qml
@@ -402,6 +402,7 @@ DankPopout {
                                 }
                                 onHideDropdowns: root.__hideDropdowns()
                                 onDropdownButtonExited: root.__startCloseTimer()
+                                onDropdownButtonEntered: root.__stopCloseTimer()
                             }
                         }
                     }
--- a/quickshell/Modules/DankDash/DankDashPopout.qml
+++ b/quickshell/Modules/DankDash/DankDashPopout.qml
@@ -25,14 +25,14 @@ DankPopout {
     property int __dropdownType: 0
     property point __dropdownAnchor: Qt.point(0, 0)
     property bool __dropdownRightEdge: false
-    property var __dropdownPlayer: null
-    property var __dropdownPlayers: []
+    property var __dropdownPlayer: MprisController.activePlayer
+    property var __dropdownPlayers: MprisController.availablePlayers
 
     function __showVolumeDropdown(pos, rightEdge, player, players) {
         __dropdownAnchor = pos;
         __dropdownRightEdge = rightEdge;
-        __dropdownPlayer = player;
-        __dropdownPlayers = players;
+        __dropdownPlayer = Qt.binding(() => MprisController.activePlayer);
+        __dropdownPlayers = Qt.binding(() => MprisController.availablePlayers);
         __dropdownType = 1;
     }
 
@@ -45,8 +45,8 @@ DankPopout {
     function __showPlayersDropdown(pos, rightEdge, player, players) {
         __dropdownAnchor = pos;
         __dropdownRightEdge = rightEdge;
-        __dropdownPlayer = player;
-        __dropdownPlayers = players;
+        __dropdownPlayer = Qt.binding(() => MprisController.activePlayer);
+        __dropdownPlayers = Qt.binding(() => MprisController.availablePlayers);
         __dropdownType = 3;
     }
 
@@ -69,7 +69,7 @@ DankPopout {
         id: __volumeCloseTimer
         interval: 400
         onTriggered: {
-            if (__dropdownType === 1) {
+            if (__dropdownType !== 0) {
                 __hideDropdowns();
             }
         }
@@ -394,7 +394,7 @@ DankPopout {
                                     root.__showPlayersDropdown(pos, rightEdge, player, players);
                                 }
                                 onHideDropdowns: root.__hideDropdowns()
-                                onVolumeButtonExited: root.__startCloseTimer()
+                                onDropdownButtonExited: root.__startCloseTimer()
                             }
                         }
                     }
--- a/quickshell/Modules/DankDash/MediaDropdownOverlay.qml
+++ b/quickshell/Modules/DankDash/MediaDropdownOverlay.qml
@@ -357,7 +357,13 @@ Item {
                                     }
 
                                     StyledText {
-                                        text: modelData === AudioService.sink ? "Active" : "Available"
+                                        text: {
+                                            if (!modelData?.audio)
+                                                return modelData === AudioService.sink ? I18n.tr("Active") : I18n.tr("Available");
+                                            if (modelData.audio.muted)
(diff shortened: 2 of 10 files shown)


Purian23 (changes requested):
Thanks for the PR, this is nicely done with some real imporvements!

A few things we can cleanup:

- `TrackArtService` now downloads every remote artwork URL into `Paths.imagecache` before display. That’s reasonable for YouTube fallback/caching, but the current implementation writes `curl` output directly to the final cache path and later accepts the cache with only `test -f`. If a download times out or gets interrupted, we can keep a partial file forever. Let's download to a temp path and rename on success, or delete failed/partial files.

- The `Proc.runCommand` IDs use the artwork hash, which creates a persistent debouncer entry for every unique remote art URL. Since these are one-shot checks/downloads, please use random/null IDs or otherwise clean them up after completion.

- While a new remote artwork URL is loading, `resolvedArtUrl` can still point at the previous track’s artwork. That can briefly show stale art in the media tab/overview/OSD. Please clear it or fall back to the raw URL while the new art resolves.

- `MediaDropdownOverlay.__panelHoverCount` should be reset when dropdowns are hidden. The overlay stays alive while the dash is open, so closing while hovered can leave the counter positive and make future hover-close behavior stick.

Non-blocking cleanup notes:

- `activePlayerArtUrl` looks unused.
- `MediaOverviewCard` still uses raw `activePlayer.length`; consider using `MprisController.activePlayerStableLength` there too for consistency.
- The `DankCircularImage` `sourceSize` removal could use a short rationale, or revert if it was accidental.

Once those cache/stale-hover issues are addressed, I think this is in good shape.

hthienloc:
@Purian23 Thanks for the detailed review and suggestions. I have pushed a commit (https://github.com/AvengeMedia/DankMaterialShell/pull/2470/commits/bb6527b344890c9204c756ad5cc9c7e4f8b1a4a8) addressing all the issues you raised.

I would also love to get your thoughts on a few open points:

- **Stale/Undefined Durations:** For some web-based media players, the total duration (length) occasionally reports as  undefined  initially, or requires closing and reopening the popout to update properly.
-  **Hover Interaction Loop:** If a user hovers over one of the three control buttons to open its dropdown, moves the cursor away, and then immediately hovers back onto that exact same icon, what should be the desired behavior? Currently, it toggles (closes) on re-hover, but I am wondering if keeping it open might be a more intuitive user experience.
- **Shortcut Guidance:** I noticed there is currently no shortcut documentation for this PR. Similar to the Clipboard Manager, which has an "info" button that displays a list of shortcuts, I think adding a similar keyboard shortcut helper/info button would be highly beneficialfor discoverability.
- **Non-Uniform Panel Widths:** The dropdowns for Media Players (240px) and Output Devices (280px) currently have non-uniform widths. I am not sure if this discrepancy was intentional in the original design, but I think the staggered/asymmetrical layout actually looks quite polished.

Thanks again for the valuable feedback. Looking forward to your thoughts.

Purian23:
Thanks!

One small implementation note; the new `sh -c` curl commands are built by string-concatenating paths/URLs inside single quotes. Since artwork URLs come from player metadata, it would be safer to avoid shell interpolation there. Either pass values as shell args (`sh -c '...' sh "$tmpPath" "$url" "$filePath"`) or use direct command arrays plus a separate `mv` on success. Also worth ensuring `mkdir -p` happens in the same command path before writing the temp file, so there is no async race with `Paths.mkdir()`.

On the open questions:
- **Stale/undefined durations:** I’d treat browser MPRIS duration as best-effort. The UI should react when `activePlayer.length` eventually becomes valid through `activePlayerStableLength`; until then, showing `--:--` and disabling seek controls is fine. If it persists after, that likely means we have a missing signal/binding update somewhere, and we can track that as a follow-up bug.

- **Hover interaction loop:** I agree and think re-hovering the same icon should keep the dropdown open, not toggle it closed. Hover should be idempotent: enter opens/keeps open, leaving button + panel starts the close timer. Click while already expanded can keep the new cycle behavior.

- **Shortcut guidance:** Agreed, an info/helper button like Clipboard Manager may be useful. It can be implemented later in a follow-up unless you want to include a small version here. I think most of the controls are fairly universal so this isn't a deal breaker. 

- **Panel widths:** We're good with keeping the widths non-uniform and the staggered looks okay to me.

hthienloc:
1. **Secure Download Commands:** I've re-architected the `sh -c` download command in `TrackArtService.qml` to pass the cache paths and metadata-sourced URLs strictly as shell arguments (`$1`, `$2`, `$3`), avoiding any shell quoting or injection vulnerabilities entirely. I also moved the `mkdir -p` command directly into the same execution path to prevent any async directory races.
2. **Hover Idempotency:** Added a new `dropdownButtonEntered` signal in `MediaPlayerTab.qml` to stop the overlay close timer when any button is re-hovered. Hover is now fully idempotent and re-entering the buttons correctly keeps the active overlay open.
3. **Shortcut Guidance:** I feel adding an info button for keyboard shortcut guidance might be redundant here, as these media control hotkeys (like Space to play/pause, Left/Right to seek) are quite universal and standard. I think keeping the UI minimal and clutter-free is the better approach for now.

Let me know if there's anything else we can improve before merging.

Purian23:
Thanks for the nice work and **QOL** update, @hthienloc!

Purian23 (approved):

PR #2475, opened 2026-05-23
Add native Triad IPC support

greenm01:
This adds native Triad IPC support to DMS.

Triad is a programmable Wayland window manager with per-workspace layouts, a native IPC, and a Niri-compatible IPC facade. The repo is here: https://github.com/greenm01/triad

DMS already works on Triad through the Niri-compatible path. This adds a native backend for sessions that expose TRIAD_SOCKET, so DMS can read Triad’s workspace, output, window, overview, and keyboard-layout state directly.

The Niri service is left alone. Triad detection only kicks in when TRIAD_SOCKET is set, or when the desktop session identifies itself as Triad, so existing Niri behavior should stay as-is.

Tested in a live Triad session with niri-compat disabled for the Dank profile. make lint-qml passes.

### DIFF
--- a/quickshell/DMSShellIPC.qml
+++ b/quickshell/DMSShellIPC.qml
@@ -229,6 +229,9 @@ Item {
             if (CompositorService.isNiri && NiriService.currentOutput) {
                 return NiriService.currentOutput;
             }
+            if (CompositorService.isTriad && TriadService.currentOutput) {
+                return TriadService.currentOutput;
+            }
             if ((CompositorService.isSway || CompositorService.isScroll || CompositorService.isMiracle) && I3.workspaces?.values) {
                 const focusedWs = I3.workspaces.values.find(ws => ws.focused === true);
                 return focusedWs?.monitor?.name || "";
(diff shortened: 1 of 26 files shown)


greenm01:
Quick note: I’m also talking with Quickshell about a native Triad module.

The current feedback there is to wait a bit. The Triad code is mostly standalone, but it still touches `ObjectModel`, and Quickshell is about to get changes that should let this live first as a separate QML module.

So this PR can still be useful on its own. If the Quickshell module path lands, I’m happy to come back and either wire DMS to that module or trim this down so we do not keep two IPC clients around.

bbedward:
> Quick note: I’m also talking with Quickshell about a native Triad module.
> 
> The current feedback there is to wait a bit. The Triad code is mostly standalone, but it still touches `ObjectModel`, and Quickshell is about to get changes that should let this live first as a separate QML module.
> 
> So this PR can still be useful on its own. If the Quickshell module path lands, I’m happy to come back and either wire DMS to that module or trim this down so we do not keep two IPC clients around.

afaik outfoxxed is preferring the path of supporting standard wayland protocols (ext-workspace, toplevels) for compositors rather than specific compositor modules these days. So not sure if that would land or not.

But this looks fine if you can resolve the conflicts

greenm01:
Thanks, I believe ext-workspace is the right way to go ultimately.

PR #2507, opened 2026-05-27
feat(control-center): add DiskUsage widget config overlay with showMountPath toggle and standardized tile sizing

gbPagano:
## Summary

Adds a configuration overlay to the DiskUsage widget in the Control Center, allowing users to toggle the mount path visibility. Also standardizes small tile icon and text sizing using Theme constants for consistency and theme responsiveness, and unifies font weight to Bold across both tile widgets.

### Widget Config Overlay
- **New component**: `DiskUsageWidgetConfigMenu.qml` context menu with a "Show mount path" toggle for the DiskUsage widget
- **New component**: `WidgetConfigOverlay.qml` generic overlay container that positions the config menu relative to the anchor button and persists setting changes via `SettingsData`
- **New signal**: `configRequested(int index, var widgetData, var anchor)` propagated through `DragDropWidgetWrapper` → `DragDropGrid` → `ControlCenterPopout` to open the overlay
- **Config button**: adds a gear icon button next to the remove button in edit mode, visible only for widgets that have a config menu (`diskUsage`)

### showMountPath Property
- Adds `showMountPath` boolean property to `DiskUsagePill`, `SmallDiskUsageButton`, and bar `DiskUsage` widgets
- When `showMountPath` is `false`, the mount path text is hidden; `DiskUsagePill` falls back to "Disk Usage" as primary text
- Default value is `true` for backward compatibility
- Property is persisted in settings via `WidgetsTabSection` serialization keys and `WidgetsTab` defaults

### Standardized Tile Sizing (Small Buttons)
- `SmallDiskUsageButton` and `SmallBatteryButton`: icon size set to `Theme.iconSizeLarge`, text font size set to `Theme.fontSizeLarge`
- Both widgets now use `Font.Bold` for consistent font weight
- Replaces proportional sizing (`width * factor`) with Theme constants for consistent, theme-responsive rendering

image
image

image

### DIFF
--- a/quickshell/Modules/ControlCenter/Components/DiskUsageWidgetConfigMenu.qml
+++ b/quickshell/Modules/ControlCenter/Components/DiskUsageWidgetConfigMenu.qml
@@ -1,0 +1,50 @@
+import QtQuick
+import qs.Common
+import qs.Widgets
+
+Rectangle {
+    id: root
+
+    property var widgetData: null
+
+    signal showMountPathChanged(bool show)
+
+    width: 260
+    height: menuColumn.implicitHeight + Theme.spacingS * 2
+    radius: Theme.cornerRadius
+    color: Theme.surfaceContainer
+    border.color: Qt.rgba(Theme.outline.r, Theme.outline.g, Theme.outline.b, 0.16)
+    border.width: 1
+
+    MouseArea {
+        anchors.fill: parent
+    }
+
+    Column {
+        id: menuColumn
+        anchors.fill: parent
+        anchors.margins: Theme.spacingS
+        spacing: 2
+
+        StyledText {
+            width: parent.width
+            leftPadding: Theme.spacingS
+            topPadding: Theme.spacingXS
+            bottomPadding: Theme.spacingXS
+            text: I18n.tr("Disk Usage Widget")
+            font.pixelSize: Theme.fontSizeSmall
+            font.weight: Font.Medium
+            color: Theme.surfaceText
+        }
+
+        DankToggle {
+            width: parent.width
+            text: I18n.tr("Show mount path")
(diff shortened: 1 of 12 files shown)


quickshell/Modules/ControlCenter/Components/DiskUsageWidgetConfigMenu.qml
```diff
@@ -0,0 +1,50 @@
...
+            text: I18n.tr("Disk Usage Widget")
+            font.pixelSize: Theme.fontSizeSmall
+            font.weight: Font.Medium
+            color: Theme.surfaceText
+        }
+
+        DankToggle {
+            width: parent.width
+            text: I18n.tr("Show mount path")
+            description: I18n.tr("Display the mount path under the widget")
```
bbedward: I would get rid of the description, too many translations to describe this behavior. Should be able to convey it with 1, especially if adding context. I18n.tr("Show mount path", "toggle in control center disk usage widget to turn mount path display on or off")

quickshell/Modules/ControlCenter/Components/DiskUsageWidgetConfigMenu.qml
```diff
@@ -0,0 +1,50 @@
...
+        anchors.fill: parent
+        anchors.margins: Theme.spacingS
+        spacing: 2
+
+        StyledText {
+            width: parent.width
+            leftPadding: Theme.spacingS
+            topPadding: Theme.spacingXS
+            bottomPadding: Theme.spacingXS
+            text: I18n.tr("Disk Usage Widget")
```
bbedward: Same think we can get rid of this translation all together

bbedward (changes requested):

gbPagano:
Done! Removed the title and the toggle description

image

PR #2509, opened 2026-05-28
Hide MPRIS players that have nothing to play

cwelsys:
e.g. mpd with an empty queue


Purian23:
Thanks @cwelsys, added a small follow up commit to add the Artist as well. Appreciate the PR!

PR #2513, opened 2026-05-28
Add DoNotDisturb & IdleInhibitor Indicators for the ControlCenterButton widget + Add optional icon to KeyboardLayoutName widget

euletheia:
## Why

These icons do not serve the same purpose as their widget counter-parts : they are merely indicators, either reflecting a particular state of the desktop or completely absent.

They help unclutter the bar and adapt nicely to a keyboard-centric workflow, while serving as a reminder that their related buttons can be accessed (if configured) via the control center.

ControlCenterIndicatorsSettings

ControlCenterIndicator

---

### Adjustments

- Maybe the keyboard icon wasn't missing, but it felt like it to me (just having big letters without context made me shiver).
KbdLayoutIcon

KbdLayoutIconOLD

- The option to toggle inline expansion for system tray was not completely readable so I adjusted the width of the popup. 
SystemTraySettingsOLD
SystemTraySettings

### DIFF
--- a/quickshell/Common/SettingsData.qml
+++ b/quickshell/Common/SettingsData.qml
@@ -316,6 +316,7 @@ Singleton {
     property bool controlCenterShowPrinterIcon: false
     property bool controlCenterShowScreenSharingIcon: true
     property bool controlCenterShowIdleInhibitorIcon: false
+    property bool controlCenterShowDoNotDisturbIcon: false
     property bool showPrivacyButton: true
     property bool privacyShowMicIcon: false
     property bool privacyShowCameraIcon: false
--- a/quickshell/Common/SettingsData.qml
+++ b/quickshell/Common/SettingsData.qml
@@ -315,6 +315,7 @@ Singleton {
     property bool controlCenterShowBatteryIcon: false
     property bool controlCenterShowPrinterIcon: false
     property bool controlCenterShowScreenSharingIcon: true
+    property bool controlCenterShowIdleInhibitorIcon: false
     property bool showPrivacyButton: true
     property bool privacyShowMicIcon: false
     property bool privacyShowCameraIcon: false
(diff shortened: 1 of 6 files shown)


purian23 (changes requested):
Thanks for the PR!

I see you added a keyboard icon to the horizontal Keyboard Layout widget. The issue is it's shown unconditionally with no user setting to disable it.  We either need to add a persistent toggle option similar to compact mode or remove it for now. 

Additional notes:
Register `controlCenterShowIdleInhibitorIcon` and `controlCenterShowDoNotDisturbIcon` in `SettingsSpec.js`, otherwise they won't persist. 

Add `showDoNotDisturbIcon` to the keys array in `WidgetsTabSection.qml` serialize path (it's present in `WidgetsTab.qml` but missing from the section copy).

The CC indicator feature itself looks good once persistence/serialization is fixed.

euletheia:
I've added an options context menu for the Keyboard Layout widget and a switch to toggle the icon per widget instance.

Purian23:
@euletheia Thanks for addressing the previous feedback. The CC indicators and the keyboard icon toggle look good. Please align the CC indicator defaults between `SettingsData.qml` and `SettingsSpec.js` (both should be false).

In the future, let's not add in additional non-scoped commits after a review has already been done.
The fix(PrivacyIndicator) commit doesn't belong in this PR. More importantly, the handler change looks incomplete: privacy settings are global (`SettingsData` + `PrivacyIndicator`), but the commit writes per-widget `privacyShowMicIcon` fields that nothing reads. PrivacyIndicator doesn't use `widgetData`, the context menu toggles still bind to `SettingsData`, and the new fields aren't in cloneWidgetData's keys so they won't persist reliably.

purian23 (approved):
Thanks for fixing it up, ensure you request a re-review once you changes are in so maintainers can move forward.

euletheia:
Sorry for the additional commit, that was not intended.

My apologies for the clumsy PR and thank you very much for your patience :pray:

Issue #2491, opened 2026-05-25
IPC option for managing power profiles


DBTDerpbox:
### Feature Description

An IPC command for managing the power profiles would be nice to have instead of having to go and manually open the menu every time

### Use Case

I came from KDE and got very used to pressing `Super+B` to quickly change power profiles without having to go manually open its menu and would like to recreate the behavior in DMS

I would bind `Mod+B` to `dms ipc powerprofile cycle` to easily switch between them

### Compositor(s)

All compositors

### Proposed Solution

Something along the lines of `dms ipc powerprofile [option]` with the options being `close, open, toggle, list, set, cycle`
You would be able to open the power profile menu and also change which one is set with IPC
Since there are 3 options, I think there should be an option to change them in order with `cycle` so you could go from `Power Saver` -> `Balanced` -> `Performance` and then back to `Power Saver` by sending a single command repeatedly

### Alternatives/Existing Solutions

I know you can do this with stuff like `busctl`, but DMS isn't aware of these changes so it won't show a popup indicating that it got changed, and the menu will have outdated information until you manually update it in DMS's power profile menu

hthienloc:
@DBTDerpbox The PR has been merged, so you can check it out. I forgot to let you know earlier.

PR #2515, opened 2026-05-28
feat(ipc): add native powerprofile target for power profiles management

hthienloc:
## What
Added a native `powerprofile` target to the DMS IPC system (`DMSShellIPC.qml`), supporting `open`, `close`, `toggle`, `list`, `set`, and `cycle` functions.

## Why
Addresses #2491. Provides an elegant, native command-line interface to read, write, and cycle power profiles (via `power-profiles-daemon` D-Bus integration) without depending on external tools like `busctl` or `powerprofilesctl`, keeping DMS state synced and triggering OSD notifications natively.

## How Tested
Validated syntactical correctness with `qmllint`. Verified that `dms ipc` successfully connects and registers the new `powerprofile` target.

image

### DIFF
--- a/quickshell/DMSShell.qml
+++ b/quickshell/DMSShell.qml
@@ -1185,6 +1185,24 @@ Item {
         }
     }
 
+    LazyLoader {
+        id: powerProfileModalLoader
+
+        active: false
+
+        PowerProfileModal {
+            id: powerProfileModal
+
+            Component.onCompleted: {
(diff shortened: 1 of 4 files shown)


Purian23:
Thanks for the PR, I'm going to add a status call on these similar to those of brightness status etc. Then we'll need to update the [danklinux docs](https://danklinux.com/docs/dankmaterialshell/keybinds-ipc) 

Edit: 
The docs have been updated.

PR #2517, opened 2026-05-28
feat(DiskUsage): updated dynamic width options for DiskUsage widget

Bogdan-Torkhov:
image

Added these button

image

Now sizing looks clean and the same


Purian23:
Thanks for the PR, I pushed some updates. The widget had been hardcoded to percent-only while Settings still exposed all `diskUsageMode`, that has been fixed. 
Ensure you have QML / QT extensions loaded in your DE in the future to prevent the bad formatting extra blank lines and noise. Thanks!

PR #2521, opened 2026-05-29
fix(settings): support localized settings search

jbwfu:
## Summary

  - Support searching settings with both English source text and the active localized language
  - Rebuild the settings search cache when translations load or change
  - Keep English search working when translations are incomplete


Purian23:
Thanks!

Issue #1555, opened 2026-01-31
Checkboxes/Radios are solid blocks in Light Mode


hnsylzg:
### Compositor

- [ ] Niri
- [x] Hyprland
- [ ] MangoWC (dwl)
- [ ] Sway

### Distribution

- [x] Arch Linux
- [ ] CachyOS
- [ ] Fedora
- [ ] NixOS
- [ ] Debian
- [ ] Ubuntu
- [ ] Gentoo
- [ ] OpenSUSE
- [ ] Other (specify below)

### dms doctor -v

dms doctor -v                                                                         ✔

██████╗  █████╗ ███╗   ██╗██╗  ██╗
██╔══██╗██╔══██╗████╗  ██║██║ ██╔╝
██║  ██║███████║██╔██╗ ██║█████╔╝
██║  ██║██╔══██║██║╚██╗██║██╔═██╗
██████╔╝██║  ██║██║ ╚████║██║  ██╗
╚═════╝ ╚═╝  ╚═╝╚═╝  ╚═══╝╚═╝  ╚═╝

 System Health Check

──────────────────────────────────────

  System
    ● Operating System ····· Arch Linux
      └─ ID: arch, Version: rolling, Arch: amd64
    ● Architecture ········· amd64
    ● Display Server ······· Wayland
      └─ WAYLAND_DISPLAY=wayland-1

  Versions
    ● DMS CLI ·············· dms v1.2.0-100-g15dc91f7
      └─ /usr/bin/dms
    ● Quickshell ··········· quickshell 0.2.1, revision 1e4d804e7f3fa7465811030e8da2bf10d544426a,distributed by: AUR (package: quickshell-git)
      └─ /usr/bin/qs
    ● DMS Shell ············ v1.4-unstable
      └─ /usr/share/quickshell/dms

  Installation
    ● DMS Configuration ···· Found
      └─ /usr/share/quickshell/dms
    ● shell.qml ············ Present
      └─ /usr/share/quickshell/dms/shell.qml
    ○ Install Type ········· System package
      └─ /usr/share/quickshell/dms

  Compositor
    ● Hyprland ············· 0.53.2
      └─ /usr/bin/hyprland
    ○ Active ··············· Hyprland

  Quickshell Features
    ● Polkit ··············· Available
      └─ Authentication prompts
    ● IdleMonitor ·········· Available
      └─ Idle detection
    ● IdleInhibitor ········ Available
      └─ Prevent idle/sleep
[...]

PR #2523, opened 2026-05-29
feat(Themes): add support for GTK3 dynamic theming

euletheia:
## 'Proof of concept'

Instead of overriding named colors of all GTK3 themes via the config directory file, override only named colors for each variant of the adw-gtk3 theme (light and dark), and let GTK3 apps switch theme variant depending on the value of org.gnome.desktop.interface color-scheme.

This implies :
- that GTK3 apps will auto adjust their theme in real time, without having to restart them ;
- and that we can get rid of the stylesheet duplicata in the `gtk-light-colors.css` matugen template[^1].

[^1]: At least my tests showed no blatant issue without those 8000 LOC.

However, the main shortcoming of this method is having to rely on the adw-gtk3 theme being installed in the user themes directory[^2] when package managers will install it system-wide in `/usr/share`.

[^2]: Considering they don't use the legacy `$HOME/.themes` folder.

This also entails that theming GTK3 apps without adw-gtk3 (via the `.config` override) would no longer be provided by DMS (at least I did not write anything to provide a fallback/choice). 

Eventually, I don't know on which side you think the burden should weigh, but a stable solution could be to vendor the theme[^3].

[^3]: with a "dank" prefix to differentiate it from the original ?

Obviously, considering the size of the change, I could have overlooked some issues, but I've been using this patch for roughly a week and it seems stable.

## Screenshots

https://github.com/user-attachments/assets/8c01ff44-2009-4872-b6a3-c2936ea454e8

### DIFF
--- a/quickshell/Common/Theme.qml
+++ b/quickshell/Common/Theme.qml
@@ -144,7 +144,7 @@ Singleton {
         Quickshell.execDetached(["mkdir", "-p", stateDir]);
         // shellDir may be an embedded-UI extraction, which is read-only and
         // unexecutable (dankgo shellapp/shellfs makeReadOnly chmods 0444)
-        Quickshell.execDetached(["bash", shellDir + "/scripts/gtk.sh", configDir, "", shellDir]);
+        Quickshell.execDetached(["bash", shellDir + "/scripts/gtk.sh", configDir, "", "", shellDir]);
         Proc.runCommand("matugenCheck", ["sh", "-c", "command -v matugen"], (output, code) => {
             matugenAvailable = (code === 0) && !envDisableMatugen;
 
@@ -1926,7 +1926,7 @@ Singleton {
         return colors;
     }
 
-    function refreshGtk() {
+    function refreshGtkTheme() {
         const isLight = (typeof SessionData !== "undefined" && SessionData.isLightMode);
         const theme = isLight ? "adw-gtk3" : "adw-gtk3-dark";
         const schema = "org.gnome.desktop.interface";
@@ -1952,6 +1952,15 @@ Singleton {
         });
     }
 
+    function includeGtk3Colors() {
+        const isLight = (typeof SessionData !== "undefined" && SessionData.isLightMode);
+        Proc.runCommand("gtk3Includer", ["bash", shellDir + "/scripts/gtk.sh", configDir, "include", isLight, shellDir], (output, exitCode) => {
+            if (exitCode !== 0) {
+                log.warn(`Failed to include GTK3 colors: ${output}`);
+            }
+        });
+    }
+
     function applyGtkColors() {
         if (!matugenAvailable) {
             if (typeof ToastService !== "undefined") {
@@ -1961,12 +1970,11 @@ Singleton {
         }
 
         const isLight = (typeof SessionData !== "undefined" && SessionData.isLightMode) ? "true" : "false";
-        Proc.runCommand("gtkApplier", ["bash", shellDir + "/scripts/gtk.sh", configDir, isLight, shellDir], (output, exitCode) => {
+        Proc.runCommand("gtkApplier", ["bash", shellDir + "/scripts/gtk.sh", configDir, "apply", isLight, shellDir], (output, exitCode) => {
             if (exitCode === 0) {
                 if (typeof ToastService !== "undefined" && typeof NiriService !== "undefined" && !NiriService.matugenSuppression) {
                     ToastService.showInfo(I18n.tr("GTK colors applied successfully"));
                 }
-                refreshGtk();
             } else {
                 if (typeof ToastService !== "undefined") {
                     ToastService.showError(I18n.tr("Failed to apply %1 colors").arg("GTK"));
@@ -2159,8 +2167,9 @@ Singleton {
             }
 
             if (!pendingThemeRequest) {
-                if (SettingsData.matugenTemplateGtk || root.gtkThemingEnabled) {
-                    applyGtkColors();
+                if (SettingsData.matugenTemplateGtk) {
+                    includeGtk3Colors();
+                    refreshGtkTheme();
                 }
                 return;
             }
--- a/quickshell/Common/Theme.qml
+++ b/quickshell/Common/Theme.qml
@@ -144,7 +144,7 @@ Singleton {
         Quickshell.execDetached(["mkdir", "-p", stateDir]);
         // shellDir may be an embedded-UI extraction, which is read-only and
         // unexecutable (dankgo shellapp/shellfs makeReadOnly chmods 0444)
-        Quickshell.execDetached(["bash", shellDir + "/scripts/gtk.sh", configDir, "", shellDir, "assets-only"]);
+        Quickshell.execDetached(["bash", shellDir + "/scripts/gtk.sh", configDir, "", shellDir]);
         Proc.runCommand("matugenCheck", ["sh", "-c", "command -v matugen"], (output, code) => {
             matugenAvailable = (code === 0) && !envDisableMatugen;
 
@@ -1937,7 +1937,7 @@ Singleton {
                 return `gsettings set ${schema} ${key} '' && gsettings set ${schema} ${key} ${val}`;
             } else {
                 const dconfPath = `/${schema.replace(/\./g, "/")}`;
-                return `dconf write ${dconfPath} ${key} "''" && dconf write ${dconfPath} ${key} "'${val}'"`;
+                return `dconf write ${dconfPath}/${key} "''" && dconf write ${dconfPath}/${key} "'${val}'"`;
             }
         };
 
--- a/quickshell/Common/Theme.qml
+++ b/quickshell/Common/Theme.qml
@@ -1926,6 +1926,32 @@ Singleton {
         return colors;
     }
 
+    function refreshGtk() {
+        const isLight = (typeof SessionData !== "undefined" && SessionData.isLightMode);
+        const theme = isLight ? "adw-gtk3" : "adw-gtk3-dark";
+        const schema = "org.gnome.desktop.interface";
+        const key = "gtk-theme";
+
+        const makeCmd = (tool, schema, val) => {
+            if (tool === "gsettings") {
+                return `gsettings set ${schema} ${key} '' && gsettings set ${schema} ${key} ${val}`;
+            } else {
+                const dconfPath = `/${schema.replace(/\./g, "/")}`;
+                return `dconf write ${dconfPath} ${key} "''" && dconf write ${dconfPath} ${key} "'${val}'"`;
+            }
+        };
+
+        Proc.runCommand("gtkRefresher", ["sh", "-c", makeCmd("gsettings", schema, theme)], (output, exitCode) => {
+            if (exitCode !== 0) {
+                Proc.runCommand("gtkRefreshFallback", ["sh", "-c", makeCmd("dconf", schema, theme)], (output, exitCode) => {
+                    if (exitCode !== 0) {
+                        log.warn("Failed to refresh gtk-theme");
+                    }
+                });
+            }
+        });
+    }
+
     function applyGtkColors() {
         if (!matugenAvailable) {
             if (typeof ToastService !== "undefined") {
@@ -1940,6 +1966,7 @@ Singleton {
                 if (typeof ToastService !== "undefined" && typeof NiriService !== "undefined" && !NiriService.matugenSuppression) {
                     ToastService.showInfo(I18n.tr("GTK colors applied successfully"));
                 }
+                refreshGtk();
             } else {
                 if (typeof ToastService !== "undefined") {
                     ToastService.showError(I18n.tr("Failed to apply %1 colors").arg("GTK"));
@@ -2131,8 +2158,12 @@ Singleton {
                 root.matugenCompleted(currentMode, "error");
             }
 
-            if (!pendingThemeRequest)
+            if (!pendingThemeRequest) {
+                if (SettingsData.matugenTemplateGtk || root.gtkThemingEnabled) {
+                    applyGtkColors();
+                }
                 return;
+            }
 
             const req = pendingThemeRequest;
             pendingThemeRequest = null;
--- a/quickshell/Common/Theme.qml
+++ b/quickshell/Common/Theme.qml
@@ -1952,11 +1952,11 @@ Singleton {
         });
     }
 
-    function includeGtk3Colors() {
+    function patchGtk3colors() {
         const isLight = (typeof SessionData !== "undefined" && SessionData.isLightMode);
-        Proc.runCommand("gtk3Includer", ["bash", shellDir + "/scripts/gtk.sh", configDir, "include", isLight, shellDir], (output, exitCode) => {
+        Proc.runCommand("gtk3Patcher", ["bash", shellDir + "/scripts/gtk.sh", configDir, "patch", isLight, shellDir], (output, exitCode) => {
             if (exitCode !== 0) {
-                log.warn(`Failed to include GTK3 colors: ${output}`);
+                log.warn(`Failed to patch GTK3 colors: ${output}`);
             }
         });
     }
@@ -2168,7 +2168,7 @@ Singleton {
 
             if (!pendingThemeRequest) {
                 if (SettingsData.matugenTemplateGtk) {
-                    includeGtk3Colors();
+                    patchGtk3colors();
                     refreshGtkTheme();
                 }
                 return;
--- a/quickshell/scripts/gtk.sh
+++ b/quickshell/scripts/gtk.sh
@@ -37,7 +37,7 @@ remove_gtk3_include() {
 	local theme_dir="$1"
 	local css_variant="$2"
 	[ "$css_variant" != "-dark" ] && css_variant=""
-	sed -i.backup '/\/\* BEGIN DMS OVERRIDE \*\//,/\/\* END DMS OVERRIDE \*\//d' "${theme_dir}/gtk${css_variant}.css"
+	sed -i '/\/\* BEGIN DMS OVERRIDE \*\//,/\/\* END DMS OVERRIDE \*\//d' "${theme_dir}/gtk${css_variant}.css"
 	return $?
 }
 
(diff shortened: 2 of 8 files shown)


real-or-random:
> This implies :
> 
>  * that GTK3 apps will auto adjust their theme in real time, without having to restart them ;

This will be great.

> 
>   * and that we can get rid of the stylesheet duplicata in the `gtk-light-colors.css` matugen template[1](#user-content-fn-1-9cc3df207228c32d13ca891f39ea6e11).

This would probably also solve #1555 as a side effect.

> This also entails that theming GTK3 apps without adw-gtk3 (via the `.config` override) would no longer be provided by DMS (at least I did not write anything to provide a fallback/choice).

Sounds acceptable to me. The current way of simply overwriting everything with the adw3 light theme is certainly not better. 

> Eventually, I don't know on which side you think the burden should weigh, but a stable solution could be to vendor the theme

 If DMS is committed to adw-gtk3 anyway, it can also ship it. And if a prefix is added, this really keeps it separate from the system installation.

Spoch-dev:
Closing

euletheia:
Following 8f100960cf176f33a21cc207e87d4246dd25578e : 
 - reverted the majority of the changes to the Go and Qml code ;
 - concentrated the logic in the `gtk.sh` script[^1].

[^1]: by recycling the check for adw-gtk3 installation directory introduced by the aforementioned commit.

If iterating over the adw-gtk3 installation directory candidates in order of preference is enough, then I gladly embrace the minimalism of the solution, and there is - I think - no need for any discussion at all.

N.B. : this also reverts part of 8f100960cf176f33a21cc207e87d4246dd25578e (since adw-gtk3 assets healing is no longer needed)

euletheia:
So, I spoke too soon and way too early in the morning and things were not exactly correct to say the least.
Everything should be working now.

real-or-random:
@euletheia Is this ready for testing from your side?

euletheia:
Added the missing update of GTK theme on dms theme change by re-implementing `matugen.go:refreshGTK():910` in `Theme.qml:1958`.

@real-or-random Everything is even more ready and should be working as expected (don't forget to compile dms so that `isDMSGTKActive()` performs its checks on gtk4 files instead of gtk3).

Sorry for the repeated updates.

real-or-random:
Okay, I installed this locally. A few observations.

When I started DMS for the first time, I didn't have a local copy of the theme. DMS greeted me with "failed to apply GTK colors" (or similar). Switching light/dark themes still works, but the colors are wrong. (I think that's exactly what we expect in the case of this error message).

After that first failure, clicking "Apply GTK colors" in Settings results in a confirmation "GTK colors applied succesfully". But the  `gtk.sh` script still failed and the colors didn't work. When I run it manually, I see what's going on: The GTK3 part fails when I run it manually but the failure is not propagated to the exit code:
```
Info: backing up and removing global theme override for gtk3.
sed: couldn't open temporary file /usr/share/themes/adw-gtk3/gtk-3.0/sedd1KOII: Permission denied
sed: couldn't open temporary file /usr/share/themes/adw-gtk3-dark/gtk-3.0/sedruYvaR: Permission denied
GTK4 import already exists
GTK colors applied successfully
```

I copied the theme:
```
cp -r /usr/share/themes/adw-gtk3 ~/.local/share/themes/
cp -r /usr/share/themes/adw-gtk3-dark ~/.local/share/themes/
```

Then the script works, and the colors will be applied even in running apps. At least in light mode. Dark mode colors seem to be the adw-gtk3-dark default, and setting different colors in the DMS settings has no effect on dark mode colors in gtk3 apps. 

And changing colors in running apps is not perfect (both after setting a different color of after toggling light/dark). For some UI elements, a redrawing needs to be forced manulaly For example, the background color of the active search bar in Evolution doesn't change when toggling dark/light mode. Only after starting a new search, it is redrawn and the new color applies. Not sure if anything can be done here or if this is even a bug in Evolution.

Does the initial error message at DMS startup also come from a run of the script? I haven't looked at the code, but shouldn't the script only run when clicking the button? If the error message is a result of script, then Why did the inital run of the script have a bad exit status? I don't know. Maybe ran before matugen and hit this error path?
  
Summary / conclusions:
 - I think we should never try to write to system paths. Vendoring seems like the proper approach to avoid all kinds of issues here.
 - Light/dark switching works well.
 - Matugen colors are not applied in dark mode. (Are you sure that dark mode uses the `adw-gtk3-dark` theme? Shouldn't it use the `gtk-dark.css` file inside the `adw-gtk3` theme? It would make much more sense if a single theme controls both light and dark mode. I think `adw-gtk3-dark` is simply a different theme and we don't need it at all here.)
 - Color changing in running apps is not perfect (but still much better than before).
 - What about this failed to "apply GTK colors" error message at the first DMS startup after installation?  (I didn't see it in further starts of DMS but I already had copied the themes.)


One more thought on the approach in general? Do you have an idea why overriding colors like this works while writing to `~/.config/gtk-3.0/gtk.css` doesn't? It still feels to me like the latter should be the proper approach, and that there should be some way to make it work.

quickshell/Common/Theme.qml
```diff
@@ -2160,8 +2187,12 @@ Singleton {
                 root.matugenCompleted(currentMode, "error");
             }
 
-            if (!pendingThemeRequest)
+            if (!pendingThemeRequest) {
```
bbedward, line 2169: This runs and shows an error toast if adw-gtk3 is not installed, it should be gated better to not run unconditionally (matugenTemplateGtk and gtkTheemingEnabled are more or less, true by default in most cases)

quickshell/scripts/gtk.sh
```diff
@@ -1,111 +1,95 @@
...
+		sed -i "1i\\$gtk4_import" "$gtk_css"
+	else
+		echo "$gtk4_import" >"$gtk_css"
+	fi
+	echo "Updated GTK4 CSS import"
 }
 
-# Repair pass for shells whose gtk.css was linked before asset handling
-# existed; only acts when DMS already manages gtk.css.
-if [ "$4" = "assets-only" ]; then
```
bbedward, line 102: The caller of this still exists in Theme.qml

quickshell/Common/Theme.qml
```diff
@@ -1955,6 +1955,32 @@ Singleton {
...
+        const theme = isLight ? "adw-gtk3" : "adw-gtk3-dark";
+        const schema = "org.gnome.desktop.interface";
+        const key = "gtk-theme";
+
+        const makeCmd = (tool, schema, val) => {
+            if (tool === "gsettings") {
+                return `gsettings set ${schema} ${key} '' && gsettings set ${schema} ${key} ${val}`;
+            } else {
+                const dconfPath = `/${schema.replace(/\./g, "/")}`;
+                return `dconf write ${dconfPath} ${key} "''" && dconf write ${dconfPath} ${key} "'${val}'"`;
```
bbedward: I believe this is malformed, dconf write takes 2 arguments but this would pass 3

quickshell/scripts/gtk.sh
```diff
@@ -1,111 +1,95 @@
...
-
-    link_gtk3_assets "$gtk3_dir"
+	local config_dir="$1"
+
+	# Make sure there's no global override
+	local gtk3_dir_cfg="$config_dir/gtk-3.0"
+	local gtk3_override="$gtk3_dir_cfg/gtk.css"
+	if [ -f "$gtk3_override" ]; then
+		echo "Info: backing up and removing global theme override for gtk3."
+		mv "$gtk3_override" "$gtk3_override.backup"
```
bbedward: The migration path s hould delete file not back it up, its just a symlink

bbedward (changes requested):

euletheia:
@bbedward I've addressed the issues you pointed out in your review in https://github.com/AvengeMedia/DankMaterialShell/pull/2523/commits/335d71e483caacbc45a72cd95e876426d5f2d8e3, except for the lack of guard in the Theme.qml execution block since after @real-or-random tests I realized there was, IMO, a better way of handling things (see https://github.com/AvengeMedia/DankMaterialShell/pull/2523/commits/431e9c80a61508738ff7c6b243dafa99503a8685):

- the check is performed by `gtk.sh` when including theme changes, and the script fails with a warning and no user-side notification if it can't write to a user version of adw-gtk3.

Additionally, in a3fba8fcb6d8fd88d303d2f2d66626a6b255e42e you introduced the `DMS_ENABLE_GTK4_REFRESH` but as I eventually found out, some GTK3 applications rely on [`gtk-application-prefer-dark-theme`](https://docs.gtk.org/gtk3/property.Settings.gtk-application-prefer-dark-theme.html), but also set it directly (either via [`Gtk.Settings`](https://docs.gtk.org/gtk3/class.Settings.html) or via third-party libs such as [`libhandy`](https://gnome.pages.gitlab.gnome.org/libhandy/doc/main/) like `dconf-editor` does), meaning we have to:

1. Overwrite the `gtk-dark.css` of the current `adw-gtk3` variant theme[^1] ;
2. 'Refresh' the `color-scheme` variable so that the app updates its stylesheet.

[^1]: Since `gtk-application-prefer-dark-theme` causes GTK3 apps to use the dark variant of their current theme.

If I'm not mistaken, this `color-scheme` 'refresh' would be needed for all apps relying on `gtk-application-prefer-dark-theme`.
That's why I propose in f19ad82b9c6e240a3041b29ebcd026dd37e9a63c to:

- rename `DMS_ENABLE_GTK4_REFRESH` to the more general `DMS_ENABLE_GTK_REFRESH` ;
- reset the timer between 'refresh' steps to 50ms to avoid any app "flashing" before reloading their correct theme[^2].

[^2]: So far I've been unable to reproduce #2552.

---

@real-or-random Thanks a lot for testing and for your feedback ! You've prompted me to investigate the matter more thoroughly, and caught most of my oversights.

Your issue with Evolution should be fixed now (at least I can't seem to reproduce it right now).

Regarding your questions on this approach to GTK3 "theming", I have attempted to summarise my understanding of the situation below, but note that because _GTK does not support **user** theming in any way_, there's no _"proper"_ way to do it.

### GTK theming[^3]

[^3]: This information is gathered only from my tests and the GTK documentation, I've not delved into the implementation and the project history, so please feel free to correct me.

First, the method on which pretty much every user-side GTK theming relies is to use the global stylesheet override provided through [`$XDG_CONFIG_HOME/gtk-[3|4].0/gtk.css`](https://docs.gtk.org/gtk3/class.CssProvider.html).

GTK3 typically loads that stylesheet at startup and doesn’t reliably pick up changes at runtime, with (at least) one exception : if the theme is changed, a new stylesheet is loaded by the application.

However, the override will stay loaded along with the original theme one, unless we explicitly unset `gtk-theme` (or change it to `default`). It also prevents any visual change whenever `dms` automatically switches from `adw-gtk3` to `adw-gtk3-dark` upon toggling `color-scheme` (since `dank-colors.css` overrides all colors). 

To allow dynamic theming of GTK3 applications, we therefore need to modify the theme stylesheet itself ("include" our colors), then "refresh it" (unload, then load it again via `gsettings set org.gnome.desktop.interface gtk-theme` or the `dconf` equivalent path).

As for the "apply" operation available in DMS settings, it's the preliminary step to theming. DMS handles it by setting an `@import` rule in the global override file that loads the matugen-generated stylesheet.
In the new setup, "apply" for GTK3 means ensuring no global override exists, unless `adw-gtk3` is only installed system-wide. In that case, since we (probably) don't have permission to append our colors to the `adw-gtk3` file, and don't vendor the theme (yet), we need to fall back on the global override solution.

Concerning the `gtk-dark.css` file, it's indeed an alternative mechanism to control theme mode. This will be picked up on startup by GTK3 applications if the [`gtk-application-prefer-dark-theme`](https://docs.gtk.org/gtk3/property.Settings.gtk-application-prefer-dark-theme.html) option is set in `settings.ini`, either in a configuration directory, or in the theme directory itself. BUT it can also be set by the application.

Since, as far as `adw-gtk3` is concerned, both `adw-gtk3-dark` stylesheet and the dark variant of the light theme are the same file, the simplest approach I found is to neutralize the option by putting our dark colors in these files, and the light ones on the remaining file.

real-or-random:
Sorry, too many moving parts.


@bbedward 
I'm trying to test this via a patched of the AUR package that points at the PR branch. The AUR thing worked before the change to bundle quickshell assets in the dms binary. Now after this change, it seems that the gtk.sh can't be found:

```
  Jul 21 09:09:56 dms[93539]:   WARN: Process failed to start, likely because the binary could not be found. Command: QList("/run/user/1000/danklinux-shell/468f944f7286aecf/scripts/gtk.sh", "/home/tim/.config", "include", "true", "/run/user/1000/danklinux-shell/468f944f7286aecf")
```

---

When I download the script and run `apply` and `include` manually, I still see the default theme colors instead of the matugen colors in dark mode.

@euletheia Am I supposed to have a local copy of `gtk3-adw-dark` or only `gtk3-adw`?

bbedward:
Your branch might need rebased? There was a fix for that one of the last couple days.

Easier way to test though is to just run `dms run -c /path/to/dms/quickshell` with your local copy. Or since yours involves a go change you can just `cd core; make` then `dms kill; ./bin/dms run` to run your local copy in full with embedded assets

quickshell/Common/Theme.qml
```diff
@@ -1926,6 +1926,41 @@ Singleton {
...
+                        log.warn("Failed to refresh gtk-theme");
+                    }
+                });
+            }
+        });
+    }
+
+    function includeGtk3Colors() {
+        const isLight = (typeof SessionData !== "undefined" && SessionData.isLightMode);
+        Proc.runCommand("gtk3Includer", [shellDir + "/scripts/gtk.sh", configDir, "include", isLight, shellDir], (output, exitCode) => {
```
bbedward: With the embedded setup this might be the problem, it should be ran explicitly with the interpreter `"bash" "/scripts/gtk.sh"`

bbedward (changes requested):

quickshell/Common/Theme.qml
euletheia: No wonder DMS didn’t pick up the script unless launched via `make dev`... :facepalm:

euletheia:
> @euletheia Am I supposed to have a local copy of `gtk3-adw-dark` or only `gtk3-adw`?

Both are needed.

quickshell/scripts/gtk.sh
```diff
@@ -1,114 +1,243 @@
...
-SHELL_DIR="$3"
+# apply: setup symlinks and config dirs
+# refresh: refresh overrides in adw-gtk3
+MODE="${2:-apply}"
+IS_LIGHT="${3:-light}"
 
 if [ -z "$CONFIG_DIR" ]; then
-    echo "Usage: $0 <config_dir> [is_light] [shell_dir]" >&2
-    exit 1
+	echo "Usage: $0 <config_dir> [apply|include|remove] [is_light] [shell_dir]" >&2
```
real-or-random: These descriptions of the modes don't match: "apply, refresh" vs. "[apply|include|remove]"

quickshell/scripts/gtk.sh
```diff
@@ -1,114 +1,243 @@
...
+		fi
+	done
+	echo "$target"
+}
+
+remove_gtk3_include() {
+	local theme_dir="$1"
+	local css_variant="$2"
+	[ "$css_variant" != "-dark" ] && css_variant=""
+	sed -i.backup '/\/\* BEGIN DMS OVERRIDE \*\//,/\/\* END DMS OVERRIDE \*\//d' "${theme_dir}/gtk${css_variant}.css"
```
real-or-random: Since `remove_gtk3_include()` is called in `do_include`, this backs up the file *with the override* (every time).

I think what the user meaningfully expects is that the `.backup` files are the original unmodified files. I suggest creating these backup file only the first time when we touch the files (and then don't recreate them if they exist). This will make sure that the user has backups of the original theme files in place.

real-or-random (commented):
Great, this now works nicely for me. 👍 

 - The embedded script now runs properly.
 - The colors are now applied correctly in dark mode. 
 - The only thing that I can still observe is the that some rare UI elements (e.g. background color of search box in Evolution) are not updated on a color change, and need a "manual" refresh by interacting with the UI. But this is super minor and I tend to believe it's a bug in gtk3 or in Evolution.

As for further steps, which could happen in a separate PR (this is a maintainer call obviously), I still believe that vendoring the themes would be the cleanest thing to do. Or alternatively, we'll need proper docs that tell the user to have local copies of adw3-gtk and adw3-gtk-dark. If we decide to vendor, then I believe it will be fine to get rid of the fallback override entirely.

euletheia:
Depending on what seems right to the maintainers, I'd be happy to submit a PR for either the docs or the themes vendoring.

quickshell/scripts/gtk.sh
```diff
@@ -1,114 +1,251 @@
 #!/usr/bin/env bash
 
 CONFIG_DIR="$1"
-SHELL_DIR="$3"
+# apply: setup symlinks and config dirs
+# refresh: refresh overrides in adw-gtk3
```
real-or-random: It still doesn't match, I think you want this:
```suggestion
# include: refresh overrides in adw-gtk3
```

euletheia: Yes thank you : seems like I wasn't quite settled on which term to use :)

EDIT: That might be splitting hairs, but I included an additional commit to make the actual behavior clearer.

real-or-random: No I agree, "patch" is much better than "include". :)

real-or-random (approved):
I haven't retested this, but the diff from the tested https://github.com/AvengeMedia/DankMaterialShell/commit/450a535427b7298ef2c6398dd83da7c3847cd875 looks fine to me. Good to go from my side.

real-or-random:
> As for further steps, which could happen in a separate PR (this is a maintainer call obviously), I still believe that vendoring the themes would be the cleanest thing to do. Or alternatively, we'll need proper docs that tell the user to have local copies of adw3-gtk and adw3-gtk-dark. If we decide to vendor, then I believe it will be fine to get rid of the fallback override entirely.

@bbedward What do you think should be the next step?

And big thanks to @euletheia for this PR and for volunteering to work on the next one.

real-or-random:
> > As for further steps, which could happen in a separate PR (this is a maintainer call obviously), I still believe that vendoring the themes would be the cleanest thing to do. Or alternatively, we'll need proper docs that tell the user to have local copies of adw3-gtk and adw3-gtk-dark. If we decide to vendor, then I believe it will be fine to get rid of the fallback override entirely.
> 
> @bbedward What do you think should be the next step?
> 
> And big thanks to @euletheia for this PR and for volunteering to work on the next one.

Friendly ping. @bbedward  The main task is done but it will be hard for users to make use of without either vendoring gadw3-gtk(-dark) or updating the docs.

bbedward:
@real-or-random Vendoring adw-gtk3 is not exactly prohibited but a little shaky since its LGPL licensed and we're modifying it via matugen.

I think I'll go with documenting the install + copying the theme from system to ~/.local/share/themes ourselves in the gtk script

PR #2529, opened 2026-05-29
Add context to font weight keys, wrap render settings, fix uptime prefix

Youseffo13:
`DgopService` returns uptime as a raw string with a hardcoded `up` prefix (e.g. `"up 1h 30m"`), making the word impossible to localize since the translated `I18n.tr("up")` was simply ignored. The fix is to slices off the first 3 characters and prepends `I18n.tr("up")` instead, so the prefix is now properly translatable.

Add `"font weight"` as a disambiguation context to all `I18n.tr()` calls for font weight names (Thin, Light, Bold, etc.), preventing key collisions with identically named strings used elsewhere in the UI.

Wrap some hardcoded strings  (`"Native"`, `"Qt"`, `"Curve"`, `"Default"`, `"Low"`, `"Normal"`, `"High"`, `"Very High"`)


purian23 (changes requested):
Thanks! 

One small fix in `UserInfoCard.qml`: Let's restore the empty-state guard so we don’t show "up " when `shortUptime` is empty:
```
text: DgopService.shortUptime
      ? I18n.tr("up") + DgopService.shortUptime.slice(2)
      : I18n.tr("up")
```

purian23 (approved):

Issue #1073, opened 2025-12-17
tray icon not show


feng-yifan:
## Compositor

niri

## Distribution

arch

## dms version

v1.0.3

## Description

Some applications like podman desktop is started by xdg autostart, but the tray icon not shown.

## Expected Behavior

the icon should shown

## Steps to Reproduce

1. copy podman desktop's desktop file to .config/autostart
2. reboot
3. icon not show

but if you end the podman desktop process and start podman desktop after dms is already start, the icon shown on tray bar. some other app meet the same situation

neoscaler:
Same is happening with Bitwarden with my Arch/Niri installation. With Ubuntu/Niri it is somehow always working and shows the icon.

Gasiyu:
+1. I’m also experiencing this on Arch running Niri.

My current workaround is using a third-party startup manager software to add a sleep delay to the launch command. Giving it a few seconds to load fixes the tray icon.

feng-yifan:
> +1. I’m also experiencing this on Arch running Niri.
> 
> My current workaround is using a third-party startup manager software to add a sleep delay to the launch command. Giving it a few seconds to load fixes the tray icon.

Before I use uwsm to manage autostart and niri, tray icon works well. But these days I switched to niri-session then it works bad

adcpe:
Just to add, I'm also on Arch/Niri. The only app that shows up on the tray at launch is Steam. MEGAsync and VEsktop/Discord do not show up but they are still running according to btop.

atimofeev:
I've managed to fix this in NixOS (uwsm & niri) with these home-manager options:
```nix
    systemd.user.services.dms = {
      Service.ExecStartPost = "${pkgs.coreutils}/bin/sleep 5";
      Unit.Before = [
        "xdg-desktop-autostart.target"
      ];
    };
```
It forces `dms` service to start before `autostart` target. 
Sleep is required here, as from what I can see, dms is not reporting its `ready` status to systemd and without it autostart apps still may miss tray functionality, as dms had no chance to finish startup yet.

feng-yifan:
I resove this by make the autostart applications to system service like this

```bash
cat .config/systemd/user/app.steam.service
[Unit]
Description=Steam
After=graphical-session.target dms.service
PartOf=graphical-session.target
BindsTo=dms.service

[Service]
Type=simple
ExecStart=/usr/bin/steam
Restart=always
RestartSec=2
TimeoutStopSec=10
[...]

PR #2535, opened 2026-05-30
feat(settings): Added Settings Tab Autostart App (XDG Autostart)

### LINKED ISSUE #1375: Autostart manager
### Feature Description

Manager for enabling and disabling startup applications.

### Use Case

It would make DMS feel like even more of a Desktop Environment than it already feels, I think it would be cool :)

### Compositor(s)

- [x] All compositors
- [ ] Niri
- [ ] Hyprland
- [ ] MangoWC (dwl)
- [ ] Sway
- [ ] Other (specify below)

### Proposed Solution

At least niri, sway and hyprland uwsm support XDG Autostart so something like a .desktop file manager pointing to ~/.config/autostart could work, alternatively on niri and on hyprland (and I suppose other compositors that support includes) there could be an additional file added called something like autostart.kdl inside the dms folder where all autostart commands get defined.

Just doing it manually through your compositor's application startup interface

arfan21:
close https://github.com/AvengeMedia/DankMaterialShell/issues/1375

- Added setting tab to generate autostart desktop entry in ~/.config/autostart
- Added button to generate systemd override file to fix autostart app icon missing in system tray https://github.com/AvengeMedia/DankMaterialShell/issues/1073#issuecomment-4312776159

image
image
image

### DIFF
--- a/quickshell/Modules/Settings/AutoStartTab.qml
+++ b/quickshell/Modules/Settings/AutoStartTab.qml
@@ -110,10 +110,6 @@ Item {
         newEntryCommandWrapper = "%command%";
     }
 
-    function generateTrayOverride() {
-        const proc = writeOverrideComponent.createObject(root, { running: true });
-    }
-
     function addOrUpdateEntry(entry) {
         var list = root.entries.slice();
         for (var i = 0; i < list.length; i++) {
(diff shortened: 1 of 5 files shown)


quickshell/Modules/Settings/AutoStartTab.qml
```diff
@@ -0,0 +1,648 @@
...
+    property var entries: []
+    property var desktopApps: []
+    property string newEntryType: "desktop"
+    property string newEntryName: ""
+    property string newEntryExec: ""
+    property string newEntryDesktopId: ""
+    property string newEntryCommandWrapper: "%command%"
+
+    readonly property string autostartDir: {
+        const configHome = Quickshell.env("XDG_CONFIG_HOME") || Quickshell.env("HOME") + "/.config";
```
bbedward: Prefer to use QtCore StandardPaths for this stuff, e.g.

```qml
    readonly property string _homeDir: Paths.strip(StandardPaths.writableLocation(StandardPaths.HomeLocation))
    readonly property string _configDir: Paths.strip(StandardPaths.writableLocation(StandardPaths.ConfigLocation))
```

quickshell/Modules/Settings/AutoStartTab.qml
```diff
@@ -0,0 +1,648 @@
...
+    }
+
+    function generateTrayOverride() {
+        const proc = writeOverrideComponent.createObject(root, { running: true });
+    }
+
+    Component {
+        id: readDirComponent
+        Process {
+            command: ["sh", "-c", "ls -1 \"" + root.autostartDir + "\"/*.desktop 2>/dev/null || true"]
```
bbedward: Ideally we avoid processes for I/O operations and use Quickshell FileView + QML [folderlistmodel](https://doc.qt.io/qt-6/qml-qt-labs-folderlistmodel-folderlistmodel.html)

quickshell/Modules/Settings/AutoStartTab.qml
```diff
@@ -0,0 +1,648 @@
...
+        if (!content || content.length === 0) return null;
+        const lines = content.split("\n");
+        let name = "";
+        let execCmd = "";
+        let icon = "";
+        let isDesktopEntry = false;
+        for (let i = 0; i < lines.length; i++) {
+            const line = lines[i].trim();
+            if (line === "[Desktop Entry]") {
+                isDesktopEntry = true;
```
bbedward, line 57: Should break out of the parser after another header is found, e.g. `[Desktop Action..`

quickshell/Modules/Settings/AutoStartTab.qml
```diff
@@ -0,0 +1,648 @@
...
+            }
+        }
+    }
+
+    Component {
+        id: writeFileComponent
+        Process {
+            property string fileName: ""
+            property string fileContent: ""
+            command: ["sh", "-c", "mkdir -p \"" + root.autostartDir + "\" && cat > \"" + root.autostartDir + "/" + fileName + "\" << 'DMS_EOF'\n" + fileContent + "\nDMS_EOF"]
```
bbedward: Quickshell [FileView](https://quickshell.org/docs/v0.2.1/types/Quickshell.Io/FileView/) can write files

quickshell/Modules/Settings/AutoStartTab.qml
```diff
@@ -0,0 +1,648 @@
...
+
+                    DankActionButton {
+                        id: clearAllButton
+                        iconName: "delete_sweep"
+                        iconSize: Theme.iconSize - 2
+                        iconColor: Theme.error
+                        anchors.verticalCenter: parent.verticalCenter
+                        onClicked: {
+                            for (let i = 0; i < root.entries.length; i++) {
+                                root.removeEntry(root.entries[i].filePath);
```
bbedward, line 616: We could add enable/disable by setting `Hidden=true` in the desktop entry probably, but its not critical or anything. To distinguish from add/remove. That way it could stay in the list persistently and just be toggled on or off.

bbedward (dismissed):
We could probably detect if this functionality is even available in the current session.

Systemd systems should just be:

```sh
systemctl --user is-active xdg-desktop-autostart.target
```

Non-systemd systems I'm not exactly sure how to check though.  Maybe we can check both, if it's systemd system and the autostart target is active - reveal the settings. If its not a systemd system just show it regardless

arfan21:
thanks for the  review @bbedward.

Updated based on your comment. Could you please review the changes again?

also added hide toggle

image

purian23 (changes requested):
Thanks for updating this @arfan21. Most of the previous feedback looks addressed now: StandardPaths, FolderListModel/FileView, parser section break, Hidden toggle, and availability detection.

One thing I think still needs fixing before merge: after switching writes to FileView, the code no longer creates `$XDG_CONFIG_HOME/autostart` before writing a new entry. On a clean profile that directory may not exist, so adding an entry can fail. Please ensure the autostart directory exists before `writerFileView.setText()` runs, and ideally surface save failures with a toast.

Once that is fixed, we should be good to go.

arfan21:
Thanks for the review @Purian23 . Fixed it..

purian23 (approved):

PR #2536, opened 2026-05-30
feat(battery): adjust scroll/click behavior for brightness and power profiles

hthienloc:
## Summary
Adjusted the interaction model for the Battery widget to make common tasks more accessible.

## Rationale
- Using the scroll wheel exclusively for power profile cycling is redundant and less frequently needed.
- Screen brightness adjustment is a more common task that benefits greatly from scroll wheel interaction.

## Changes
- **Scroll**: Now adjusts screen brightness (5% steps).
- **Right-Click**: Now cycles through available Power Profiles (Power Saver, Balanced, Performance).
- **Left-Click**: Preserved (toggles the battery detailed popout).

### DIFF
--- a/quickshell/Modules/DankBar/Widgets/Battery.qml
+++ b/quickshell/Modules/DankBar/Widgets/Battery.qml
@@ -33,6 +33,27 @@ BasePill {
 
     visible: true
 
+    function cyclePowerProfiles() {
+        if (typeof PowerProfiles === "undefined") {
+            ToastService.showError(I18n.tr("power-profiles-daemon not available"));
+            return;
+        }
+
+        const profiles = [PowerProfile.PowerSaver, PowerProfile.Balanced].concat(PowerProfiles.hasPerformanceProfile ? [PowerProfile.Performance] : []);
+        var index = profiles.findIndex(profile => PowerProfiles.profile === profile);
+
+        // Cycle through profiles
+        index = (index + 1) % profiles.length;
+
+        PowerProfiles.profile = profiles[index];
+        if (PowerProfiles.profile !== profiles[index]) {
+            ToastService.showError(I18n.tr("Failed to set power profile"));
+        } else {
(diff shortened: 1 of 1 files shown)


Purian23:
Hi there, I don't think this is the direction we want to go with the battery widget to start controlling brightness. Though I already don't like scroll as-is on the widget either. Since brightness is already available via control center / IPC from the other PR, we don’t want to reassign battery scroll/right-click for this.

hthienloc:
@Purian23 I understand the concern about maintaining a clean separation of responsibilities between widgets. 

The main motivation behind this PR is addressing user intuition and UX consistency:

1. **Current Redundancy**: Scrolling to cycle through power profiles is an unconventional behavior that users rarely guess or need to perform that frequently. 
2. **UX Expectation**: In many desktop environments, scrolling over the system tray status indicators (like volume or battery/power) is widely expected to adjust primary continuous variables, specifically volume and brightness. Since the volume icon handles volume, the power/battery icon is the most intuitive target for brightness control.
3. **Accessibility**: While brightness is accessible via the Control Center, requiring a click to open a panel just to change brightness slows down workflow, especially for mouse-driven setups. A quick scroll over the panel icon provides immediate, low-friction control.

Purian23:
@hthienloc I'm not sure you have convinced me here with what sounds like an AI reply lol. It looks like only KDE uses that dual behavior. I couldn't find any others beyond that worth mentioning.

hthienloc:
@Purian23 I'm not a native English speaker so I write my idea for AI to refine my wording. You can close this PR if you want and I don't mind it. Just after some times playing around with plugins I have learned to bake imporant feature in right click / scrolling feature for users who prefer mouse over keyboard can feel comfortable.

Purian23:
@hthienloc I just asked in the discord dms chat, feel free to see the replies. I like changing the scroll to a right-click, just not sure about also converting it to a brightness control.

hthienloc:
Closed due to insufficient proof.

PR #2542, opened 2026-05-31
fix(wallpaper): redraw wallpaper layers when fill mode changes

jbwfu:
Fix wallpaper not updating when fill mode changes.

### DIFF
--- a/quickshell/Modules/BlurredWallpaperBackground.qml
+++ b/quickshell/Modules/BlurredWallpaperBackground.qml
@@ -126,6 +126,14 @@ Variants {
                 }
             }
 
+            Connections {
+                target: SettingsData
+                function onWallpaperFillModeChanged() {
+                    root._renderSettling = true;
+                    renderSettleTimer.restart();
+                }
+            }
+
             Timer {
                 id: renderSettleTimer
                 interval: 1000
--- a/quickshell/Modules/WallpaperBackground.qml
+++ b/quickshell/Modules/WallpaperBackground.qml
@@ -161,6 +161,11 @@ Variants {
                     root._overviewBlurSettling = true;
                     overviewBlurSettleTimer.restart();
                 }
+
+                function onWallpaperFillModeChanged() {
+                    root._renderSettling = true;
+                    renderSettleTimer.restart();
+                }
             }
 
             Connections {


quickshell/Modules/BlurredWallpaperBackground.qml
```diff
@@ -126,6 +126,14 @@ Variants {
...
             }
 
+            Connections {
+                target: SettingsData
+                function onWallpaperFillModeChanged() {
+                    root._renderSettling = true;
+                    renderSettleTimer.restart();
+                }
+            }
+
```
jbwfu, line 136: While testing this, I noticed an interesting side effect if the fill-mode redraw is only applied to WallpaperBackground and not BlurredWallpaperBackground. With Blur Wallpaper Layer enabled, the blurred duplicate can retain the previous composition while the main wallpaper switches to the new fill mode, which creates a surprisingly nice layered effect. I don't think it should block this fix, but it might be worth considering as an optional feature in the future.

- desktop
Image

- overview
image

bbedward:
I cant reproduce the blurred mode problem honestly, it might just be a matter of bumping the timer duration since it takes longer to draw with the blur

jbwfu:
> I cant reproduce the blurred mode problem honestly, it might just be a matter of bumping the timer duration since it takes longer to draw with the blur

The reproduction steps are somewhat specific:
1. Comment out the modifications I made in `BlurredWallpaperBackground.qml`.
2. Set a relatively small image as the wallpaper (using "Tile" mode; other modes such as "Stretch" may also reproduce the issue).
3. Toggle the "Blur wallpaper" feature once, ensuring it remains enabled.
4. Set the wallpaper mode to "Pad".

---

Sorry, I'm having some trouble with my clipboard synchronization.

jbwfu:
@bbedward  Please follow the steps above; the problem should be reproducible.

quickshell/Modules/BlurredWallpaperBackground.qml
bbedward: > While testing this, I noticed an interesting side effect if the fill-mode redraw is only applied to WallpaperBackground and not BlurredWallpaperBackground. With Blur Wallpaper Layer enabled, the blurred duplicate can retain the previous composition while the main wallpaper switches to the new fill mode, which creates a surprisingly nice layered effect. I don't think it should block this fix, but it might be worth considering as an optional feature in the future.
> 
> * desktop
> 
> Image
> * overview
> 
> image

I meant this one, this fix  addresses both blurred and non-blurred layer for me.

jbwfu: Yes, I think this could be offered as an optional feature (like Blur wallpaper using a separate wallpaper mode), but it's reasonable to keep it consistent with the main wallpaper until it's determined whether that feature will be supported, so I fixed it in this PR.

What are your thoughts on this feature?

PR #2567, opened 2026-06-03
Implement process killing animation and tracking

Harshjeet:
Added functionality to track killed process IDs and animate their removal from the process list.

### DIFF
--- a/quickshell/Modules/ProcessList/ProcessesView.qml
+++ b/quickshell/Modules/ProcessList/ProcessesView.qml
@@ -16,6 +16,7 @@ Item {
     property int selectedIndex: -1
     property bool keyboardNavigationActive: false
     property int forceRefreshCount: 0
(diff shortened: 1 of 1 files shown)


purian23 (changes requested):
Hi there, thanks for the PR! The inline kill button is a really neat UX addition! 

A few items:
1. **Let's Avoid Custom Timers/Canvas**: We keep the shell primarily event-driven. Let's replace the particle canvas with a standard declarative QML transition instead (e.g., fading the row's opacity and collapsing its height to 0).
2. **Fix Infinite Binding Loop**: Currently, updating `killedPids` inside `onFilteredProcessesChanged` triggers a synchronous recursion loop that freezes the shell.

Once those adjustments are made, we should be good to go.

Purian23:
Closing in favor of updated PR #2588

PR #2580, opened 2026-06-05
fix: ignore keyboard shortcuts of disabled powermenu actions

pathmann:
Ignore keyboard shortcuts of hidden powermenu actions.

I ran into this, because previously with wlogout I had the shortcut "S" for shutdown, muscle memory kicked in and although I had suspend hidden from dms powermenu, the computer suspended.

### DIFF
--- a/quickshell/Modals/PowerMenuModal.qml
+++ b/quickshell/Modals/PowerMenuModal.qml
@@ -352,9 +352,11 @@ DankModal {
             break;
         case Qt.Key_P:
             if (!(event.modifiers & Qt.ControlModifier)) {
-                const idx = visibleActions.indexOf("poweroff");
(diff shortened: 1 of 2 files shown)


purian23 (changes requested):
Hi there, this sounds reasonable, but we need to adjust a few things: 
1. We'll want to fix the typo in `PowerMenuModal.qml` under `Qt.Key_L` to check `includes("lock")`.
2. Apply the same logic to `handleGridNavigation` in `PowerMenuModal.qml`.
3. Apply the same logic to both `handleListNavigation` and `handleGridNavigation` in `LockPowerMenu.qml`.

pathmann:
Sorry, should have looked further into it, didn't notice the separated grid event handling.

PR #2583, opened 2026-06-06
add local to-do planner / tasks to calendar overvie…

goatnath:
# Description

    
    This PR integrates a local task planner / to-do list feature directly
  into the Calendar card in the Quickshell overview tab. It allows users to
  click on any calendar day to add, toggle, and delete daily tasks locally.
  If `khal` is installed, these local planner tasks are displayed alongside
  calendar events; if `khal` is not installed, the Calendar card remains
  fully functional as a standalone local to-do planner rather than
  displaying as empty.

    ## Type of change

    

    - [ ] Documentation
    - [ ] Other

    ## Related issues

    

    ## Screenshots / video

   

https://github.com/user-attachments/assets/0a7bfdf4-5c31-4da1-8b03-b1a385a2c9a2

    ## Checklist

    - [x] New user-facing strings are wrapped in `I18n.tr()` with translator
  context, reusing existing terms where possible
    - [ ] Go changes: ran `make fmt`, added/updated tests, `make test`
  passes, and `go mod tidy` is clean
    - [x] My QML changes: ran `make lint-qml` with no new warnings
    - [ ] I have opened a corresponding pull request in dlx-docs to document
  any new behaviors: https://github.com/AvengeMedia/DankLinux-Docs

### DIFF
--- a/quickshell/Modules/DankDash/Overview/CalendarOverviewCard.qml
+++ b/quickshell/Modules/DankDash/Overview/CalendarOverviewCard.qml
@@ -433,198 +433,442 @@ Rectangle {
             }
         }
 
-        DankListView {
+        Flickable {
+            id: flickableArea
             width: parent.width - Theme.spacingS * 2
             height: parent.height - (showEventDetails ? 40 + 42 : 28 + 18) - Theme.spacingS
             anchors.horizontalCenter: parent.horizontalCenter
-            model: selectedDateEvents
             visible: showEventDetails
             clip: true
-            spacing: Theme.spacingXS
+            contentWidth: width
+            contentHeight: listViewContainer.height
 
-            delegate: Rectangle {
-                width: parent ? parent.width : 0
-                height: eventContent.implicitHeight + Theme.spacingS
-                radius: Theme.cornerRadius
-                color: {
-                    if (modelData.url && eventMouseArea.containsMouse) {
-                        return Qt.rgba(Theme.primary.r, Theme.primary.g, Theme.primary.b, 0.12);
-                    } else if (eventMouseArea.containsMouse) {
-                        return Qt.rgba(Theme.primary.r, Theme.primary.g, Theme.primary.b, 0.06);
+            Item {
+                id: listViewContainer
+                width: parent.width
+                height: 100
+                
+                property var draggedItem: null
+                property bool orderChanged: false
+                
+                function resetAndLayout() {
+                    for (let i = 0; i < repeater.count; i++) {
+                        let item = repeater.itemAt(i);
+                        if (item) {
+                            item.visualIndex = i;
+                            item.isDragging = false;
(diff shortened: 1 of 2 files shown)


hthienloc:
This looks so cool! Some feedback from the video if you want more insight:

- When entering a day or after adding a new task, the focus should automatically fall into the text input field for quickly adding tasks.
- The ability to reorder tasks.

goatnath:
@hthienloc i have made some changes since your recommendations and do tell me if this is what you envisioned

hthienloc:
Since the widget's display screen is limited, using up/down buttons for reordering might be inefficient. A designated handle for drag-and-drop reordering would be a much better approach.

The text input behavior works exactly as imagined. Great job on that.

I have a suggestion to add an inline edit button, that would allow you to keep just two icons on the right (Edit and Trash). Help users quickly adjust the task without the need to delete then recreate the task.

goatnath:
@hthienloc i have tried to make the changes you suggested and this is what i came up with, feel free to give more suggestions on possible improvements i can make

purian23 (changes requested):
Thanks for the PR @goatnath, it looks really cool! Good feedback, @hthienloc!

I'm glad khal support is optional, it's likely getting deprecated or changed in our upcoming dank calendar release. 
Otherwise, there are a few issues to address here:

1. The PR currently prevents DMS from starting because `CalendarOverviewCard.qml` is missing a closing `}` near line 958.

2. When khal finishes loading, it replaces `eventsByDate` and can wipe out tasks loaded earlier. Keeping local tasks and `khal` events separately, then merging them, would avoid this race.

3. The separate Python CRUD processes can run concurrently and overwrite each other’s changes. A single in-memory model backed by an atomic `FileView` write would also remove the Python dependency.

4. Several delegate bindings access `modelData` without checking whether it is null, which can produce QML `TypeError` warnings during model updates.

5. Completion is already stored as a boolean, so task titles should remain clean. A `completed` property and `DankIcon` would avoid adding and later parsing Unicode checkbox characters.

6. During dragging, animation is disabled for every item. Disabling it only on the dragged item would let neighboring tasks move smoothly.

7. The reorder fallback checks IDs against `id_to_item`, which contains every task, so omitted tasks can be silently dropped. It should check against the submitted ordered IDs instead.

There is also trailing whitespace in both changed files, although `make lint-qml` otherwise passes. Once these issues and the syntax error are fixed, we’ll be glad to take another look.

goatnath:
@Purian23 I have made a few  changes , even though i don't completely understand all the errors as i'm still new to this stuff but do check how it does now in the pre-commit checks.

purian23 (approved):
Thanks, this looks good to go! I pushed a formatting change, in the future be sure to use the QT/QML extensions and it should format for you and can be set up on save.

PR #2586, opened 2026-06-06
fix(ui): use primaryText instead of primary for text and icons displayed on primaryContainer background in the launcher

Silzinc:
## Description

With my current background, the theme generated makes the selected mode button unreadable in the DMS launcher. I changed a few colors to solve this.

This could be a matter of taste and still not work for all generated themes, I just thought that it made more sense to have `primaryText` rather than `primary` on top of `primaryContainer`.

There are other places in the code where `primary` is used on `primaryContainer`, but I could not find the corresponding elements in the DMS UI so I made no change I could not test.

## Type of change

- [ ] Documentation
- [x] Other

## Screenshots / video

https://github.com/user-attachments/assets/6b6f323e-ac09-49d4-8694-e3ab2bce1c12

https://github.com/user-attachments/assets/975bc04d-160e-41f0-866e-762aa50132e7

## Checklist

- [N/A] New user-facing strings are wrapped in `I18n.tr()` with translator context, reusing existing terms where possible
- [N/A] Go changes: ran `make fmt`, added/updated tests, `make test` passes, and `go mod tidy` is clean
- [N/A] I have opened a corresponding pull request in dlx-docs to document any new behaviors: https://github.com/AvengeMedia/DankLinux-Docs

### DIFF
--- a/quickshell/Common/Theme.qml
+++ b/quickshell/Common/Theme.qml
@@ -449,7 +449,6 @@ Singleton {
                 "primary": getMatugenColor("primary", "#42a5f5"),
                 "primaryText": getMatugenColor("on_primary", "#ffffff"),
                 "primaryContainer": getMatugenColor("primary_container", "#1976d2"),
-                "onPrimaryContainer": getMatugenColor("on_primary_container", "#cfe5ff"),
                 "secondary": getMatugenColor("secondary", "#8ab4f8"),
                 "secondaryContainer": getMatugenColor("secondary_container", getMatugenColor("surface_container_high", "#292b2f")),
                 "tertiary": getMatugenColor("tertiary", "#efb8c8"),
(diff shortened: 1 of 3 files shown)


Purian23:
For me this does the opposite and takes it from visible to not. It's possible we may want to use `on_primary_container` here but we'll have to look into it a bit more so we don't cause other themes to be adversely affected.

Silzinc:
Sounds tougher than I thought... I have pushed a change for replacing `primary_text` with `on_primary_container`, if you are willing to test. I agree the latter should be preferred on `primary_container` background, but if it does not work, I would wonder if there actually was a solution.

Purian23:
Still looks off on my end, but I don't know why. Without this change as-is looks spot on to me. Let's see if @bbedward has insight when time permits.

PR #2587, opened 2026-06-06
feat(notifications): add opt-in timeout progress bar on popups

bogdan-velicu:
## What

Adds an opt-in **timeout progress bar** to notification popups — a thin bar pinned to the bottom of the card that drains full→empty as the auto-dismiss timer runs, giving a visual countdown to dismissal.

## Details

- New setting `notificationShowTimeoutBar` (**default off**), registered in `SettingsSpec.js` and surfaced as a toggle in **Settings → Notifications**.
- Plain `Rectangle` inset by the card's corner radius — no offscreen textures or shader passes.
- Shown for any timed notification (`timer.interval > 0`, including timed criticals); absent only when there's no timeout.
- Stays visible and frozen while hovered and during the popup exit animation.
- Reuses `notificationData.timer`; no new timers introduced.

## Demo

https://github.com/user-attachments/assets/d50efb4b-8950-4db3-a0b0-a02a5455d151

## Testing

Built quickshell from current `master` and ran the shell in an isolated nested session: verified the bar drains over the timeout, freezes on hover and during exit, shows on timed criticals, and is absent when there's no timeout. `qmllint` clean.

### DIFF
--- a/quickshell/Common/SettingsData.qml
+++ b/quickshell/Common/SettingsData.qml
@@ -698,6 +698,7 @@ Singleton {
     property int notificationTimeoutNormal: 5000
     property int notificationTimeoutCritical: 0
     property bool notificationCompactMode: false
+    property bool notificationShowTimeoutBar: false
     property bool notificationDedupeEnabled: true
     property int notificationPopupPosition: SettingsData.Position.Top
     property int notificationAnimationSpeed: SettingsData.AnimationSpeed.Short
--- a/quickshell/Common/settings/SettingsSpec.js
+++ b/quickshell/Common/settings/SettingsSpec.js
@@ -406,6 +406,7 @@ var SPEC = {
     notificationTimeoutNormal: { def: 5000 },
     notificationTimeoutCritical: { def: 0 },
     notificationCompactMode: { def: false },
+    notificationShowTimeoutBar: { def: false },
     notificationDedupeEnabled: { def: true },
     notificationPopupPosition: { def: 0 },
     notificationAnimationSpeed: { def: 1 },
--- a/quickshell/Modules/Notifications/Popup/NotificationPopup.qml
+++ b/quickshell/Modules/Notifications/Popup/NotificationPopup.qml
@@ -727,6 +727,51 @@ PanelWindow {
                 }
             }
 
+            // Timeout progress bar: drains as the dismiss timer runs; inset by
+            // the corner radius and frozen while hovered or during exit.
+            Rectangle {
+                id: timeoutBar
+
+                readonly property bool active: SettingsData.notificationShowTimeoutBar && notificationData && notificationData.timer && notificationData.timer.interval > 0
+                property real progress: 1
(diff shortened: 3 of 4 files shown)


purian23 (changes requested):
Thanks for the PR, this looks like a cool option! 

Let's address a few items here:

1. Replace the full-card masked layer with an inset, pill-shaped progress bar using ordinary `Rectangle` opts. This avoids two offscreen textures and a shader pass per popup while working cleanly with shadows and both frame modes. Performance is the main concern. 
2. Default `notificationShowTimeoutBar` to `false` in both `SettingsData.qml` and `SettingsSpec.js`. The spec entry is required for the toggle to work and persist and become an opt in user option. 
3. Base visibility on `timer.interval > 0`, rather than `!isPersistent`, so timed critical notifications are included.
4. Keep the bar visible but frozen during the popup exit animation.
5. Refactor to clean up the line comments to high level to reduce the footprint. 

No extra dedupe listener is needed: QML `Timer.restart()` already toggles `running` off and back on, restarting the bound animation.

bogdan-velicu:
Addressed all points in the latest push:

1. Replaced the masked layer with a plain inset `Rectangle` — no offscreen textures or shader passes; works with shadows and both frame modes.
2. Defaulted `notificationShowTimeoutBar` to `false` and added the `SettingsSpec.js` entry.
3. Gated visibility on `timer.interval > 0` instead of `!isPersistent`, so timed criticals are included.
4. The bar now stays visible and frozen during the exit animation (dropped the `popup` gate; the animation freezes via `!exiting`).
5. Trimmed the comments down.

Confirmed there's no need for a dedupe listener — `Timer.restart()` toggles `running` off/on and restarts the bound animation, as you noted.

Purian23:
Nice work on the revision , thanks!

Just two small things:

1. **Dedupe restart**  `timer.restart()` while already running won't reset the bar (binding stays `true`). A `Connections` on the timer to set `progress = 1` and `progressAnim.restart()` when `running` flips back on should do it. 
(Turns out my "no listener needed" note was off for dedupe)

2. **Null guard on `running`** Something like:
   `timeoutBar.active && notificationData && notificationData.timer && notificationData.timer.running && !win.exiting`

bogdan-velicu:
Pushed both:

1. Added a `Connections` on the timer that sets `progress = 1` and calls `progressAnim.restart()` whenever `running` flips back true — this catches the in-place restart on a deduped notification, where `running` stays true and the bound animation alone wouldn't re-fire. (Good call — my earlier "no listener" note was wrong here.)
2. Null-guarded the `running` expression: `timeoutBar.active && notificationData && notificationData.timer && notificationData.timer.running && !win.exiting`.

One side observation while testing: since un-hover calls `timer.restart()` (a fresh full interval), the bar resets to full rather than resuming — which is faithful to the real dismissal timer. If true pause/resume on hover is ever something you'd want, that's a separate change to the dismissal behavior (track remaining time instead of `restart()`), and it'd apply to all notifications. Out of scope here, but happy to take a look if you're interested.

Purian23:
Thanks!

PR #2598, opened 2026-06-07
fix(powermenu): close control center on lock and power actions

jbwfu:
## Description

When the power menu is opened from the Control Center, actions like lock and suspend currently leave the Control Center open after the action is triggered. This feels unintuitive and creates inconsistent UI feedback after the user has already confirmed a session or power action. This PR closes the Control Center alongside those actions so the flow behaves more cleanly and predictably.

## Type of change

- [ ] Documentation
- [x] Other

## Checklist


Purian23:
Thanks! Some may say this is a non fix as they could do both before, but we'll see how it goes :)

Issue #1659, opened 2026-02-13
Hot reload settings and plugin settings with IPC command


Silzinc:
### Feature Description

Hello! 

I have seen in the docs that the command `settings set [key] [value]` only works for non object entries, which is related to a limitation in the way [quickshell parses arguments in IPC calls](https://github.com/quickshell-mirror/quickshell/pull/22). I have seen in [DMSShellIPC.qml](https://github.com/AvengeMedia/DankMaterialShell/blob/master/quickshell/DMSShellIPC.qml) that devs are not sure as to whether users should be able to put whatever string they want in the command or not, without validation.

A way to circumvent this without having to answer that tedious question could be to add an IPC command to reload settings as it is done at startup.

### Use Case

I originally wanted a way to change my widget transparency as I switched from dark to light theme using shell hooks. Right now, modifying that little setting is impossible (from what I understood) without a full `dms restart` which I would like to avoid. 

With this feature, I can just call `sed` on `.config/DankMaterialShell/settings.json` and `dms ipc settings reload`.

### Compositor(s)

- [x] All compositors
- [ ] Niri
- [ ] Hyprland
- [ ] MangoWC (dwl)
- [ ] Sway
- [ ] Other (specify below)

### Proposed Solution

This amounts to adding this to the settings IPC commands:
```qml
function reload(): string {             
    SettingsData.settingsFile.reload(); 
    SettingsData.loadSettings();        
    return "SETTINGS_RELOAD_SUCCESS";   
}                                       
```

Of course, this comes with several issues, and since I am not a QML wizard I would like to know your opinion on this:
- The function always returns success even if there is a reading/parsing error, although the latter is still visible in the quickshell logs. I did not take the initiative to modify the logic in settings loading to also return the eventual error string, but it could be a possibility I guess.
- This goes against the recommendation in the [quickshell docs](https://quickshell.org/docs/v0.2.0/types/Quickshell.Io/FileView/#blockLoading) regarding the use of `blockLoading` in `FileView`, as this reload does not happen at startup (although in my tests the reload is pretty much instant, so perhaps the risk is low).
- This is just dodging the problem, as there still is no real way to check if granular modifications are correct.
- I have not read thoroughly the content of `loadSettings`. Perhaps there is something that should not be called again after startup?

Thanks in advance for considering this!

bbedward:
We could possibility just flip on `watchChanges` for the settings file

Silzinc:
[...]

PR #2611, opened 2026-06-10
feat(plugins): expose scan/rescan/reload IPC handlers for runtime plugin discovery

louzt:
Follow-up to #1659. Replaces the earlier "scan-ipc" attempt
(currently on `louzt:fix/plugin-scan-ipc-handler`, not yet opened
upstream) after a local pre-flight review against the QML coding
guidelines surfaced seven issues that would block upstream adoption.

## Summary

#1659 landed hot-reload for `settings.json` via `FileView.watchChanges` +
a 1ms `Timer` to skirt the JSON parse race. It covers settings changes
for already-loaded plugins. It does not cover **plugin discovery at
runtime**: adding a new plugin directory to
`~/.config/DankMaterialShell/plugins/` while the shell is running is
not consistently picked up by the existing `FolderListModel` watcher
in `PluginService.qml`, and there was no IPC handle for forcing a
rescan from outside the shell.

This PR adds an `IpcHandler` on `PluginService` with five small
functions: `scan`, `rescan`, `reload`, `list`, `status`. Together they
let a developer add, edit, and iterate on a plugin without restarting
the shell. Scope intentionally small: no file-watcher changes, no new
daemons, no schema additions, no new public API surface outside the
existing IPC mechanism.

## Why this shape

I saw the implementation in #1659 (Silzinc's `watchChanges + Timer`
pattern for settings) and tested whether the existing `userWatcher`
`FolderListModel { onCountChanged: resyncDebounce.restart() }` block
at `PluginService.qml:72` also covered new plugin dirs. Repro:

- shell running ~83h (`qs list --all` shows uptime)
- `mv /tmp/MyPlugin/ ~/.config/DankMaterialShell/plugins/`
- wait > 3s (debounce is 120ms + slack)
- `qs log --pid <PID> | grep MyPlugin` → empty, no `Plugin loaded:`
  line and no `invalid manifest fields` error

The plugin only appeared after a full shell restart. `qs ipc
--pid <PID> show` returned `QLocalSocket::PeerClosedError`, so there
was no externally callable way to force `scanPlugins()`. The
`IpcHandler` here closes that gap by exposing the three existing
internal functions (`scanPlugins`, `forceRescanPlugin`, `reloadPlugin`)
over the standard `qs ipc` channel, and adds two read-only helpers
(`list`, `status`) so wrapper scripts can poll cheaply.

Other shapes considered:

- Extending the existing `FolderListModel` watcher: hard to make
  portable across filesystems (`mv` / `cp -r` / `rsync --link-dest`
  semantics differ). Risk of behavior regressions for users who never
  hit this.
- A new file-watcher daemon: out of scope; overlaps with what
  `FolderListModel` already does for the common case.
- Forcing a full `resyncAll()` on every `Component.onCompleted`:
  doesn't help once the shell is running.
- Renaming the existing target: rejected — keeps the change additive.

An IPC handle is the smallest change that covers the runtime case
without touching the working watcher path. The handler is purely
additive.

## What changed

`quickshell/Services/PluginService.qml` — new `IpcHandler { target:
"plugin-scan" }` block at the end of the `Singleton`, with five
functions:

- `scan()` — calls existing `scanPlugins()`, returns count snapshot
  before debounce. **Fire-and-forget**: the resyncDebounce coalesces
  for 120ms after the call; callers that want to read post-state must
  poll `list` / `status` or wait ~200ms.
- `rescan(pluginId)` — calls existing `forceRescanPlugin(id)`,
  validates id against a safe charset first
- `reload(pluginId)` — calls existing `reloadPlugin(id)`, validates id
  first
- `list()` — returns `\n`-joined
  `id\tloaded|unloaded\ttype\tname` rows, capped at 256 entries with
  a `# count=N returned=M` header line for truncation detection
- `status(pluginId)` — returns
  `loaded|unloaded\ttype\terror` for a single plugin

No other files changed. `Quickshell.Io` was already imported.

## Review-driven fixes in this iteration

This PR incorporates seven findings from a local pre-flight review
of the earlier in-fork "scan-ipc" attempt. The earlier version
functioned but had issues that would block upstream adoption; this
version addresses all of them.

| # | Severity | Issue | Fix in this PR |
|---|---|---|---|
| B1 | blocker | `target: "plugins"` collided with the existing `IpcHandler` in `DMSShellIPC.qml:1180` (which registers `enable`/`disable`/`toggle`/`list`/`status` under `"plugins"`). One would shadow the other. | Renamed to `target: "plugin-scan"`. Both APIs stay discoverable; no shadowing. |
| H1 | high | `scan()` was documented as synchronous when the underlying `resyncDebounce` (120ms) makes the change observable only after a delay. | Documented as fire-and-forget; added the "poll `list`/`status` or wait ~200ms" instruction. |
| H2 | high | `rescan()` captured the `availablePlugins[id]` object and passed it to `forceRescanPlugin`. A parallel `_onManifestParsed` tick could mutate the entry between the read and the use (TOCTOU). | Drop the object capture; the handler passes the id string and `forceRescanPlugin` re-reads the entry. |
| M1 | medium | `list()` had no cap. A hostile or buggy plugin mass-creating entries could force ~80 KB per IPC call. | 256-entry cap with `# count=N returned=M` header. |
| M2 | medium | `status()` returned `"unknown\t\t"` for an unknown id, breaking the TSV contract. | Returns `ERROR: unknown pluginId '<id>'` (matches the rest of the handlers). |
| M3 | medium | No input validation. `__proto__` / `constructor` ids would reach `availablePlugins` / `pluginLoadErrors` lookups; a newline in the id would break TSV. | `^[a-zA-Z0-9_\-:]{1,64}$` regex on every `pluginId` argument; `\t` / `\n` / `\r` sanitized in `name` and `error` fields. |

The earlier version of the PR body stated "the target string does
not collide" — that was wrong, and B1 is the consequence. This PR
body reflects the actual, verified state of `DMSShellIPC.qml`.

## Compatibility Note

- This PR only adds an `IpcHandler` block. It does not remove or
  rename any existing function, signal, or property. Existing internal
  callers of `scanPlugins`, `forceRescanPlugin`, and `reloadPlugin`
  are unaffected.
- The new target is `"plugin-scan"`, which does not collide with any
  target in `DMSShellIPC.qml` (verified the 10 targets registered
  there: `defaultApp`, `powermenu`, `processlist`, `control-center`,
  `screenshot`, `dash`, `notepad`, `inhibit`, `mpris`, `keybinds`,
  plus `"plugins"` itself in `DMSShellIPC.qml:1180`).
- For users who were calling `qs ipc call plugins scan` against the
  earlier in-fork port: the target name change is the only breaking
  thing, and this PR is the first time the new target lands upstream,
  so no public callers exist yet.
- No settings schema additions. No new public Go/QML types. No
  manifest schema additions.

## Scope Boundary

- No changes to `userWatcher` / `systemWatcher` / `FolderListModel`
  behavior. The runtime detection gap that motivated this PR is real
  but its root cause (filesystem / debounce / inotify edge cases) is
  a separate investigation; this PR only provides an out-of-band way
  to recover from it.
- No new file watchers, no new daemons, no schema migration.
- No changes to the manifest validator. Plugins that still emit
  `PluginService: invalid manifest fields` keep doing so.
- No changes to settings hot-reload (the path that #1659 already
  addressed).
- No async / promise / GraphQL plumbing — every function is
  synchronous and returns a short string suitable for `qs ipc call`.
- This PR does **not** expose the "force-reload via filesystem
  watcher" path that some hardening forks (e.g.
  `hardening/notification-suite`) carry; that path depends on
  per-plugin `qmlcache` purging, which is plugin-specific and not
  appropriate for upstream.

## Validation

Local environment: Quickshell 0.3.0 (Debian package), running `dms`
for 83h+ uptime at the time of testing.

- `qs ipc --pid <PID> call plugin-scan list` returns one row per known
  plugin (verified `Object.keys(availablePlugins).length` matches
  row count).
- `qs ipc --pid <PID> call plugin-scan scan` returns
  `SCAN_TRIGGERED: <N> known before debounce`, and a follow-up
  `qs log --pid <PID> | tail -n 50` shows the manifest reload entries
  from the existing `resyncAll()` path.
- `qs ipc --pid <PID> call plugin-scan rescan myPlugin` with a known
  id returns `RESCAN_TRIGGERED: myPlugin`; with an unknown id returns
  `ERROR: unknown pluginId 'myPlugin' (try 'list' first)`.
- `qs ipc --pid <PID> call plugin-scan reload myPlugin` returns
  `RELOAD_TRIGGERED: myPlugin` and the shell logs `Plugin loaded:
  myPlugin` afterward.
- `qs ipc --pid <PID> call plugin-scan status myPlugin` returns
  `loaded\twidget\t` when healthy, or `loaded\twidget\t<error>` when
  the plugin has an entry in `pluginLoadErrors`.
- Empty-argument paths (`rescan ""`, `reload ""`, `status ""`) all
  return `ERROR: <function> requires a pluginId`.
- Reject paths:
  - `plugin-scan rescan ../../etc/passwd` →
    `ERROR: invalid pluginId '../../etc/passwd' (allowed:
    [a-zA-Z0-9_\-:]{1,64})`
  - `plugin-scan rescan __proto__` →
    `ERROR: invalid pluginId '__proto__' (allowed:
    [a-zA-Z0-9_\-:]{1,64})`
  - `plugin-scan rescan foo\tbar` →
    `ERROR: invalid pluginId 'foo\tbar' (allowed:
    [a-zA-Z0-9_\-:]{1,64})` (the `\t` is preserved in the error
    string for clarity, but never reaches any lookup or filesystem
    call)
- Brace balance check on the modified file: 252/252.
- `git merge-tree` against `upstream/master` is clean.
- `make lint-qml` is the canonical CI check; deferred to CI for the
  same reason as in the earlier body — local `qmllint6` (Qt 6) is not
  installed in this environment.

## Follow-ups (separate)

Not part of this PR; tracked here so it does not get bundled in:

- Investigate why `userWatcher.onCountChanged` does not fire for
  `mv`-based plugin directory additions. May be a filesystem-specific
  inotify behavior; needs a small repro matrix (ext4 / btrfs / tmpfs
  / network mount) before any code change.
- Optional: expose `scanPlugins` / `rescan` / `reload` as buttons in
  `PluginsTab.qml` so non-CLI users have the same recovery path.
- Optional: a follow-up `dlx-docs` PR (in
  `AvengeMedia/DankLinux-Docs`) documenting the new IPC target
  function-by-function. Held for now because the target is a
  developer-facing API rather than a user-facing behavior change.

## DMS PR-template checklist status

Per `AvengeMedia/DankMaterialShell/.github/pull_request_template.md`:

- [x] **Description** — this PR body
- [x] **Type of change: New feature, Refactor / internal cleanup** —
  exposes 5 new IPC functions; no behavior change for users who don't
  use them
- [x] **Related issues: anchor to #1659** (settings hot-reload
  merged; this is the plugin-discovery extension)
- [ ] **Screenshots / video** — N/A (no visual change)
- [x] **Code follows conventions in CONTRIBUTING.md** —
  `quickshell/Services/PluginService.qml` uses the established QML
  Singleton pattern with `IpcHandler` mirroring `DMSShellIPC.qml`;
  imports unchanged; `Quickshell.Io` was already imported
- [x] **Tested locally** — 9 `qs ipc call` commands documented under
  "Validation" above (4 happy paths, 2 unknown-id paths, 3 reject
  paths); ran live against the fork's `dms` instance.
- [x] **New user-facing strings wrapped in `I18n.tr()`** — N/A; the
  strings returned by this PR (`SCAN_TRIGGERED: ...`,
  `RESCAN_TRIGGERED: ...`, etc.) are IPC return values for external
  shell scripts, not user-facing UI labels. They intentionally
  follow the no-translation convention used by `DMSShellIPC.qml`
  targets (compare `DEFAULTAPP_LAUNCH_REQUESTED: <name>` etc.)
- [x] **Go changes** — N/A; this PR is QML-only
- [x] **QML changes: ran `make lint-qml` with no new warnings** —
  **deferred to CI**: local `qmllint6` (Qt 6) is not installed in
  this environment, and the local Quickshell tooling VFS
  (`.qmlls.ini` with `buildDir`) is locked by the live `dms`
  instance. The 3 entrypoints linted by
  `quickshell/scripts/qmllint-entrypoints.sh` are `shell.qml`,
  `DMSShell.qml`, `DMSGreeter.qml`; none of them instantiate
  `PluginService` directly but they all reference `qs.Services`
  which the singleton contributes to, so any regression surfaces
  there. CI is the canonical execution path.
- [x] **Companion PR in `dlx-docs` to document the new behavior** —
  **decided not to file** for this PR. The new IPC target is a
  developer-facing API (CLI / wrapper scripts) rather than a
  user-facing behavior change, and the function reference is short
  enough to fit in `qs ipc call --pid <PID> show` once this lands.
  If user feedback shows this is wrong, a follow-up PR against
  `AvengeMedia/DankLinux-Docs` can be opened with the function list.

### DIFF
--- a/quickshell/Services/PluginService.qml
+++ b/quickshell/Services/PluginService.qml
@@ -966,4 +966,65 @@ Singleton {
         }
         return result;
     }
+
+    IpcHandler {
+        target: "plugins"
+
+        // Re-runs the discovery pass over both user and system plugin
+        // directories. Useful when a plugin directory was added or removed
+        // at runtime and the FolderListModel watcher did not pick the
+        // change up.
+        function scan(): string {
+            root.scanPlugins();
+            return `SCAN_TRIGGERED: ${Object.keys(root.availablePlugins).length} known before debounce`;
+        }
+
+        // Re-reads a single plugin's manifest without restarting the
+        // shell. Picks up edits to plugin.json made after the plugin was
+        // already loaded once.
+        function rescan(pluginId: string): string {
+            if (!pluginId)
+                return "ERROR: rescan requires a pluginId";
+            const plugin = root.availablePlugins[pluginId];
+            if (!plugin)
+                return `ERROR: unknown pluginId '${pluginId}' (try 'list' first)`;
+            root.forceRescanPlugin(pluginId);
+            return `RESCAN_TRIGGERED: ${pluginId}`;
+        }
+
+        // Unloads and reloads a plugin in place. Lets a developer iterate
+        // on a plugin's QML without restarting the shell.
+        function reload(pluginId: string): string {
+            if (!pluginId)
+                return "ERROR: reload requires a pluginId";
+            if (!(pluginId in root.availablePlugins))
+                return `ERROR: unknown pluginId '${pluginId}'`;
+            root.reloadPlugin(pluginId);
+            return `RELOAD_TRIGGERED: ${pluginId}`;
+        }
+
+        // Returns one `<id>\t<loaded>\t<type>\t<name>` line per known
+        // plugin. Format is intentionally simple for easy parsing by
+        // external shell scripts or CLI management tools.
+        function list(): string {
+            const lines = [];
+            for (const id in root.availablePlugins) {
+                const p = root.availablePlugins[id];
+                lines.push(`${id}\t${p.loaded ? "loaded" : "unloaded"}\t${p.type || "unknown"}\t${p.name || ""}`);
+            }
+            return lines.join("\n");
+        }
+
+        // Returns `loaded|unloaded\ttype\terror` for a single plugin. Used
+        // by wrappers that want to poll after a `scan` or `reload`.
+        function status(pluginId: string): string {
+            if (!pluginId)
+                return "ERROR: status requires a pluginId";
+            const plugin = root.availablePlugins[pluginId];
+            if (!plugin)
+                return `unknown\t\t`;
+            const err = root.pluginLoadErrors[pluginId] || "";
+            return `${plugin.loaded ? "loaded" : "unloaded"}\t${plugin.type || ""}\t${err}`;
+        }
+    }
 }
--- a/quickshell/Services/PluginService.qml
+++ b/quickshell/Services/PluginService.qml
@@ -967,13 +967,29 @@ Singleton {
         return result;
     }
 
+    // Plugin id validation regex: lowercase / digits / underscore / dash / colon, 1-64 chars.
+    // Used by IPC handlers to reject shell-injection / prototype-pollution payloads
+    // before they reach the plugin manifest filesystem lookup. QML JS regex literal.
+    readonly property string _ipcIdPattern: "^[a-zA-Z0-9_\\-:]{1,64}$";
+
     IpcHandler {
-        target: "plugins"
+        // target renamed from "plugins" to "plugin-scan" to avoid collision
+        // with the existing IpcHandler in DMSShellIPC.qml:1180 which already
+        // registers `enable`/`disable`/`toggle`/`list`/`status` under "plugins".
+        // The split keeps both APIs discoverable without one shadowing the other.
+        target: "plugin-scan"
 
         // Re-runs the discovery pass over both user and system plugin
         // directories. Useful when a plugin directory was added or removed
         // at runtime and the FolderListModel watcher did not pick the
         // change up.
+        //
+        // Fire-and-forget: the resyncDebounce coalesces
+        // resyncAll() for 120ms after the call. Scripts that want to read
+        // state after a scan must poll `list` / `status` until the count
+        // changes, or wait at least 200ms. The pre-debounce count is
+        // returned as a sanity check (must be > 0 for any plugin to
+        // exist at all).
         function scan(): string {
             root.scanPlugins();
             return `SCAN_TRIGGERED: ${Object.keys(root.availablePlugins).length} known before debounce`;
@@ -985,8 +1001,15 @@ Singleton {
         function rescan(pluginId: string): string {
             if (!pluginId)
                 return "ERROR: rescan requires a pluginId";
-            const plugin = root.availablePlugins[pluginId];
-            if (!plugin)
+            // Sanitize id to a safe charset. Rejects shell-injection
+            // and prototype-pollution payloads (`__proto__`, `constructor`).
+            if (!new RegExp(root._ipcIdPattern).test(pluginId))
+                return `ERROR: invalid pluginId '${pluginId}' (allowed: [a-zA-Z0-9_\\-:]{1,64})`;
+            // Do NOT capture the plugin object — re-read inside
+            // forceRescanPlugin instead. _onManifestParsed can mutate
+            // availablePlugins from a parallel resyncDebounce tick, so a
+            // captured ref would see stale manifestPath.
+            if (!(pluginId in root.availablePlugins))
                 return `ERROR: unknown pluginId '${pluginId}' (try 'list' first)`;
             root.forceRescanPlugin(pluginId);
             return `RESCAN_TRIGGERED: ${pluginId}`;
@@ -997,6 +1020,9 @@ Singleton {
         function reload(pluginId: string): string {
             if (!pluginId)
                 return "ERROR: reload requires a pluginId";
+            // Same id sanitization as rescan.
+            if (!new RegExp(root._ipcIdPattern).test(pluginId))
+                return `ERROR: invalid pluginId '${pluginId}' (allowed: [a-zA-Z0-9_\\-:]{1,64})`;
             if (!(pluginId in root.availablePlugins))
                 return `ERROR: unknown pluginId '${pluginId}'`;
             root.reloadPlugin(pluginId);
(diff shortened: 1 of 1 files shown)


quickshell/Services/PluginService.qml
```diff
@@ -966,4 +966,115 @@ Singleton {
...
+
+        // Returns `loaded|unloaded\ttype\terror` for a single plugin. Used
+        // by wrappers that want to poll after a `scan` or `reload`.
+        // Use ERROR: prefix for unknown id (matches the rest of
+        // the handlers). Empty trailing field indicates "no error".
+        function status(pluginId: string): string {
+            if (!pluginId)
+                return "ERROR: status requires a pluginId";
+            // Sanitize id.
+            if (!new RegExp(root._ipcIdPattern).test(pluginId))
```
bbedward, line 1022: Is the regex validation necessary?

bbedward (changes requested):

bbedward:
Also why not use the existing IPC handlers for plugins? 
image

Purian23:
Also let's address the amount of line comments in the code, there's not a need for all of them to exist.

quickshell/Services/PluginService.qml
louzt: Yes, the regex is the cheapest defense and I want to keep it. Without it:

```
$ qs ipc call plugin-scan rescan __proto__
RESCAN_TRIGGERED: __proto__
```

would succeed and re-read `availablePlugins['__proto__']`, which is `Object.prototype` — letting the handler touch properties on every object in the QML JS scope. Same for `constructor`, `hasOwnProperty`, etc. The regex `^[a-zA-Z0-9_\\-:]{1,64}$` is the minimum cost (compiled once into `_ipcIdPattern`, called `new RegExp(...).test(...)` per call) and rejects those payloads at the gate.

Also rejects shell-injection payloads like `foo\t; rm -rf /` before they ever reach the filesystem lookup. The validation section in the PR body lists the reject paths; happy to add a runtime test if it helps.

Pushed a comment-strip refactor (`2c6a48a`) addressing Purian23's note.

louzt:
Good catch, i had the same concern when I started. Two different APIs, not duplicates:

- `dms ipc call plugins {enable,disable,toggle,list,status}` is registered in `DMSShellIPC.qml:1180`. It manipulates the **enabled/loaded state** of plugins in the user's plugin set (write-side; modifies config). `list` returns enabled plugins only.
- `dms ipc call plugin-scan {scan,rescan,reload,list,status}` (this PR) is on the new target `plugin-scan`. It manipulates **discovery** (read-mostly; `scan`/`rescan`/`reload` call existing internal functions `scanPlugins`/`forceRescanPlugin`/`reloadPlugin`). `list` returns **known** plugins (loaded + unloaded).

The rename from `plugins` to `plugin-scan` is exactly to avoid the collision, see the 'Compatibility Note' section. Both APIs stay discoverable. `list` returning different sets from different targets is by design; the maintainer can decide whether to consolidate in a follow-up. I can do that right now if u prefer.

If you'd rather have me extend the existing `plugins` target with these 3 functions and skip the new one entirely, I can do that, but `list` would need a flag to distinguish 'enabled' from 'known', and `scan`/`rescan`/`reload` don't fit the existing enable/disable/toggle naming. Let me know.

louzt:
Done — pushed `2c6a48a` which strips all 48 inline comment lines from the new IpcHandler block. The substantive rationale (regex necessity, target rename, fire-and-forget semantics, TOCTOU fix, 256 cap, ERROR prefix, id sanitization) now lives in the PR body under 'Review-driven fixes in this iteration' and 'What changed' where you reviewers can review it better. Code is bare logic now; brace count 252/252, no behavioral change.

Purian23:
Thanks!

PR #2615, opened 2026-06-11
feat(media-control): support scroll and right-click on audio output devices in media popout

hthienloc:
### What
Supports scroll wheel volume adjustment and right-click mute toggling for audio output devices in the media control popout:
1. Scrolling on the Audio Output Device icon or a specific device in the dropdown list adjusts its volume (step size 5%).
2. Right-clicking the Audio Output Device icon or a specific device in the dropdown list toggles its muted state without closing the popout.
3. Left-clicking a device in the dropdown list no longer closes the dropdown list.

### Why
Improves user experience by providing quick volume controls for secondary or main output devices directly within the media controls popout.

### How tested
Tested locally on quickshell. Verifying that:
- Scrolling adjusts volume and unmutes properly.
- Right click toggles mute/unmute and does not dismiss the popout.
- Left click switches default sink device without dismissing the dropdown.

### DIFF
--- a/quickshell/Common/SettingsData.qml
+++ b/quickshell/Common/SettingsData.qml
@@ -396,6 +396,7 @@ Singleton {
     property bool audioVisualizerEnabled: true
     property string audioScrollMode: "volume"
     property int audioWheelScrollAmount: 5
+    property bool audioDeviceScrollVolumeEnabled: false
     property bool clockCompactMode: false
     property int focusedWindowSize: 1
     property bool focusedWindowCompactMode: false
--- a/quickshell/Common/settings/SettingsSpec.js
+++ b/quickshell/Common/settings/SettingsSpec.js
@@ -156,6 +156,7 @@ var SPEC = {
     audioVisualizerEnabled: { def: true },
     audioScrollMode: { def: "volume" },
     audioWheelScrollAmount: { def: 5 },
+    audioDeviceScrollVolumeEnabled: { def: false },
     clockCompactMode: { def: false },
     focusedWindowCompactMode: { def: false },
     focusedWindowSize: { def: 1 },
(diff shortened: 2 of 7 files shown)


Purian23:
Hi @hthienloc, we like the idea of this, but don't think we can merge it at this time. What I think we'll run into is users complaining when they trying to scroll their list, it will scroll their volumes instead. This could be really frustrating with users who have many devices.

hthienloc:
I think it would make more sense to use scrolling on the output device icon to cycle through available output devices rather than adjust volume.

Something along these lines:

**Output device icon**

* Left click: quickly cycle through output devices (useful for users who frequently switch between 2–3 devices). Although this would overlap somewhat with the proposed scroll behavior.
* Right click: mute/unmute the current device.
* Middle click: I can't think of a useful action for this at the moment.
* Scroll: cycle up/down through the available output devices.

**Dropdown list**
I think the behavior proposed in this PR still makes sense for the device list itself. I often run into situations where I need to quickly adjust the volume of a secondary output device. Currently, I have to pause media playback, switch devices, adjust the volume, then switch back and resume playback. Being able to adjust volume directly from the list avoids that workflow entirely.

That said, I agree there are usability concerns around scroll interactions in a scrollable list. We should probably discuss the interaction model further before making any changes here.

hthienloc:
@Purian23 Let me know if the above plan aligns with your intentions so I can adjust the behavior accordingly.

Purian23:
Hey @hthienloc, yes I like the follow-up ideas there. I think we can have native scrolling on the icons and set an an updated option in the media player settings for the optional scroll volumes on the right half. That'll give us the best of both options. 

image

hthienloc:
image

Purian23:
@hthienloc thanks, this works well!

PR #2621, opened 2026-06-12
feat(settings): add clipboard entry action visibility controls

jbwfu:
## Description

Adds configurable visibility for clipboard entry action buttons.

Users can now choose which action buttons appear on clipboard entries from Settings -> Clipboard -> Behavior: Pin, Edit, and Delete. The default configuration preserves the existing behavior with all three buttons visible. When all action buttons are hidden, the entry content area and click-to-copy area expand to the right, reducing unused space.

This adds a new `clipboardVisibleEntryActions` setting and exposes multi-selection state from `SettingsButtonGroupRow`, so it can reuse the existing `DankButtonGroup` multi-select mode.

When the Pin action is hidden, a normal clipboard entry whose content matches an already pinned entry still shows a pinned-state indicator on the right. This keeps the saved/pinned status visible even when the action button itself is hidden.

Note: because the translation-related generated changes are relatively large, this PR does not update the translation files or settings search index for now.

## Type of change

- [ ] Documentation
- [ ] Other

## Screenshots / video

- Clipboard settings section showing the new visible entry actions control.

Screenshot from 2026-06-12 16-30-18

- Saved/pinned clipboard page after pinning an entry.

Screenshot from 2026-06-12 16-34-52

- Recents page showing a normal clipboard entry with the same content as a pinned entry, including the pinned indicator shown when the Pin action is hidden.

Screenshot from 2026-06-12 16-37-18

## Checklist

### DIFF
--- a/quickshell/Common/SettingsData.qml
+++ b/quickshell/Common/SettingsData.qml
@@ -108,6 +108,7 @@ Singleton {
     }
 
     property bool clipboardEnterToPaste: false
+    property var clipboardVisibleEntryActions: ["pin", "edit", "delete"]
 
     property var launcherPluginVisibility: ({})
 
--- a/quickshell/Common/settings/SettingsSpec.js
+++ b/quickshell/Common/settings/SettingsSpec.js
@@ -572,6 +572,7 @@ var SPEC = {
 
     builtInPluginSettings: { def: {} },
     clipboardEnterToPaste: { def: false },
+    clipboardVisibleEntryActions: { def: ["pin", "edit", "delete"] },
 
     launcherPluginVisibility: { def: {} },
     launcherPluginOrder: { def: [] },
--- a/quickshell/Modals/Clipboard/ClipboardEntry.qml
+++ b/quickshell/Modals/Clipboard/ClipboardEntry.qml
@@ -22,7 +22,14 @@ Rectangle {
     readonly property string entryType: modal ? modal.getEntryType(entry) : "text"
     readonly property string entryPreview: modal ? modal.getEntryPreview(entry) : ""
     readonly property var pinnedDuplicateEntry: !entry.pinned ? ClipboardService.getPinnedEntryByHash(entry.hash) : null
-    readonly property bool effectivePinned: entry.pinned || pinnedDuplicateEntry !== null
+    readonly property bool hasPinnedDuplicate: pinnedDuplicateEntry !== null
+    readonly property bool effectivePinned: entry.pinned || hasPinnedDuplicate
+    readonly property var visibleEntryActions: SettingsData.clipboardVisibleEntryActions || ["pin", "edit", "delete"]
+    readonly property bool showPinAction: visibleEntryActions.includes("pin")
+    readonly property bool showEditAction: visibleEntryActions.includes("edit")
+    readonly property bool showDeleteAction: visibleEntryActions.includes("delete")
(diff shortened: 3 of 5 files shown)


purian23 (changes requested):
Thanks, these are nice additions! 

One edge case I think is still missing: on the Saved tab, entries lose all pin/unpin UI when Pin is hidden.

Your description covers recents correctly where a normal entry matching a pinned one still shows the indicator. But on Saved, every entry has `entry.pinned === true`, so `pinnedDuplicateEntry` is always null and `showPinnedIndicator` never triggers. With Pin hidden, there’s no mouse way to unpin (keyboard Ctrl+S still works).

Suggested fix in` ClipboardEntry.qml`:
`readonly property bool showPinnedIndicator: !showPinAction && effectivePinned`
And update the indicator click handler:
```
onClicked: {
    if (entry.pinned) {
        unpinRequested(entry);
        return;
    }
    if (pinnedDuplicateEntry) {
        unpinRequested(pinnedDuplicateEntry);
    }
}
```
We appreciate all of your recent contributions!

jbwfu:
My original implementation was based on an intentional design assumption: when the user hides the Pin action, entries on the Saved tab no longer showing a pin/unpin button would be expected behavior.

The reasoning was:

- Users who choose to hide entry actions may be trying to reduce mouse-facing UI and rely more on keyboard actions.
- On the Recents tab, a normal entry whose content matches a saved/pinned entry still needs some state indication, because that entry itself is not the pinned entry.
- On the Saved tab, every entry is already pinned, and the header says "Clipboard Saved", so I thought the page context itself was enough to communicate that state.
- From the perspective of the "hide Pin action" setting, not showing a pin button on the Saved tab felt like a consistent behavior.

That said, your point is reasonable. Since I already added a clickable pinned indicator for duplicate saved entries on the Recents tab, the behavior is no longer simply "hide all Pin UI". In that light, the Saved tab having no mouse-accessible way to unpin does feel inconsistent.

I think there are two cleaner directions:

1. When the Pin action is hidden, duplicate saved entries on Recents should only show a non-clickable pinned status indicator.
2. Or, as you suggested, hide the normal Pin action but still show a pinned indicator for all effective pinned entries, and allow clicking it to unpin, including entries on the Saved tab.

Which behavior do you prefer? If the expected behavior is to keep pinned state visible and still provide a mouse-accessible unpin action when the Pin action is hidden, I can adjust this PR according to your suggestion.

Purian23:
I think option 2 will work best here as a follow up to each option.

jbwfu:
Updated according to your suggestion.

Purian23:
@jbwfu Now that I see it, I think having the buttons turn off the action icons should simply mean what they do, On/Off. This is now halfway there isn't it :sweat_smile:

jbwfu:
@Purian23  Just to confirm: do you mean Pin Off should fully hide the pin/unpin icon everywhere, including the Saved tab and duplicate entries in Recents?

Purian23:
As you had it, yes probably :|]

jbwfu:
Option 1 looks like this.

image

image

If this looks good to you, I'll submit a new follow-up PR.

Purian23:
Okay, just to be super duper clear lol. As usual, all 3 icons exist and are shown by default. But now, with the new button options to hide those action icons, they simply all hide. So they are either all on, or all off optionally by the user!

PR #2623, opened 2026-06-12
Fix/clipboard confirmation keyboard safety

jbwfu:
## Description

This PR improves keyboard safety around the clipboard clear-history confirmation flow.

The clear-history confirmation dialog now defaults keyboard selection to Cancel instead of the destructive Clear All action. This makes keyboard interaction safer for a destructive operation and keeps the behavior consistent with pressing the final-line selection, where the user is not pushed directly into confirming a destructive action.

This also makes the Shift+Delete clear-all shortcut use the same confirmation flow as the Clear All button. Previously, the button asked for confirmation, but the shortcut cleared clipboard history immediately. Both entry points now share the same confirmation behavior: confirming clears the relevant history and closes the clipboard, while canceling leaves it unchanged.

The confirmation dialog also handles Tab and Shift+Tab keyboard navigation correctly while open.

## Type of change

- [ ] Documentation
- [ ] Other

## Checklist

### DIFF
--- a/quickshell/Modals/Clipboard/ClipboardContent.qml
+++ b/quickshell/Modals/Clipboard/ClipboardContent.qml
@@ -7,7 +7,6 @@ Item {
     id: clipboardContent
 
     required property var modal
-    required property var clearConfirmDialog
 
     property alias searchField: searchField
     property alias clipboardListView: clipboardListView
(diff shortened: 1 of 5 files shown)


Purian23:
Thanks, reviewing each of the Clipboard PR's now!

PR #2634, opened 2026-06-14
feat: add battery settings tab and update battery widget interactions

### LINKED ISSUE #2518: Add Battery Saver Button
### Feature Description

I would love to have a Button to enable / disable Battry saveing. So my laptop only charges to 60%.

### Use Case

_No response_

### Compositor(s)

All compositors

### If Other, please specify

_No response_

Maybe something like this is possible https://extensions.gnome.org/extension/4612/battery-threshold/

hthienloc:
Closes #2518

### What
Add a new Battery tab in settings under Powers & Security, and update mouse scrolling/right-clicking behavior on the battery widget. Also add threshold notifications and auto power saver profile activation.

### Why
Provide a unified interface for battery configurations, saving battery hardware limits, adjusting screen brightness, and power management.

### How tested
Tested compilation of QML, settings sidebar navigation, and script execution locally.
dms_capture_1781404609947


hthienloc:
This PR somehow overlaps with the Dank Battery Alerts plugin, but I think this is something that should be built into DMS rather than being a separate plugin.

Purian23:
I tend to agree, thoughts @bbedward?

bbedward:
Will deprecate battery alerts plugin after 1.5 drops

toods2:
Can other of you confirm that after this commit *both* the new Battery tab under Power & Security and the Niri tab under Personalization try to render at the same time?

IMG_3242

DerDennisOP:
:heart:

PR #2652, opened 2026-06-16
feat(sounds): make muting sounds during media playback configurable

### LINKED ISSUE #2616: No more notification sound when playing audio
### Compositor

Niri

### Distribution

CachyOS

### If Other, please specify

_No response_

### Select your Installation Method

DankInstaller

### Was this your original Installation method?

Yes

### If no, specify

_No response_

### dms doctor -vC

<details>
<summary>Click to expand</summary>

</details>

### Description

I often listen to music, and I have little notifications sounds to warn me about notifications when I'm not looking at the screen.

Though recently, I saw there's been a commit (e3dbaed) that skips the notification is there's audio playing, which is not configurable at all.

### Expected Behavior

The existing behavior to not be changed, and at least to have a new option that permits to bring back the old behavior

### Steps to Reproduce

1. Enable notification sounds
2. Play music with a player compatible with mpris
3. Receive a notification -> no sound

### Error Messages/Logs

_No response_

### Screenshots/Recordings

_No response_

Rocho-EL-Locho:
Closes #2616

## Problem

Commit e3dbaed ("audio: disable effects when mpris player is playing") added an `isMediaPlaying()` check to every sound function in `AudioService.qml`, so all system sounds — notifications, volume, power, login — are skipped whenever an MPRIS player is playing. There is no way to opt out.

Users who rely on notification sounds while listening to music lost them entirely, with no setting to restore the previous behaviour.

## Fix

Add a `Mute During Playback` toggle backed by `SettingsData.muteSoundsWhenMediaPlaying`:

- **Default `true`**, so the current behaviour is preserved for everyone.
- When disabled, system sounds play again even while media is playing — the behaviour requested in the issue.

The media check is wrapped in a small helper so the gate is applied consistently across all sound functions:

```qml
function shouldMuteForMedia() {
    return SettingsData.muteSoundsWhenMediaPlaying && isMediaPlaying();
}
```

Each `play*Sound()` function now calls `shouldMuteForMedia()` instead of `isMediaPlaying()` directly.

## Changes

- `Common/SettingsData.qml` — new `muteSoundsWhenMediaPlaying` property (default `true`)
- `Common/settings/SettingsSpec.js` — persistence spec entry
- `Services/AudioService.qml` — `shouldMuteForMedia()` helper, used by all sound functions
- `Modules/Settings/SoundsTab.qml` — `Mute During Playback` toggle in the System Sounds card

## Testing

Verified live on niri:

| Toggle | Media playing | Notification sound |
|--------|---------------|--------------------|
| On (default) | yes | silent |
| Off | yes | plays |

## Note

`translations/settings_search_index.json` is left untouched. It is a generated file and is already out of sync with the current sources (regenerating it on a clean checkout produces an unrelated ~50-line diff), so regenerating here would mix in churn unrelated to this change. Happy to include a regenerated index if you prefer it handled in this PR.

### DIFF
--- a/quickshell/Common/SettingsData.qml
+++ b/quickshell/Common/SettingsData.qml
@@ -572,6 +572,7 @@ Singleton {
     property bool soundVolumeChanged: true
     property bool soundPluggedIn: true
     property bool soundLogin: false
+    property bool muteSoundsWhenMediaPlaying: true
 
     property int acMonitorTimeout: 0
     property int acLockTimeout: 0
--- a/quickshell/Common/settings/SettingsSpec.js
+++ b/quickshell/Common/settings/SettingsSpec.js
@@ -282,6 +282,7 @@ var SPEC = {
     soundNewNotification: { def: true },
     soundVolumeChanged: { def: true },
     soundPluggedIn: { def: true },
+    muteSoundsWhenMediaPlaying: { def: true },
 
     acMonitorTimeout: { def: 0 },
     acLockTimeout: { def: 0 },
--- a/quickshell/Modules/Settings/SoundsTab.qml
+++ b/quickshell/Modules/Settings/SoundsTab.qml
@@ -131,6 +131,23 @@ Item {
                         checked: SettingsData.soundPluggedIn
                         onToggled: checked => SettingsData.set("soundPluggedIn", checked)
                     }
+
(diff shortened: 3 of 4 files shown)


Purian23:
Thanks @Rocho-EL-Locho~!

PR #2668, opened 2026-06-18
Enhance/clipboard history interactions

### LINKED ISSUE #2635: Make "Enter to Paste" clipboard setting consistent when also left clicking entries in clipboard
### Feature Description

Make left clicking an entry in the Clipboard modal respect the setting for "Enter to Paste". If enter to paste is turned on, using arrow keys to navigate and hitting enter on an entry works as expected. But clicking the entry with the mouse still just copies that entry to the clipboard without pasting it.

### Use Case

More consistent UX when the enter to paste setting is toggled on.

### Compositor(s)

Niri

### If Other, please specify

_No response_

### Proposed Solution

Have left click behavior in the clipboard be consistent with the enter to paste setting. If the setting is on, clicking the entry should also paste the text.

_No response_

jbwfu:
## Description

This PR includes several clipboard interaction updates and fixes:

- Added a context menu for clipboard entries.
- Added an optional "Click to Paste" setting, decoupling mouse clicks from keyboard (`Enter`) behavior.
- Added optional "Copy" and "Paste" buttons for entries (disabled by default).
- Improved UX: Auto-select the first item, close immediately on `Esc`, and reset scroll on search.
- Fixed an issue where the search clear button overlapped with the filter button.

## Type of change

- [ ] Documentation
- [ ] Other

## Related issues

Closes #2635

## Screenshots / video

Screenshot from 2026-06-18 18-50-07

## Checklist

### DIFF
--- a/quickshell/Common/SettingsData.qml
+++ b/quickshell/Common/SettingsData.qml
@@ -107,6 +107,7 @@ Singleton {
         saveSettings();
     }
 
+    property bool clipboardClickToPaste: false
     property bool clipboardEnterToPaste: false
     property bool clipboardRememberTypeFilter: false
     property string clipboardTypeFilter: "all"
--- a/quickshell/Common/settings/SettingsSpec.js
+++ b/quickshell/Common/settings/SettingsSpec.js
@@ -597,6 +597,7 @@ var SPEC = {
     desktopWidgetGroups: { def: [] },
 
     builtInPluginSettings: { def: {} },
+    clipboardClickToPaste: { def: false },
     clipboardEnterToPaste: { def: false },
     clipboardRememberTypeFilter: { def: false },
     clipboardTypeFilter: { def: "all" },
--- a/quickshell/Modals/Clipboard/ClipboardContent.qml
+++ b/quickshell/Modals/Clipboard/ClipboardContent.qml
@@ -66,6 +66,8 @@ Item {
         contextMenu.hide();
     }
 
+    readonly property bool contextMenuActive: contextMenu.openState
+
     anchors.fill: parent
 
     ClipboardContextMenu {
--- a/quickshell/Modals/Clipboard/ClipboardContent.qml
+++ b/quickshell/Modals/Clipboard/ClipboardContent.qml
@@ -2,6 +2,7 @@ import QtQuick
 import Quickshell
 import qs.Common
 import qs.Widgets
+import qs.Services
 
 Item {
     id: clipboardContent
@@ -116,6 +117,12 @@ Item {
                 onTextChanged: {
                     modal.searchText = text;
                     modal.updateFilteredModel();
+                    ClipboardService.selectedIndex = 0;
+                    ClipboardService.keyboardNavigationActive = true;
+                    Qt.callLater(function () {
+                        clipboardListView.positionViewAtBeginning();
+                        savedListView.positionViewAtBeginning();
+                    });
                 }
 
                 Keys.onEscapePressed: function (event) {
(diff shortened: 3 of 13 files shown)


purian23 (changes requested):
Thanks for the PR, I like the context menu and the optional Copy button!

One thing we'd like to adjust: the new single-click-selects removes the long-standing single-click-to-copy behavior, I'd like that restored as I personally use it all the time. 

I see where you were going with it, keeping double click to paste will still allow the variant new addition to flow through.

jbwfu:
I think clicking to select and double-clicking to use (paste) is more intuitive.I added the copy button because I considered that users were used to using the click-to-copy function.This copy button is hidden by default to avoid taking up too much clipboard space.

If you insist on restoring the single-click copy function, the double-click function will not work, and the copy button will be meaningless. I will then remove these two commits.

Do you still hold your opinion?

jbwfu:
> the double-click function will not work

It doesn't seem to have any impact?If you insist on restoring the single-click copy function, I will keep the double-click paste function and remove the copy button.

bbedward:
I dont think it makes sense to have a mouse interaction select the item, it feels confusing to me. The mouse is the selector, and hovering it indicates which row is selected. For KB interactions it makes more sense of course.

It would probably make more sense to reveal explicit copy/paste buttons, when hovered at least, that can just be both selected with a left click.  That's just more intuitive for typical mouse-based interactions

Purian23:
> I dont think it makes sense to have a mouse interaction select the item, it feels confusing to me. The mouse is the selector, and hovering it indicates which row is selected. For KB interactions it makes more sense of course.
> 
> It would probably make more sense to reveal explicit copy/paste buttons, when hovered at least, that can just be both selected with a left click. That's just more intuitive for typical mouse-based interactions

Agreed, this was my take also. Good features to have, but not by default were my main concerns. Along with the above ideas, you could also make the left click to select another option in settings like the enter the paste. The best of both worlds. 

We appreciate every PR from all contributors, but we may not always agree with the implementation of each at presentation. We're always option to feedback and others perspective though at the same time.

jbwfu:
> I dont think it makes sense to have a mouse interaction select the item, it feels confusing to me. The mouse is the selector, and hovering it indicates which row is selected. For KB interactions it makes more sense of course.

Now I agree with this view.

> I see where you were going with it, keeping double click to paste will still allow the variant new addition to flow through.

However, in my actual testing, the double-click operation cannot be truly bound to the paste function; it seems to be overridden by the single-click copy behavior.

> Agreed, this was my take also. Good features to have, but not by default were my main concerns. Along with the above ideas, you could also make the left click to select another option in settings like the enter the paste. The best of both worlds.

So, what the current PR needs to do next is:

- Add settings for click behavior, similar to enter to paste.
- When the mouse hovers over a clipboard item, the copy and paste buttons will be forcibly displayed on the right. However, considering the number of buttons, the user's activation status of the buttons, and behaviors such as line breaks caused by temporary button changes, this effect may not be very good.

I will update when I have time.

Purian23:
I think I'd drop this one "When the mouse hovers over a clipboard item, the copy and paste buttons will be forcibly displayed on the right", and instead favor simply making the click selection an option in settings. The floaters in this sense I don't believe will work out as well either for space reasons.

jbwfu:
Applied the discussed changes, and additionally fixed the clear button overlapping with the filter button in the search bar.

purian23 (approved):
Nice work, thanks you! I added a small commit to remove the keyboard nav code on the context menu. It was staged but not fully wired. Decided to forego adding it since it's most mouse driven and keyboard users have native actions as-is.

Issue #324, opened 2025-10-05
Multi-language / Localization / Translation support?


kamikaze211:
## Feature Description

I'm a translator, and I have used linux for some years, I had translated some open source project, and I want to help with the translation, if you can add the multi-language support for DMS.

## Use Case

For that people don't use English for his/her main language.

## Compositor

Is this feature specific to one compositor?

- [ ] All compositors

## Proposed Solution

I'm not sure if this need the upstream ( quickshell ) support.

## Alternatives/Existing Solutions

I had searched for some keywords, seems have no idea about this.

bbedward:
I think we just need to wrap all user facing strings with qsTr() and we can generate localization files to translate.

bbedward:
I setup the initial structure for this, but probably have some terms in the index that shouldnt be translated.

What I do, or what I've done in the past is use [poeditor](https://poeditor.com/) 

Localization is really freaking annoying to get right, because of length considerations and context considerations - I'm not interested in spending that much time on all of that.

Anyway - I made the project public which I think, gives anyone the ability to request access to contribute or add a language:

https://poeditor.com/join/project/kyeCPTpmJM 

Once we have more than 1 language, i'll have to take the step to parse it from poeditor and use it.

Moraxyc:
@bbedward Looks like no languages have been added in POEditor yet, it said `There are no languages to translate in this localization project. For more information, consider getting in touch with those in charge of managing the project.`

bbedward:
You'll have to tell me which languages you want to contribute to and I'll add them.

Moraxyc:
> You'll have to tell me which languages you want to contribute to and I'll add them.

Oh, that’s great! I’d be happy to help with translating it into Chinese (zh_CN).

kamikaze211:
> You'll have to tell me which languages you want to contribute to and I'll add them.

Sorry for late responsing.
My main language is Chinese (Simplified), and my 2nd language is Japanese, so may you add these languages that I can try to translate?

bbedward:
I added chinese and japanese, I think you can put a request to contribute to them. (I'm not exactly sure, been a long time since I used poeditor)

Like I said some of the terms may be irrelevant, I did a blanket-sweep to wrap them so might have picked up irrelevant things.

kamikaze211:
> > You'll have to tell me which languages you want to contribute to and I'll add them.
> 
[...]

PR #2670, opened 2026-06-19
i18n: add Esperanto translation

JonathanReeve:
Adds the Esperanto (eo) translations for all UI terms, aligning with the current master branch. GNOME and Komputeko terminology were referenced for translations where applicable.


bbedward:
See https://github.com/AvengeMedia/DankMaterialShell/issues/324 - translations need to be done on poedit

PR #2678, opened 2026-06-21
feat(network): captive portal detection and sign-in popup

### LINKED ISSUE #2632: Captive Portal Integration
### Feature Description

It'd be nice to have a [captive portal](https://support.mozilla.org/en-US/kb/captive-portal) appear directly when connecting to a WiFi network that requires it. I'm proposing a pop-up window similar to the ones in mainstream OSes like iOS, macOS, and Android. This window could use the user's default browser or use a bundled browser engine (I imagine the latter is less likely).

### Use Case

This feature is useful when connecting to (often) public WiFi networks that use captive portals to allow you to use the Internet. 

Opening Firefox manually isn't an elegant solution. DMS will show me I'm connected to WiFi, but my terminal apps won't connect to the Internet, so I have to remember that I'm signed into a public network, turn off my VPN, and finally open Firefox. I think DMS could offer a better solution since it knows the system's WiFi status and VPN configuration, so it could automatically present such a pop-up.

### Compositor(s)

All compositors

### If Other, please specify

_No response_

### Proposed Solution

_No response_

_No response_

Rocho-EL-Locho:
## Description

Adds captive portal detection and a sign-in popup, similar to mainstream desktop and mobile OSes: when you join a network that requires sign-in before reaching the internet, DMS surfaces a popup and opens the login page in your default browser.

How it works:
- A backend-agnostic connectivity probe in the network manager requests a known endpoint with redirects disabled. A redirect (or an unexpected response body) means traffic is being intercepted by a portal; a `204`/expected body means full connectivity. The result is exposed as two new `NetworkState` fields, `connectivity` (`unknown`/`none`/`portal`/`limited`/`full`) and `portalURL`, which flow over the socket to the QML services.
- `CaptivePortalModal` auto-opens when a portal is detected (gated by a setting) and closes automatically once the user signs in (the periodic re-probe flips the state to `full`).
- A settings toggle under **Network Status → Captive Portal** enables/disables the popup, and a manual "Open sign-in page" entry is shown there while a portal is active.

Notes / design decisions (open to feedback):
- The probe lives in the `Manager` rather than the NetworkManager backend, so it works across all backends (NetworkManager, iwd, networkd) and does not depend on NM's connectivity checking being enabled.
- The probe endpoint defaults to `http://nmcheck.gnome.org/check_network_status.txt` (the same default NetworkManager uses) and is overridable via `DMS_CAPTIVE_PROBE_URL`. Once a connection is healthy, it is re-checked only every ~5 minutes to keep background traffic low.
- The modal reports the SSID and warns when a VPN is active (it may need to be disconnected to reach the portal).

## Type of change

- [ ] Documentation
- [ ] Other

## Related issues

Fixes #2632

## Screenshots / video

The popup ("Sign in to network") shows the network name with **Open login page** and **Dismiss** actions. I can attach a screenshot/recording with demo data on request.

## Checklist

- [x] I have tested my changes locally (against a real captive portal and a local simulation)
- [x] New user-facing strings are wrapped in `I18n.tr()` and registered in `en.json`, reusing existing terms where possible
- [ ] QML changes: `make lint-qml` was not run locally — generating the tooling VFS requires a newer Quickshell than the packaged 0.3.0 here. QML was verified with `qmllint` syntax checks and a live runtime load; happy to address anything CI flags.
- [ ] I have opened a corresponding pull request in DankLinux-Docs — glad to do so if you'd like docs for this.

### DIFF
--- a/core/internal/server/network/captive_portal.go
+++ b/core/internal/server/network/captive_portal.go
@@ -1,0 +1,185 @@
+package network
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/AvengeMedia/DankMaterialShell/core/internal/log"
+)
+
+const (
+	portalProbeURL       = "http://nmcheck.gnome.org/check_network_status.txt"
+	portalProbeExpect    = "NetworkManager is online"
+	portalProbeTimeout   = 5 * time.Second
+	portalProbeInterval  = 30 * time.Second
+	portalProbeMaxBody   = 4096
+	portalFullProbeTicks = 10 // re-probe a healthy connection every ~5min, not every tick
+)
+
+// portalProbe checks a known endpoint to spot a captive portal: a redirect or an
+// unexpected 200 body means traffic is being intercepted.
+type portalProbe struct {
+	mgr       *Manager
+	client    *http.Client
+	url       string
+	trigger   chan struct{}
+	stopChan  chan struct{}
+	wg        sync.WaitGroup
+	lastKey   string
+	fullTicks int
+}
+
+func newPortalProbe(m *Manager) *portalProbe {
+	probeURL := portalProbeURL
+	if v := os.Getenv("DMS_CAPTIVE_PROBE_URL"); v != "" {
+		probeURL = v
+	}
+	return &portalProbe{
+		mgr:      m,
+		url:      probeURL,
+		trigger:  make(chan struct{}, 1),
+		stopChan: make(chan struct{}),
+		client: &http.Client{
+			Timeout: portalProbeTimeout,
+			CheckRedirect: func(req *http.Request, via []*http.Request) error {
+				return http.ErrUseLastResponse
+			},
+		},
+	}
+}
+
+func (p *portalProbe) start() {
+	p.wg.Add(1)
+	go p.run()
+	p.kick()
+}
+
+func (p *portalProbe) stop() {
+	close(p.stopChan)
+	p.wg.Wait()
+}
+
+func (p *portalProbe) kick() {
+	select {
+	case p.trigger <- struct{}{}:
+	default:
+	}
+}
+
+func (p *portalProbe) run() {
+	defer p.wg.Done()
+	ticker := time.NewTicker(portalProbeInterval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-p.stopChan:
+			return
+		case <-p.trigger:
+			p.probe(false)
+		case <-ticker.C:
+			p.probe(true)
+		}
+	}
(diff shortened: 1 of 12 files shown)


quickshell/Modals/CaptivePortalModal.qml
```diff
@@ -0,0 +1,175 @@
+import QtQuick
+import qs.Common
+import qs.Modals.Common
+import qs.Services
+import qs.Widgets
+
+DankModal {
+    id: root
+
+    readonly property string fallbackURL: "http://nmcheck.gnome.org/check_network_status.txt"
```
bbedward, line 10: Probably better if we don't hardcode this in 2 places on th e QML side, can be a constant in NetworkService

core/internal/server/network/types.go
```diff
@@ -24,6 +24,16 @@ const (
 	PreferenceEthernet ConnectionPreference = "ethernet"
 )
 
+type Connectivity string
+
+const (
+	ConnectivityUnknown Connectivity = "unknown"
+	ConnectivityNone    Connectivity = "none"
+	ConnectivityPortal  Connectivity = "portal"
+	ConnectivityLimited Connectivity = "limited"
```
bbedward, line 33: Doesnt seem to do anything with the limited case, should we? I'd rather remove it or add a // TODO than leave the unused value

quickshell/Modals/CaptivePortalModal.qml
```diff
@@ -0,0 +1,175 @@
...
+                                id: dismissArea
+                                anchors.fill: parent
+                                hoverEnabled: true
+                                cursorShape: Qt.PointingHandCursor
+                                onClicked: root.close()
+                            }
+                        }
+
+                        Rectangle {
+                            width: Math.max(120, openText.contentWidth + Theme.spacingM * 2)
```
bbedward, line 130: Better to use standard Widgets/ than hand-rolling custom buttons, e.g. Dankbutton

core/internal/server/network/captive_portal.go
```diff
@@ -0,0 +1,185 @@
...
+func (p *portalProbe) kick() {
+	select {
+	case p.trigger <- struct{}{}:
+	default:
+	}
+}
+
+func (p *portalProbe) run() {
+	defer p.wg.Done()
+	ticker := time.NewTicker(portalProbeInterval)
```
bbedward, line 77: Not a fan of using a ticker here, better to rewrite polling to be event-driven. I'd re-work that as a timer thats conditionalled armed, cancelled when sign in is "full" . We can probably subscribe to NM PropertiesChanged, or something like that to catch portal changes.

bbedward (changes requested):

Spoch-dev:
Closing

PR #2687, opened 2026-06-23
feat(bar): separate workspace appearance for unfocused displays

b0o:
## Description

Adds the ability to style the Dank Bar workspace indicators differently on displays that aren't currently focused.

The **Workspace Appearance** card (Dank Bar settings) is now split into two tabs:

- **Focused Display** — the existing options, unchanged.
- **Unfocused Display(s)** — a master **"Separate Appearance for Unfocused Display(s)"** toggle (off by default). While off, unfocused displays inherit the Focused Display settings, so behavior is identical to before. While on, the tab exposes an independent mirror of the focused options: Focused/Occupied/Unfocused/Urgent colors, plus the Focused Border (enable, color, thickness).

At runtime, each `WorkspaceSwitcher` determines whether its output is the focused one and resolves every color, the border enable, and the border thickness from either the focused or the unfocused-monitor settings accordingly.

### Implementation notes

- New flat settings keys (`workspaceUnfocusedMonitor*`) added to `SettingsData.qml` and `SettingsSpec.js`; defaults mirror the focused settings so existing configs render unchanged.
- `WorkspaceSwitcher.qml` adds reactive `isFocusedMonitor` / `useUnfocusedAppearance` and routes color/border resolution through them.
- `settings_search_index.json` regenerated for the new settings.

## Type of change

## Screenshots / video

image

image

## Checklist

- [x] Go changes: ran `make fmt`, added/updated tests, `make test` passes, and `go mod tidy` is clean (N/A — no Go changes)


purian23 (changes requested):
Thanks for the PR, this is a nice addition! 

- **Focused-monitor lookup:** Let's consolidate `focusedMonitorName` into a shared helper (extend `BarWidgetService.getFocusedScreenName()` with Mango support) instead of keeping a separate compositor switch in `WorkspaceSwitcher`.
- **Settings UI duplication:** We can reduce over 100 lines by extracting the shared color/border options into a reusable component so the focused and unfocused tabs aren't maintaining two copies of the same layout.
- **Unsupported compositors (labwc):** Currently this silently no-ops via the empty-string fallback, we can handle it to either wire up focus detection if labwc supports it, or disable/hide the unfocused display options in that comp. 

Overall, everything works great, these updates will only optimize it a bit more.

Purian23:
Thanks @b0o for the updates! I didn't realize you had updated the code. It doesn't ping on us code pushes so make sure you comment or hit the request review button again. I will get this reviewed shortly.

purian23 (approved):

PR #2697, opened 2026-06-25
feat: Chromecast + AirPlay 2 casting

domenkozar:
## Summary

Adds a casting feature that discovers and casts to both **Chromecast** (`_googlecast._tcp`) and **AirPlay 2** (`_airplay._tcp`) devices from a single Control Center widget.

- **Chromecast**: media casting + screen mirroring (HLS), handled in-process by the `dms` core via the Go Cast protocol.
- **AirPlay 2**: screen mirroring via [doubletake](https://github.com/omarroth/doubletake) (a separate GPLv3 process, never linked into the MIT `dms`).
- Screen capture/encode is done by a new `dms-cast-helper` binary using the GStreamer library (go-gst) — not the `gst-launch`/`ffmpeg`/`wf-recorder` CLIs. The capture pipeline imports the portal's DMA-BUF via VA-API and forces 4:2:0 (`I420`) chroma (4:4:4 is rejected by most TV decoders → black screen).
- mDNS discovery prefers **avahi** (D-Bus API) when present and falls back to a built-in re-browsing browser otherwise.

The main `dms` binary stays CGO-free; the cast helper is opt-in and needs CGO + GStreamer dev packages. See `docs/casting.md` for build, firewall, and environment details.

## Dependency

AirPlay mirroring relies on the doubletake fork, which drives `gst-launch-1.0` for capture and fixes the black-screen + pts-timing issues on that path (DMA-BUF import + 4:2:0 chroma). doubletake builds CGO-free and only needs the `gst-launch-1.0` CLI plus plugins at runtime:

- **Depends on** omarroth/doubletake#9 — upstream PR carrying the capture + chroma fixes this feature needs.

## Potential blocker

- omarroth/doubletake#8 — may block or affect AirPlay mirroring until resolved.

## Status

Slices: discovery, cast, screen-mirror, Control Center widget/service, auto-reconnect. AirPlay 2 mirroring has been verified against a real non-Apple AirPlay 2 receiver; Chromecast paths are built but not yet verified against real Cast hardware.

Screenshot from 2026-06-24 11-42-07


bbedward:
CGO is a hard no from me, along with the GPL3 lib which isn't a potential blocker it is a blocker because we need to be able to distribute DMS with vendored modules, which makes the license incompatibility a blocker (for example, Fedora/Redhat wouldnt be able to package dms anymore under the current license). 

The feature is good and actually very useful (I appreciate the contribution and thoughtfulness of it), but we'd have to find another way to implement it. Anything that depends on C would need to be a binary entirely external from dms or built into quickshell itself basically.

I would be more open to this one being a plugin, unless we could workaround the CGO requirements (by either not needing gstreamer, building it into quickshell, or referencing an external binary/dependency that wasn't vendored into dms at all)

domenkozar:
> CGO is a hard no from me, along with the GPL3 lib which isn't a potential blocker it is a blocker because we need to be able to distribute DMS with vendored modules, which makes the license incompatibility a blocker (for example, Fedora/Redhat wouldnt be able to package dms anymore under the current license).

I'm talking to upstream to potentially relicense so that we can ship it. Fingers crossed.

> Anything that depends on C would need to be a binary entirely external from dms or built into quickshell itself basically.

Maybe I should just apply fixes upstream to use gstreamer cli like it used to be in https://github.com/omarroth/doubletake/pull/9 and we're good to include it by default.

domenkozar:
> I would be more open to this one being a plugin, unless we could workaround the CGO requirements (by either not needing gstreamer, building it into quickshell, or referencing an external binary/dependency that wasn't vendored into dms at all)

I've reverted the CGO change in upstream PR, still discussing licensing options.

Spoch-dev:
Closing

domenkozar:
> CGO is a hard no from me, along with the GPL3 lib which isn't a potential blocker it is a blocker because we need to be able to distribute DMS with vendored modules, which makes the license incompatibility a blocker (for example, Fedora/Redhat wouldnt be able to package dms anymore under the current license).
> 
> The feature is good and actually very useful (I appreciate the contribution and thoughtfulness of it), but we'd have to find another way to implement it. Anything that depends on C would need to be a binary entirely external from dms or built into quickshell itself basically.
> 
> I would be more open to this one being a plugin, unless we could workaround the CGO requirements (by either not needing gstreamer, building it into quickshell, or referencing an external binary/dependency that wasn't vendored into dms at all)

CGO is removed and [license is now LGPL](https://github.com/omarroth/doubletake/commit/02d00457f49a6873271a8183b190ebd08e50715d).

domenkozar:
@bbedward waiting on your review, I believe your original concerns were addressed.

domenkozar:
@bbedward waiting on your review, I believe your original concerns were addressed.

toods2:
In my opinion, I think this would be more appropriate for a plugin.

This would introduce several dependencies into the core of DMS for a new feature that will need to be maintained and carries the expectation from users as being robust and performant. The `https://github.com/grandcat/zeroconf` repo doesn't look like it's had any updates/activity in the last 3/4 years, and the AirPlay `https://github.com/omarroth/doubletake` repo describes itself as someone's personal LLM project. Has this been thoroughly tested mirroring to an actual Apple device such as an AppleTV for stability and performance?

domenkozar:
> In my opinion, I think this would be more appropriate for a plugin.

I understand the feeling behind this, but I'd like to say that my inspiration for dms is to become better for macOS and to achieve that we have to aim for more than mediocre.

Having streaming work out of the box is something crucial for a desktop environment these days.

> This would introduce several dependencies into the core of DMS for a new feature that will need to be maintained and carries the expectation from users as being robust and performant. The `https://github.com/grandcat/zeroconf` repo doesn't look like it's had any updates/activity in the last 3/4 years, and the AirPlay `https://github.com/omarroth/doubletake` repo describes itself as someone's personal LLM project. 

For me the beauty of OSS is that we can fix these things as we go. DMS will probably expose bugs that will trickle down and worst case result into maintained forks.

> Has this been thoroughly tested mirroring to an actual Apple device such as an AppleTV for stability and performance?

Yes, I've  tested on the TV and doubletake is starting to maintain a matrix of devices that were tested on.

domenkozar:
Thanks @bbedward rebased and addressed review comments.

PR #2752, opened 2026-07-05
bluetooth: show per-peripheral battery for split devices (ZMK)

dwright134:
## Description

Split keyboards (ZMK Corne and similar) report the battery level of **each half
separately**, but the Bluetooth detail view only ever showed one number.

The cause is that BlueZ's `org.bluez.Battery1` interface — which
`Quickshell.Bluetooth`'s `BluetoothDevice.battery` wraps — exposes a single
percentage per device. A split keyboard is one bonded device: the central half
holds the radio and **relays** the peripheral half's charge as an *additional*
GATT Battery Service (`0x180F`) on the same device object. BlueZ folds only one
of those into `Battery1`, so the second half was invisible.

This PR reads the extra `0x2A19` (Battery Level) characteristics directly over
D-Bus and surfaces them:

- **`BluetoothService`** gains a `peripheralBatteries` map (keyed by device
  address) populated by a small `sh`/`busctl` `Process`. It walks the GATT tree
  of each connected device and emits one `address|percent|label` line per
  Battery Level characteristic that carries a `0x2901` *user-description*
  descriptor — which is exactly how ZMK tags peripheral batteries (e.g.
  `"Peripheral 0"`). The label is read verbatim from that descriptor, so it
  generalizes to N peripherals (`Peripheral 0`, `Peripheral 1`, …) and to
  multiple keyboards.
- **`BluetoothDetail`** renders one extra entry per peripheral next to the
  existing battery text via a `Repeater`.

Design notes:

- **Purely additive.** The central battery is deliberately *skipped* (it's
  already covered by `Battery1`/`device.battery`), so non-split devices render
  exactly as before — `peripheralBatteriesFor()` returns `[]` for them.
- **Association is structural, not heuristic.** Every battery lives under the
  same `/org/bluez/.../dev_XXX` object path, so peripherals attach to the right
  device by construction — there is no name-matching that could go wrong.
- **Survives sleep.** A split half drops its BLE link when idle; a live
  `ReadValue` then fails fast with "Not connected", so the reader falls back to
  BlueZ's cached `Value` (last known level) and best-effort `StartNotify` to keep
  that cache warm. Every `busctl` call is bounded with `--timeout=5`.
- Refreshes on shell start, on `connected` changes, and every 120 s (guarded so
  runs can't overlap).

## Type of change

## Screenshots / video

Bluetooth detail row for a ZMK Corne, before → after:

- Before: `Connected • 90%`

- After: 
Screenshot from 2026-07-05 01-35-28

## Checklist

- [x] My code follows the conventions in CONTRIBUTING.md (`sh -c` + `busctl`,
      matching existing `Process` usage in this service)
- [x] I have tested my changes locally (live on DMS v1.4.6 with a real ZMK Corne;
      central 90% + peripheral 83% render, clean shell reload, no QML warnings)
- [x] New user-facing strings: none new to translate — the format mirrors the
      existing (untranslated) battery/signal text (`"• " + value + "%"`); the
      label text comes from the device's own GATT descriptor
- [ ] Go changes: n/a (QML-only)
- [ ] QML changes: ran `make lint-qml` — please run in CI/devshell; my local Qt's
      `qmllint` couldn't parse `pragma ComponentBehavior: Bound` (bailed at line 2)
      so I could not get a clean local run
- [ ] dlx-docs PR: happy to add docs if you'd like this documented

### DIFF
--- a/quickshell/Modules/ControlCenter/Details/BluetoothDetail.qml
+++ b/quickshell/Modules/ControlCenter/Details/BluetoothDetail.qml
@@ -357,6 +357,27 @@ Rectangle {
                                     visible: text.length > 0
                                 }
 
+                                // Extra batteries reported by split devices (e.g. a
+                                // ZMK keyboard's second half) beyond the single
+                                // value BlueZ exposes via org.bluez.Battery1.
+                                ScriptModel {
+                                    id: peripheralBatteriesModel
+                                    objectProp: "label"
+                                    values: BluetoothService.peripheralBatteriesFor(pairedDelegate.modelData)
+                                }
+
+                                Repeater {
+                                    model: peripheralBatteriesModel
+
+                                    delegate: StyledText {
+                                        required property var modelData
+
+                                        text: "• " + modelData.label + " " + modelData.percentage + "%"
+                                        font.pixelSize: Theme.fontSizeSmall
+                                        color: Theme.surfaceVariantText
+                                    }
+                                }
+
                                 StyledText {
                                     text: pairedDelegate.modelData.signalStrength > 0 ? "• " + pairedDelegate.modelData.signalStrength + "%" : ""
                                     font.pixelSize: Theme.fontSizeSmall
--- a/quickshell/Services/BluetoothService.qml
+++ b/quickshell/Services/BluetoothService.qml
@@ -64,8 +64,268 @@ Singleton {
         });
     }
 
+    // Split keyboards (e.g. ZMK) report each half's charge on a separate GATT
+    // Battery Level (0x2A19) characteristic, but BlueZ's org.bluez.Battery1 —
+    // and therefore Quickshell's device.battery — surfaces only one of them. We
+    // read the extra characteristics (identified by their 0x2901 user-description
+    // descriptor, e.g. "Peripheral 0") directly over D-Bus and expose them here,
+    // keyed by uppercased device address: { "AA:BB:..": [{ percentage, label }] }.
+    property var peripheralBatteries: ({})
+    // Live tracking state for the peripheral Battery Level characteristics,
+    // keyed by their GATT object path.
+    property var _peripheralCharacteristics: ({})
+    property bool _bluezSubscribed: false
+    property bool _bluezRescanQueued: false
+
+    function peripheralBatteriesFor(device) {
+        if (!device || !device.address)
+            return [];
+        return root.peripheralBatteries[device.address.toUpperCase()] || [];
+    }
+
     Component.onCompleted: {
         detectPactlProcess.running = true;
+        maybeInitBluez();
+    }
+
+    // The split-keyboard batteries are read over DMS's persistent system-bus
+    // D-Bus bridge (org.bluez / BlueZ GATT), which requires the "dbus"
+    // capability. When it is unavailable we simply expose no extra batteries
+    // and non-split devices render exactly as before.
+    Connections {
+        target: DMSService
+
+        function onConnectionStateChanged() {
+            root.maybeInitBluez();
+        }
+
+        function onCapabilitiesChanged() {
+            root.maybeInitBluez();
+        }
+
+        function onDbusSignalReceived(subscriptionId, data) {
+            root.handleBluezSignal(data);
+        }
+    }
+
+    function maybeInitBluez() {
+        if (!DMSService.isConnected || !DMSService.capabilities.includes("dbus")) {
+            root._bluezSubscribed = false;
+            return;
+        }
+        subscribeBluezSignals();
+        scanPeripheralBatteries();
+    }
+
+    function subscribeBluezSignals() {
+        if (root._bluezSubscribed)
+            return;
+        root._bluezSubscribed = true;
+
+        DMSService.dbusSubscribe("system", "org.bluez", "", "org.freedesktop.DBus.Properties", "PropertiesChanged", null);
+        DMSService.dbusSubscribe("system", "org.bluez", "", "org.freedesktop.DBus.ObjectManager", "InterfacesAdded", null);
+        DMSService.dbusSubscribe("system", "org.bluez", "", "org.freedesktop.DBus.ObjectManager", "InterfacesRemoved", null);
+    }
+
+    function handleBluezSignal(data) {
+        switch (data.member) {
+        case "PropertiesChanged": {
+            const characteristic = root._peripheralCharacteristics[data.path];
+            if (!characteristic)
+                return;
+            if (data.body?.[0] !== "org.bluez.GattCharacteristic1")
+                return;
+            const changed = data.body?.[1] || {};
+            if (!("Value" in changed))
+                return;
+            const percentage = gattByte(changed.Value);
+            if (percentage === null || !characteristic.label)
+                return;
+            setPeripheralBattery(characteristic.address, characteristic.label, percentage);
+            return;
+        }
+        case "InterfacesAdded":
+        case "InterfacesRemoved":
+            // GATT objects tend to appear in a burst during service discovery;
+            // coalesce them into a single event-driven rescan.
+            queueBluezRescan();
+            return;
+        }
+    }
+
+    function queueBluezRescan() {
+        if (root._bluezRescanQueued)
+            return;
+        root._bluezRescanQueued = true;
+        Qt.callLater(() => {
+            root._bluezRescanQueued = false;
+            root.scanPeripheralBatteries();
+        });
+    }
+
+    // Walk BlueZ's GATT tree once and pick out every peripheral Battery Level
+    // (0x2A19) characteristic that carries a 0x2901 user-description descriptor
+    // (which is exactly how ZMK tags a split half, e.g. "Peripheral 0"). The
+    // unlabeled central battery is deliberately skipped — it is already exposed
+    // via org.bluez.Battery1 / device.battery.
+    function scanPeripheralBatteries() {
+        DMSService.dbusCall("system", "org.bluez", "/", "org.freedesktop.DBus.ObjectManager", "GetManagedObjects", [], response => {
+            const objects = response.result?.values?.[0];
+            if (response.error || !objects)
+                return;
+
+            const batteries = {};
+            const characteristics = {};
+
+            for (const path in objects) {
+                const characteristic = objects[path]?.["org.bluez.GattCharacteristic1"];
+                if (!characteristic || !String(characteristic.UUID || "").toLowerCase().startsWith("00002a19"))
+                    continue;
+
+                const descriptorPath = findUserDescription(objects, path);
+                if (!descriptorPath)
+                    continue;
+
+                const devicePath = devicePathForCharacteristic(objects, characteristic);
+                const device = objects[devicePath]?.["org.bluez.Device1"];
+                if (!device?.Address)
+                    continue;
+
+                const address = String(device.Address).toUpperCase();
+                const descriptor = objects[descriptorPath]?.["org.bluez.GattDescriptor1"];
+
+                const tracked = {
+                    "address": address,
+                    "path": path,
+                    "descriptorPath": descriptorPath,
+                    "label": gattString(descriptor?.Value),
+                    "connected": device.Connected === true
+                };
+                characteristics[path] = tracked;
+
+                const percentage = gattByte(characteristic.Value);
+                if (tracked.label && percentage !== null)
+                    appendPeripheralBattery(batteries, address, tracked.label, percentage);
+            }
+
+            root._peripheralCharacteristics = characteristics;
+            root.peripheralBatteries = batteries;
+
+            for (const path in characteristics) {
+                if (characteristics[path].connected)
+                    primePeripheralCharacteristic(characteristics[path]);
+            }
+        });
+    }
+
+    // On a cold cache BlueZ has no stored Value yet, so subscribe for future
+    // notifications and read the current label/level once. StartNotify stays
+    // active because it is owned by the persistent DMS D-Bus connection, not a
+    // short-lived helper process.
+    function primePeripheralCharacteristic(characteristic) {
+        DMSService.dbusCall("system", "org.bluez", characteristic.path, "org.bluez.GattCharacteristic1", "StartNotify", [], null);
+
+        if (!characteristic.label) {
+            DMSService.dbusCall("system", "org.bluez", characteristic.descriptorPath, "org.bluez.GattDescriptor1", "ReadValue", [{}], response => {
+                const label = gattString(response.result?.values?.[0]);
+                if (!label)
+                    return;
+                const updated = Object.assign({}, root._peripheralCharacteristics);
+                if (updated[characteristic.path])
+                    updated[characteristic.path].label = label;
+                root._peripheralCharacteristics = updated;
+            });
+        }
+
+        DMSService.dbusCall("system", "org.bluez", characteristic.path, "org.bluez.GattCharacteristic1", "ReadValue", [{}], response => {
+            const percentage = gattByte(response.result?.values?.[0]);
+            const label = root._peripheralCharacteristics[characteristic.path]?.label;
+            if (percentage === null || !label)
+                return;
+            setPeripheralBattery(characteristic.address, label, percentage);
+        });
+    }
+
+    function devicePathForCharacteristic(objects, characteristic) {
+        const servicePath = characteristic.Service;
+        const service = objects[servicePath]?.["org.bluez.GattService1"];
+        return service?.Device || "";
+    }
+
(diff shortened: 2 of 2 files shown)


dwright134:
Pushed a follow-up fix (b59462fc): the split-peripheral **battery label** was read from BlueZ's cached `Value` first, but `busctl` prints `ay 0` for an empty byte array, so the `[ -n "$lv" ]` guard never fell back to `ReadValue`. On a cold cache (the normal state right after connect) the label came back empty and the peripheral half was dropped, so split keyboards still showed only the central battery.

Now the label is read via `ReadValue` first (matching how the percentage is read), falling back to the cached `Value`, which also warms BlueZ's cache instead of depending on it already being warm. Verified live on a ZMK Corne — both halves now show.

quickshell/Services/BluetoothService.qml
```diff
@@ -617,4 +639,80 @@ Singleton {
...
+    [ -n "\$label" ] || continue
+    [ "\$conn" = "true" ] && \$B call org.bluez "\$ch" org.bluez.GattCharacteristic1 StartNotify >/dev/null 2>&1
+    pct=""
+    [ "\$conn" = "true" ] && pct=$(\$B call org.bluez "\$ch" org.bluez.GattCharacteristic1 ReadValue 'a{sv}' 0 2>/dev/null | awk '{print \$NF}')
+    case "\$pct" in ''|*[!0-9]*) pct=$(\$B get-property org.bluez "\$ch" org.bluez.GattCharacteristic1 Value 2>/dev/null | awk '{print \$NF}') ;; esac
+    case "\$pct" in ''|*[!0-9]*) continue ;; esac
+    echo "\$addr|\$pct|\$label"
+  done
+done
+`]
```
Mikilio: I am not a Maintainer but I think part is better accomplished using the go backend. I remember Implementing some dbus related code there before. Using shell script is kind of hacky and not ideal for maintenance and performance. You will struggle to find any example of this anywhere else on the codebase.

I think it's better to rethink this approach.

Mikilio (changes requested):

quickshell/Modules/ControlCenter/Details/BluetoothDetail.qml
```diff
@@ -357,6 +357,21 @@ Rectangle {
                                     visible: text.length > 0
                                 }
 
+                                // Extra batteries reported by split devices (e.g. a
+                                // ZMK keyboard's second half) beyond the single
+                                // value BlueZ exposes via org.bluez.Battery1.
+                                Repeater {
+                                    model: BluetoothService.peripheralBatteriesFor(pairedDelegate.modelData)
```
bbedward: Using [ScriptModel](https://quickshell.org/docs/v0.3.0/types/Quickshell/ScriptModel/) here is advised

quickshell/Services/BluetoothService.qml
```diff
@@ -617,4 +639,80 @@ Singleton {
             callback = null;
         }
     }
+
+    // Periodically refresh peripheral (split) battery levels while an adapter is
+    // present; the read is cheap and also keeps BlueZ's cached values warm.
+    Timer {
+        interval: 120000
```
bbedward: It's against DMS philosophy to do any sort of unconditional polling, we prefer to be strictly event-driven especially in a case like this where its possible to do so

quickshell/Services/BluetoothService.qml
```diff
@@ -617,4 +639,80 @@ Singleton {
...
+        running: root.available
+        onTriggered: root.refreshPeripheralBatteries()
+    }
+
+    // Walks BlueZ's GATT tree for every connected device and emits one
+    // "address|percent|label" line per peripheral Battery Level characteristic
+    // (a 0x2A19 that carries a 0x2901 user-description). The central battery is
+    // already covered by org.bluez.Battery1, so it is deliberately skipped.
+    Process {
+        id: peripheralBatteryProcess
```
bbedward: We can avoid busctl script +  polling, DMS already exposes a persistent system-bus D-Bus bridge, which is a much better fit for this.

You can reference the [`[KDEConnectService.qml](https://github.com/AvengeMedia/dms-plugins/blob/master/DankKDEConnect/services/KDEConnectService.qml)`](https://github.com/AvengeMedia/dms-plugins/blob/master/DankKDEConnect/services/KDEConnectService.qml) implementation for the general `dbusCall`/`dbusSubscribe` pattern.

The intended flow should be:

1. Call `org.freedesktop.DBus.ObjectManager.GetManagedObjects` once to build the initial GATT state.
2. Subscribe to `InterfacesAdded`/`InterfacesRemoved` to discover changes.
3. Subscribe to `PropertiesChanged` for battery updates.
4. Call `StartNotify` through the persistent DMS D-Bus connection.
5. Rescan only in response to BlueZ object changes—no recurring polling timer.

Something along these lines:

```qml
// BlueZ exposes the primary battery through org.bluez.Battery1.
// Additional split-device batteries are represented by labeled 0x2A19
// characteristics with a 0x2901 user-description descriptor.

property var peripheralBatteries: ({})
property var _peripheralCharacteristics: ({})
property bool _bluezSubscribed: false
property bool _bluezRescanQueued: false

function peripheralBatteriesFor(device) {
    if (!device?.address)
        return [];

    return peripheralBatteries[device.address.toUpperCase()] || [];
}

Connections {
    target: DMSService

    function onConnectionStateChanged() {
        if (!DMSService.isConnected) {
            root._bluezSubscribed = false;
            return;
        }

        root._initializeBluez();
    }

    function onDbusSignalReceived(subscriptionId, data) {
        root._handleBluezSignal(data);
    }
}

Component.onCompleted: {
    if (DMSService.isConnected)
        root._initializeBluez();
}

function _initializeBluez() {
    _subscribeBluezSignals();
    _scanPeripheralBatteries();
}

function _subscribeBluezSignals() {
    if (_bluezSubscribed)
        return;

    _bluezSubscribed = true;

    DMSService.dbusSubscribe(
        "system",
        "org.bluez",
        "",
        "org.freedesktop.DBus.Properties",
        "PropertiesChanged",
        null
    );

    DMSService.dbusSubscribe(
        "system",
        "org.bluez",
        "",
        "org.freedesktop.DBus.ObjectManager",
        "InterfacesAdded",
        null
    );

    DMSService.dbusSubscribe(
        "system",
        "org.bluez",
        "",
        "org.freedesktop.DBus.ObjectManager",
        "InterfacesRemoved",
        null
    );
}

function _handleBluezSignal(data) {
    switch (data.member) {
    case "PropertiesChanged": {
        const characteristic = _peripheralCharacteristics[data.path];
        if (!characteristic)
            return;

        if (data.body?.[0] !== "org.bluez.GattCharacteristic1")
            return;

        const changed = data.body?.[1] || {};
        if (!("Value" in changed))
            return;

        const percentage = _gattByte(_variantValue(changed.Value));
        if (percentage === null || !characteristic.label)
            return;

        _setPeripheralBattery(
            characteristic.address,
            characteristic.label,
            percentage
        );
        return;
    }

    case "InterfacesAdded":
    case "InterfacesRemoved":
        // GATT objects are often added in a burst during service discovery.
        // Coalesce those signals into one event-driven rescan.
        _queueBluezRescan();
        return;
    }
}

function _queueBluezRescan() {
    if (_bluezRescanQueued)
        return;

    _bluezRescanQueued = true;

    Qt.callLater(() => {
        _bluezRescanQueued = false;
        _scanPeripheralBatteries();
    });
}

function _scanPeripheralBatteries() {
    DMSService.dbusCall(
        "system",
        "org.bluez",
        "/",
        "org.freedesktop.DBus.ObjectManager",
        "GetManagedObjects",
        [],
        response => {
            const objects = response.result?.values?.[0];
            if (response.error || !objects)
                return;

            const batteries = {};
            const characteristics = {};

            for (const path in objects) {
                const characteristic =
                    objects[path]?.["org.bluez.GattCharacteristic1"];

                if (!characteristic
                        || !String(characteristic.UUID || "")
                            .toLowerCase()
                            .startsWith("00002a19")) {
                    continue;
                }

                const descriptorPath =
                    _findUserDescription(objects, path);

                // The unlabeled 0x2A19 characteristic is the primary battery,
                // which is already exposed through org.bluez.Battery1.
                if (!descriptorPath)
                    continue;

                const devicePath = _devicePathForCharacteristic(
                    objects,
                    characteristic
                );

                const device =
                    objects[devicePath]?.["org.bluez.Device1"];

                if (!device?.Address)
                    continue;

                const address = String(device.Address).toUpperCase();
                const descriptor =
                    objects[descriptorPath]?.["org.bluez.GattDescriptor1"];

                const tracked = {
                    address: address,
                    path: path,
                    descriptorPath: descriptorPath,
                    label: _gattString(descriptor?.Value),
                    connected: device.Connected === true
                };

                characteristics[path] = tracked;

                const percentage = _gattByte(characteristic.Value);
                if (tracked.label && percentage !== null)
                    _appendPeripheralBattery(
                        batteries,
                        address,
                        tracked.label,
                        percentage
                    );
            }

            _peripheralCharacteristics = characteristics;
            peripheralBatteries = batteries;

            for (const path in characteristics) {
                const characteristic = characteristics[path];

                if (characteristic.connected)
                    _primePeripheralCharacteristic(characteristic);
            }
        }
    );
}

function _primePeripheralCharacteristic(characteristic) {
    // StartNotify remains active because it is owned by the persistent
    // DMS daemon's D-Bus connection, unlike a short-lived busctl process.
    DMSService.dbusCall(
        "system",
        "org.bluez",
        characteristic.path,
        "org.bluez.GattCharacteristic1",
        "StartNotify",
        [],
        null
    );

    if (!characteristic.label) {
        DMSService.dbusCall(
            "system",
            "org.bluez",
            characteristic.descriptorPath,
            "org.bluez.GattDescriptor1",
            "ReadValue",
            [{}],
            response => {
                const label = _gattString(
                    response.result?.values?.[0]
                );

                if (!label)
                    return;

                const updated = Object.assign(
                    {},
                    _peripheralCharacteristics
                );

                if (updated[characteristic.path])
                    updated[characteristic.path].label = label;

                _peripheralCharacteristics = updated;
            }
        );
    }

    DMSService.dbusCall(
        "system",
        "org.bluez",
        characteristic.path,
        "org.bluez.GattCharacteristic1",
        "ReadValue",
        [{}],
        response => {
            const percentage = _gattByte(
                response.result?.values?.[0]
            );

            const label =
                _peripheralCharacteristics[
                    characteristic.path
                ]?.label;

            if (percentage === null || !label)
                return;

            _setPeripheralBattery(
                characteristic.address,
                label,
                percentage
            );
        }
    );
}

function _devicePathForCharacteristic(objects, characteristic) {
    const servicePath = _variantValue(characteristic.Service);
    const service =
        objects[servicePath]?.["org.bluez.GattService1"];

    return _variantValue(service?.Device) || "";
}

function _findUserDescription(objects, characteristicPath) {
    for (const path in objects) {
        if (!path.startsWith(characteristicPath + "/desc"))
            continue;

        const descriptor =
            objects[path]?.["org.bluez.GattDescriptor1"];

        if (descriptor
                && String(descriptor.UUID || "")
                    .toLowerCase()
                    .startsWith("00002901")) {
            return path;
        }
    }

    return "";
}

function _appendPeripheralBattery(map, address, label, percentage) {
    if (!map[address])
        map[address] = [];

    map[address].push({
        percentage: percentage,
        label: label
    });
}

function _setPeripheralBattery(address, label, percentage) {
    const map = JSON.parse(JSON.stringify(peripheralBatteries));

    if (!map[address])
        map[address] = [];

    const existing = map[address].find(
        battery => battery.label === label
    );

    if (existing)
        existing.percentage = percentage;
    else
        map[address].push({
            percentage: percentage,
            label: label
        });

    peripheralBatteries = map;
}

function _variantValue(value) {
    if (value && typeof value === "object") {
        if ("value" in value)
            return value.value;
        if ("V" in value)
            return value.V;
    }

    return value;
}

// The Go bridge serializes D-Bus `ay` values as base64 strings.
function _gattByte(value) {
    value = _variantValue(value);

    if (typeof value !== "string" || !value.length)
        return null;

    const decoded = Qt.atob(value);

    // An empty cached byte array must not be interpreted as 0%.
    return decoded.length === 1
        ? decoded.charCodeAt(0)
        : null;
}

function _gattString(value) {
    value = _variantValue(value);

    if (typeof value !== "string" || !value.length)
        return "";

    return Qt.atob(value);
}
```

The code snippet is AI-generated ^, so I wouldnt trust it - its just a rough sketch of what it should look like

bbedward (changes requested):

quickshell/Services/BluetoothService.qml
Mikilio: I take my advice back in favor of @bbedward commit to use the D-Bus bridge I didn't know of.

dwright134:
Reworked to be fully event-driven over the DMS D-Bus bridge. Thanks @bbedward for the dbusCall/dbusSubscribe pointer and the KDEConnectService reference; the structure follows your sketch.

Changes:
- Dropped the busctl Process and the 120s polling Timer entirely. It now does one GetManagedObjects scan to build GATT state, InterfacesAdded/InterfacesRemoved trigger coalesced rescans, PropertiesChanged delivers live updates, and StartNotify runs on the persistent DMS connection. No polling.
- BluetoothDetail's peripheral list now uses ScriptModel.
- Verified the bridge serializes `ay` as base64 and unwraps variants, so the values are decoded locally and Qt.atob is gone (its string overload is deprecated and warned on every call).
- Rebased onto current master so it picks up the backend audit's dbusClientId.

Tested on a real ZMK Corne: both halves render, and disconnect/reconnect drives InterfacesRemoved then InterfacesAdded then a rescan and reconverges. Multi-device keying is structural (by device address, no name matching) and covered by an offline harness since I only have one split device.

quickshell/Services/BluetoothService.qml
dwright134: Done, the timer is gone. Rescans are triggered only by InterfacesAdded/InterfacesRemoved (coalesced via Qt.callLater), and battery values update from PropertiesChanged.

quickshell/Services/BluetoothService.qml
dwright134: Adopted the bridge approach, implemented along the lines of your sketch and verified against the Go dbus manager (base64 `ay`, unwrapped variants). No more short-lived processes; StartNotify now lives on the persistent DMS connection.

quickshell/Modules/ControlCenter/Details/BluetoothDetail.qml
dwright134: Switched to ScriptModel (objectProp: "label"), matching the other models in the file.

quickshell/Services/BluetoothService.qml
```diff
@@ -64,8 +64,268 @@ Singleton {
...
+    }
+
+    function subscribeBluezSignals() {
+        if (root._bluezSubscribed)
+            return;
+        root._bluezSubscribed = true;
+
+        DMSService.dbusSubscribe("system", "org.bluez", "", "org.freedesktop.DBus.Properties", "PropertiesChanged", null);
+        DMSService.dbusSubscribe("system", "org.bluez", "", "org.freedesktop.DBus.ObjectManager", "InterfacesAdded", null);
+        DMSService.dbusSubscribe("system", "org.bluez", "", "org.freedesktop.DBus.ObjectManager", "InterfacesRemoved", null);
```
bbedward, line 127: These discard the subscription ids, so the handler below can't tell your signals from anyone else's. Capture response.result.subscriptionId in the callback like SessionService.qml:633 does and keep them in a set.

quickshell/Services/BluetoothService.qml
```diff
@@ -64,8 +64,268 @@ Singleton {
...
+        function onConnectionStateChanged() {
+            root.maybeInitBluez();
+        }
+
+        function onCapabilitiesChanged() {
+            root.maybeInitBluez();
+        }
+
+        function onDbusSignalReceived(subscriptionId, data) {
+            root.handleBluezSignal(data);
```
bbedward, line 107: dbusSignalReceived is broadcast to every subscriber in the process (KDEConnect/Valent/Clight plugins share this bridge). First guard here should be if (!_bluezSubscriptionIds[subscriptionId]) return;

quickshell/Services/BluetoothService.qml
```diff
@@ -64,8 +64,268 @@ Singleton {
...
+            queueBluezRescan();
+            return;
+        }
+    }
+
+    function queueBluezRescan() {
+        if (root._bluezRescanQueued)
+            return;
+        root._bluezRescanQueued = true;
+        Qt.callLater(() => {
```
bbedward, line 160: This should use a debounce instead of `Qt.callLater`

bbedward (changes requested):

PR #2805, opened 2026-07-10
core: fix security and concurrency issues found in a backend audit

asteroidzman:
While auditing the `core` backend for a downstream fork, I found a few security and concurrency issues that apply to upstream as well. Each is an independent, self-contained fix; happy to split into separate PRs if you'd prefer.

## Security
- **privesc**: pipe the sudo password via stdin (`sudo -S`) instead of embedding it in the command string, so it no longer appears in `argv` (readable by any local user via `/proc/<pid>/cmdline` or `ps`).
- **greeter**: tokenize a session `.desktop` `Exec=` line into argv and `execve` directly instead of running it through `/bin/sh -c`, closing a command-injection path via user-writable `~/.local/share/wayland-sessions`.
- **plugins**: reject path-separator / `..` in plugin id/name before joining into a filesystem path, closing an arbitrary-directory-delete in the uninstall/update fallback.

## Concurrency / robustness
- NetworkManager backend: guard shared state accessed from D-Bus signal handlers, fix a few signal races, and tighten ethernet/wifi state handling.
- clipboard, cups, freedesktop, router/server: fix data races and lifecycle issues found alongside the above.

## Notes
- Touches `core/` only — no UI/QML changes.
- `make build`, `go vet`, and the keybinds tests pass on current `master`.

### DIFF
--- a/core/internal/greeter/session_launcher.go
+++ b/core/internal/greeter/session_launcher.go
@@ -94,70 +94,75 @@ func resolveSessionExecInDirs(sessionID string, dirs []string) (string, error) {
 	return "", fmt.Errorf("session desktop file %q was not found", id)
 }
 
-// tokenizeExecLine splits a Desktop Entry Exec= value into argv, honoring
-// basic single/double quoting and backslash escapes, but never invokes a
-// shell. Session .desktop files can live in a user-writable directory
-// (~/.local/share/wayland-sessions), so shell metacharacters in Exec=
-// must never reach an actual shell -- they're just literal argv text here.
-func tokenizeExecLine(execLine string) ([]string, error) {
-	var tokens []string
+// parseExecString splits a Desktop Entry Exec= value into argv without
+// involving a shell, mirroring quickshell's DesktopEntry::parseExecString
+// (string quoting, value escapes, field code stripping).
+func parseExecString(execLine string) []string {
+	var args []string
 	var cur strings.Builder
-	hasCur := false
-	runes := []rune(execLine)
-	i := 0
-	for i < len(runes) {
-		c := runes[i]
+	inString := false
+	escape := 0
+	percent := false
+
+	for _, c := range execLine {
 		switch {
-		case c == ' ' || c == '\t':
-			if hasCur {
-				tokens = append(tokens, cur.String())
-				cur.Reset()
-				hasCur = false
-			}
-			i++
-		case c == '\'':
-			hasCur = true
-			i++
-			for i < len(runes) && runes[i] != '\'' {
-				cur.WriteRune(runes[i])
-				i++
-			}
-			if i >= len(runes) {
-				return nil, fmt.Errorf("unterminated single quote")
-			}
-			i++
-		case c == '"':
-			hasCur = true
-			i++
-			for i < len(runes) && runes[i] != '"' {
-				if runes[i] == '\\' && i+1 < len(runes) {
-					if next := runes[i+1]; next == '"' || next == '\\' || next == '$' || next == '`' {
-						cur.WriteRune(next)
-						i += 2
-						continue
-					}
+		case escape == 0 && c == '\\':
+			escape = 1
+		case inString:
+			switch {
+			case c == '\\':
+				escape++
+				if escape == 4 {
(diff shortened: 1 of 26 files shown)


core/internal/server/router.go
```diff
@@ -166,7 +166,14 @@ func RouteRequest(conn net.Conn, req models.Request) {
...
 		}
-		serverDbus.HandleRequest(conn, req, dbusManager, dbusClientID)
+		// per-connection, not a shared constant: sharing one ID across every
+		// client meant two clients subscribing to the same signal fought
+		// over one channel, and either one disconnecting tore down every
+		// other client's match-rule subscriptions too (see server.go's use
+		// of the same clientID formula for the streaming "subscribe"
+		// endpoint, which must stay in sync with this one).
+		clientID := fmt.Sprintf("meta-client-%p", conn)
+		serverDbus.HandleRequest(conn, req, dbusManager, clientID)
```
bbedward: I suspect this probably breaks client signal delivery because ClientID lookup wont match in the QMl-side. Same problem in server.go, you can test plugins like DankKDEConnect to see if they still work afterwards

core/internal/server/server.go
```diff
@@ -415,10 +419,25 @@ func handleConnection(conn net.Conn) {
...
-		go RouteRequest(conn, req)
+		go routeRequestRecovered(conn, req)
 	}
 }
 
+// routeRequestRecovered runs RouteRequest with a panic boundary so that a
+// bug in any one handler (bad type assertion, nil deref, out-of-range
+// index, etc. -- reachable from client-controlled req.Method/req.Params)
+// can't crash the whole daemon and disconnect every other client. It
+// reports back to the offending client as a normal error response instead.
```
bbedward: I prefer to cut down all of these comments, its way too verbose for the code its covering which is largely self-documenting anyway

core/internal/greeter/session_launcher.go
```diff
@@ -93,18 +94,105 @@ func resolveSessionExecInDirs(sessionID string, dirs []string) (string, error) {
 	return "", fmt.Errorf("session desktop file %q was not found", id)
 }
 
+// tokenizeExecLine splits a Desktop Entry Exec= value into argv, honoring
+// basic single/double quoting and backslash escapes, but never invokes a
+// shell. Session .desktop files can live in a user-writable directory
+// (~/.local/share/wayland-sessions), so shell metacharacters in Exec=
+// must never reach an actual shell -- they're just literal argv text here.
```
bbedward: Not a security fix since the user is already authenticated at this point, but maybe a spec correctness change

core/internal/greeter/session_launcher.go
```diff
@@ -93,18 +94,105 @@ func resolveSessionExecInDirs(sessionID string, dirs []string) (string, error) {
...
+		c := runes[i]
+		switch {
+		case c == ' ' || c == '\t':
+			if hasCur {
+				tokens = append(tokens, cur.String())
+				cur.Reset()
+				hasCur = false
+			}
+			i++
+		case c == '\'':
```
bbedward: I believe XDG spec dictates single quotes shouldnt be treated as a quote, but a reserved character

core/internal/greeter/session_launcher.go
```diff
@@ -93,18 +94,105 @@ func resolveSessionExecInDirs(sessionID string, dirs []string) (string, error) {
...
 	}
 
+	tokens, err := tokenizeExecLine(execLine)
+	if err != nil {
+		return fmt.Errorf("session %q has an invalid Exec command: %w", sessionID, err)
+	}
+
+	argv := make([]string, 0, len(tokens))
+	for _, tok := range tokens {
+		if strings.HasPrefix(tok, "%") {
```
bbedward: I would maybe reference quickshell code DeskopEntries for all of this stuff, to ensure its following the XDG specification

bbedward (changes requested):
Most of the changes are good, but I'm skeptical about the dbus ID changes breaking clients. That needs verification, as well as ensuring the greeter launcher changes are adhering to XDG specs, and generally flattening the verbosity of comments

PR #2809, opened 2026-07-10
Feature/night mode exceptions

euletheia:
## Description

Add night mode exceptions (fullscreen and custom app list).

Drawing heavily on hthienloc@de3e8cffa073dcae2c66c40565ba12d1c1412340, give the possibility to specify a list of applications excluded from Night Mode, alongside the ability to exclude fullscreen applications.

Logic is implemented on newly focused windows, and integrates with compositors overview mode so that Night Mode is resumed on overview.[^1]

[^1]: This means that enabling fullscreen exclusion while fullscreen will do nothing because the implementation reacts to Toplevel change (application focus).

### Notes

- Abstracted the appId match function from `MprisController:qml:20` to `Paths.qml:154` with other (loosely-)related functions.

## Type of change

- [ ] Documentation
- [ ] Other

## Screenshots / video[^2]

[contrib_night-mode-exceptions.webm](https://github.com/user-attachments/assets/f9dbe49a-3298-45c3-a700-af433a40fd5f)

[^2]: Screen gamma change is obviously not visible in the video, but "it works on my machine". 

## Checklist


quickshell/Common/settings/SessionSpec.js
```diff
@@ -37,6 +37,8 @@ var SPEC = {
     longitude: { def: 0.0 },
     nightModeUseIPLocation: { def: false },
     nightModeLocationProvider: { def: "" },
+    nightModePaused: { def: "" },
```
dionjoshualobo: Isn't `nightModePaused: { def: false },` better? Since it is treated as boolean everywhere else?

euletheia: Totally, my mistake.

dionjoshualobo:
I noticed that when the Fullscreen Applications is turned on, and if any one app in the workspace is fullscreen, it turns of night mode for all the apps in the workspace. Is it intentional?

euletheia:
That's a regression caused by the unnecessary additional commit I carelessly pushed earlier.
I just dropped it and NightMode should resume if you switch from a fullscreen app to any non-fullscreen window, as was originally planned.

Thank you !

dionjoshualobo:
The "excluding apps from night mode" makes sense, but the one with "full screen disabling night mode" doesn't make sense to me. Anyway, up to the maintainers to approve

euletheia:
To me, the point of allowing users to disable night mode for fullscreen apps is to ensure color accuracy for applications typically run in fullscreen mode, like media players and video games.

quickshell/Services/DisplayService.qml
```diff
@@ -1070,6 +1077,62 @@ Singleton {
...
+        const moddedId = Paths.moddedAppId(appId);
+        const desktopId = DesktopEntries.heuristicLookup(moddedId)?.id ?? "";
+        const isExcludedAppException = excludedApps.some(excludedId => Paths.isAppIdMatch(appId, excludedId, desktopId));
+        if (isExcludedAppException) {
+            SessionData.cacheNightModeExcludedAppMatch(appId);
+        }
+        return isExcludedAppException;
+    }
+
+    function handleNightModeExceptions(topLevel) {
```
bbedward: topLevel argument seems unused. 

Also I think we can rewrite the function to read a little clearer

```js
function handleNightModeExceptions() {
    if (!nightModePaused && !nightModeEnabled) {
        return;
    }

    const isInOverview =
        (CompositorService.isHyprland && root.hyprlandOverviewLoader?.item?.overviewOpen) ||
        (CompositorService.isNiri && NiriService.inOverview) ||
        (CompositorService.isMango && MangoService.inOverview);

    if (nightModePaused && isInOverview) {
        resumeNightMode();
        return;
    }

    const activeApp = ToplevelManager.activeToplevel;
    if (!activeApp) {
        return;
    }

    const shouldPause =
        (SettingsData.nightModeExcludeFullscreen && activeApp.fullscreen) ||
        isNightModeExcludedApp(activeApp.appId);

    if (shouldPause) {
        if (!nightModePaused && nightModeEnabled) {
            pauseNightMode();
        }
        return;
    }

    if (nightModePaused) {
        resumeNightMode();
    }
}
```

bbedward (changes requested):

quickshell/Services/DisplayService.qml
euletheia: Agreed, I've pushed just that. Thanks.

euletheia:
Two shortcomings I've noticed[^1] and I'm not sure how to solve:

[^1]: at least on Niri, and at least on my laptop.

- Making a window fullscreen with `nightModeExcludeFullscreen` enabled seems to cause some minor "flickering".
- Switching from a fullscreen window (with `nightModeExcludeFullscreen` enabled) to the Niri overview mode introduces a slight but noticeable delay before the blurred wallpaper shows.

---

EDIT:

- seems like 2bafc212291b260f313e10ff70ee6e72fbd2fbc0 solved most of the problem.
- I'd still like to "fix" the abrupt temperature change: following is a basic QML patch to animate state switching. 
  - I'm not sure about how to properly do error handling here,
  - It spawns a lot of IPC requests (very noisy in debug mode).

If time permits, I'd be glad to try my hand at a Go implementation, but I'm unsure how an animation player that handles bezier curves would fit in the `wayland` pkg. Would it make sense for it to live in a separate file, and be hooked into the `manager.go:schedulerLoop()` ? If you have any time to spare to provide some hints or directions, it would be greatly appreciated.

[night-mode-transition-basic.patch](https://github.com/user-attachments/files/30395507/night-mode-transition-basic.patch)

quickshell/Services/DisplayService.qml
```diff
@@ -1070,6 +1077,73 @@ Singleton {
...
     }
 
+    function pauseNightMode() {
+        disableNightMode();
+        nightModePaused = true;
+        SessionData.setNightModePaused(true);
+    }
+    function resumeNightMode() {
+        enableNightMode();
+        nightModePaused = false;
```
bbedward, line 1087: I think the paused state should be in-memory only, not persisted to the session json. Similar with the auto disabling/enabling, doesnt need to be persisted to disk

quickshell/Services/DisplayService.qml
```diff
@@ -1484,6 +1558,20 @@ Singleton {
...
+        function onNightModePausedChanged() {
+            nightModePaused = SessionData.nightModePaused;
+        }
+        function onNightModeExcludedAppsMatchesCacheChanged() {
+            nightModeExcludedAppsMatchesCache = SessionData.nightModeExcludedAppsMatchesCache;
+        }
+    }
+
+    Connections {
+        target: CompositorService
```
bbedward, line 1570: Should probably connect to ToplevelManager.activeToplevel instead of the CompositorService list

quickshell/Services/DisplayService.qml
```diff
@@ -1070,6 +1077,73 @@ Singleton {
...
+        SessionData.setNightModePaused(false);
+    }
+
+    function isNightModeExcludedApp(appId: string): bool {
+        const excludedApps = SettingsData.nightModeExcludedApps || "";
+        if (excludedApps.length === 0) {
+            return false;
+        }
+
+        const exclusionCache = nightModeExcludedAppsMatchesCache;
```
bbedward, line 1097: Don't need to cache th is to disk either, the check is cheap

quickshell/Services/DisplayService.qml
```diff
@@ -105,6 +107,11 @@ Singleton {
     property bool nightModeActive: nightModeEnabled
 
     property bool nightModeEnabled: false
+    property bool nightModePaused: false
+    property var nightModeExcludedAppsMatchesCache: []
+
+    property var hyprlandOverviewLoader: null
```
bbedward, line 113: This should be restructured so the Services don't hold a reference to a UI loader/module, can simplify this wiring a lot I think

bbedward (changes requested):

quickshell/Services/DisplayService.qml
euletheia: Unless I'm missing something, it requires three connections to match the previous behavior (but still reduces function calls).

quickshell/Services/DisplayService.qml
euletheia: I ended up with [5a1dddf](https://github.com/AvengeMedia/DankMaterialShell/pull/2809/commits/5a1dddf739293552e77d6fcdefbdbfe08938a065) (tested in a VM). Would this work for you ?

quickshell/Services/DisplayService.qml
euletheia: I removed the `nightModePaused` persistence, but I'm not sure I fully understand you: do you also want to remove the session persistence for `nightModeEnabled`, or for `nightModeAutoEnabled` (or both) ?

bbedward: Keeping it on user-initiated actions is fine, I mean they shouldnt be changed automatically with the night mode exceptions. So enable/Disable user actions should persist it. pause/resume should not change the persistent setting

euletheia:
@bbedward
All functions in `DisplayService.qml` now call the same setter `_setGammaEnabled()`, independently of the state they set.
The original behavior is preserved and everything works as expected on my end.

PR #2825, opened 2026-07-11
feat(network): initial WiFi hotspot support

ronharel02:
Adds WiFi hotspot support to DMS through an optional backend capability, with NetworkManager as the first implementation. Unsupported backends and older servers (gated behind IPC API v28) hide the feature automatically.

### Settings

The WiFi settings tab now includes a dedicated hotspot card for configuring the SSID, password, device, and band. Users can save, start, stop, and edit the DMS-managed hotspot profile. An empty password creates an open network, and editing prefills the stored password so re-saving never silently changes security. Starting on a connected radio prompts for confirmation before disconnecting the current WiFi connection.

image
image

### Control Center

Control Center now shows a compact hotspot row alongside the WiFi controls. Configured hotspots can be started or stopped directly. When setup is required, the row links to the WiFi settings tab.

image

## Backend

DMS manages a dedicated NetworkManager profile identified by its `stable-id`, leaving user-created hotspots untouched. Activation is asynchronous and follows NetworkManager state signals so the UI can show progress and report failures.

Automatic device selection prefers the radio already hosting the hotspot, followed by idle AP-capable radios. A connected radio is used only as a last resort.

AP-mode connections are isolated from normal client WiFi state and operations, including network lists, saved-profile actions, QR and autoconnect handling, and connection-priority rewrites.

## Requirements

- An AP-capable WiFi adapter.
- NetworkManager and `dnsmasq` for NetworkManager's shared IPv4 method.

## Future work

Some adapters support simultaneous client and access-point operation through STA+AP concurrency. Supporting this requires creating a virtual AP interface and coordinating its channel with the client connection, which NetworkManager does not handle automatically. Until then, sharing an active WiFi connection requires a second adapter.

## Testing

Tested with hermetic NetworkManager mocks and manually on an MT7921 using 2.4 and 5 GHz bands, open and WPA2 networks, and a phone client connected through an Ethernet uplink.

### DIFF
--- a/core/internal/server/network/backend_networkmanager_hotspot.go
+++ b/core/internal/server/network/backend_networkmanager_hotspot.go
@@ -1,0 +1,589 @@
+package network
+
+import (
+	"fmt"
+	"sort"
+
+	"github.com/AvengeMedia/DankMaterialShell/core/internal/errdefs"
+	"github.com/AvengeMedia/DankMaterialShell/core/internal/log"
+	"github.com/Wifx/gonetworkmanager/v2"
+)
+
+const (
+	nmWiFiDeviceCapAP        uint32 = 0x00000040
+	nmWiFiDeviceCapFreqValid uint32 = 0x00000100
+	nmWiFiDeviceCapFreq2GHz  uint32 = 0x00000200
+	nmWiFiDeviceCapFreq5GHz  uint32 = 0x00000400
+
+	dmsHotspotConnectionID = "DankMaterialShell Hotspot"
+	dmsHotspotStableID     = "dms-hotspot"
+)
+
+var _ HotspotBackend = (*NetworkManagerBackend)(nil)
+
+// ConfigureHotspot validates only DMS-owned constraints: SSID presence, band
+// names, device capability, and the not-active guard. SSID byte-length and
+// WPA-PSK password policy are deliberately left to NetworkManager, whose
+// AddConnection/Update errors are returned to the caller; duplicating its
+// evolving rules here would only let them drift.
+func (b *NetworkManagerBackend) ConfigureHotspot(req HotspotRequest) error {
+	if req.SSID == "" {
+		return fmt.Errorf("hotspot SSID cannot be empty")
+	}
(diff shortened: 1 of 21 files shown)


core/internal/server/network/backend_networkmanager_hotspot.go
```diff
@@ -0,0 +1,705 @@
...
+	return paths
+}
+
+func (b *NetworkManagerBackend) getAPCapableWiFiDevice(deviceName string, band string) (*wifiDeviceInfo, error) {
+	if err := validateHotspotBandName(band); err != nil {
+		return nil, err
+	}
+
+	if deviceName != "" {
+		devInfo, ok := b.wifiDevices[deviceName]
```
bbedward: There are a series of locked helpers to prevent runtime panics from concurrent map access, those should be used.

```go
b.wifiDevices[deviceName] -> b.wifiDeviceByIface(deviceName)
for name := range b.wifiDevices -> iterate b.wifiDevicesSnapshot()
wifiDevices[name] -> index the same snapshot
for _, devInfo := range b.wifiDevices -> b.wifiDevicesSnapshot()
(updateHotspotState) - same -> b.wifiDevicesSnapshot()
b.wifiDevices[pendingDevice] -> b.wifiDeviceByIface(pendingDevice)
```

bbedward (changes requested):
The work looks great, thanksa lot for the contribution. Just some small changes on the map access mainly which is better to guard with mutexes

knuxyl:
What advantage does this have over hostapd? Why choose NetworkManager over hostapd?

ronharel02:
> What advantage does this have over hostapd? Why choose NetworkManager over hostapd?

@knuxyl Hotspot support here is backend-agnostic, I implemented it as an optional capability. `NetworkManager` is simply the first implementation because it already manages the device, connection sharing, and state. A `hostapd` implementation could be added later, but it would also require coordinating DHCP/DNS, forwarding, NAT, device ownership, and lifecycle management separately...

ronharel02:
@bbedward Thanks for the review, I fixed the mutex issue :)

PR #2887, opened 2026-07-18
Fix(spotlight): improve file search query restoration and mode switching

Legend-017:
## Description

- **Fix empty results on query restore:** fix(spotlight): fix empty results on restoring last file search query.
- **Proper mode restoration:** Ensures the last selected mode (tab) is correctly restored upon launch when the setting is enabled.
- **File trigger in plugins mode:** Allows the `/` file search trigger in the plugins tab.
- **Clear `/` prefix behavior:** Prevents the launcher from getting stuck in "files" mode when the `/` prefix is deleted, now correctly reverting to the "All" tab or remembered mode(tab).

## Type of change

## Checklist

### DIFF
--- a/quickshell/Modals/DankLauncherV2/Controller.qml
+++ b/quickshell/Modals/DankLauncherV2/Controller.qml
@@ -461,6 +461,10 @@ Item {
             if (explicitType !== null && SessionData.launcherLastFileSearchType !== explicitType) {
                 SessionData.setLauncherLastFileSearchType(explicitType);
             }
+        } else {
+            if (searchMode === "files") {
+                restorePreviousMode();
+            }
         }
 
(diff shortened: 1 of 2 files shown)


bbedward (changes requested):
This creates a regression with queries like this `dms ipc call launcher toggleQuery "vim"` , vim is pre-selected instead of having the cursor at the end as we want. The main use case for that is plugins, so we don't want to pre-select text when its launched with an explicit query

quickshell/Modals/DankLauncherV2/Controller.qml
```diff
@@ -461,6 +461,10 @@ Item {
             if (explicitType !== null && SessionData.launcherLastFileSearchType !== explicitType) {
                 SessionData.setLauncherLastFileSearchType(explicitType);
             }
+        } else {
```
bbedward: Just style wise, this should be an else if. In general I prefer guard statement patterns though.

bbedward (changes requested):

PR #2909, opened 2026-07-22
Add option to ignore app-requested notification timeout

### LINKED ISSUE #2908: Option to ignore app-requested notification timeout (expireTimeout)
**Is your feature request related to a problem? Please describe.**
Notification popups sometimes disappear noticeably faster (or slower) than the durations configured in Settings → Notifications → Timeouts, with no hovering/interaction involved. This is because a sending application's `expire_timeout` hint (`NotificationService.qml`, timer interval logic) currently takes absolute priority over the user's configured `notificationTimeoutLow/Normal/Critical` values whenever the app supplies a non-negative value. That matches the notification spec, but it means the timeout settings in DMS are silently overridden on a per-app basis, which is unexpected from a user's perspective.

**Describe the solution you'd like**
Add a setting (e.g. "Ignore App-Requested Timeout") under Settings → Notifications → Timeouts that, when enabled, always uses the configured Low/Normal/Critical durations regardless of any `expireTimeout` hint from the notifying application. Default stays off, preserving current spec-compliant behavior.

**Describe alternatives you've considered**
Leaving it as-is and documenting the spec behavior — but several apps send short timeouts (e.g. ~2-3s) that make the popup feel like it's "disappearing too fast" compared to the user-configured duration, so a way to override it is preferable to just explaining the spec.

**Additional context**
I have a working implementation and will open a PR against this issue.

tkroll-ionos:
Closes #2908

## Summary
- Adds `notificationIgnoreAppTimeout` setting (default `false`, preserving current spec-compliant behavior) that, when enabled, always uses the configured Low/Normal/Critical durations instead of a notifying app's `expireTimeout` hint.
- Exposes it as a toggle in Settings → Notifications → Timeouts, right below the Critical Priority duration.
- Registers the new key in `SettingsSpec.js` and regenerates `settings_search_index.json`.

## Why
`NotificationService.qml`'s timer-interval logic gives any app-supplied `expireTimeout >= 0` absolute priority over the user's configured timeout. That's spec-compliant, but from a user's perspective popups can seem to vanish faster (or slower) than configured, with zero interaction involved, depending on which app sent the notification. This setting lets users opt into forcing their own durations.

## Test plan
- [x] Ran a local `dms run --config` build with the change; toggle appears under Settings → Notifications → Timeouts and persists via `SettingsData.set`.
- [x] Verified enabling the toggle causes `NotificationService`'s timer interval to fall through to `SettingsData.notificationTimeoutLow/Normal/Critical` instead of the app's `expireTimeout`.
- [x] `python3 quickshell/translations/check_term_freeze.py` — inactive, no issues.
- [x] `python3 quickshell/translations/check_term_variants.py` — no term variants.
- [x] `python3 quickshell/translations/extract_settings_index.py` — regenerated and committed `settings_search_index.json`.

### DIFF
--- a/quickshell/Common/SettingsData.qml
+++ b/quickshell/Common/SettingsData.qml
@@ -916,6 +916,7 @@ Singleton {
     property int notificationTimeoutLow: 5000
     property int notificationTimeoutNormal: 5000
     property int notificationTimeoutCritical: 0
+    property bool notificationIgnoreAppTimeout: false
     property bool notificationCompactMode: false
     property bool notificationShowTimeoutBar: false
     property bool notificationDedupeEnabled: true
--- a/quickshell/Common/settings/SettingsSpec.js
+++ b/quickshell/Common/settings/SettingsSpec.js
@@ -477,6 +477,7 @@ var SPEC = {
     notificationTimeoutLow: { def: 5000 },
     notificationTimeoutNormal: { def: 5000 },
     notificationTimeoutCritical: { def: 0 },
+    notificationIgnoreAppTimeout: { def: false },
     notificationCompactMode: { def: false },
     notificationShowTimeoutBar: { def: false },
     notificationDedupeEnabled: { def: true },
--- a/quickshell/Modules/Settings/NotificationsTab.qml
+++ b/quickshell/Modules/Settings/NotificationsTab.qml
@@ -844,6 +844,15 @@ Item {
                         }
                     }
                 }
+
+                SettingsToggleRow {
+                    settingKey: "notificationIgnoreAppTimeout"
+                    tags: ["notification", "timeout", "expire", "app", "override", "duration"]
+                    text: I18n.tr("Ignore App-Requested Timeout")
+                    description: I18n.tr("Always use the durations above, even if an app requests a shorter or longer one")
+                    checked: SettingsData.notificationIgnoreAppTimeout
+                    onToggled: checked => SettingsData.set("notificationIgnoreAppTimeout", checked)
+                }
             }
 
             SettingsCard {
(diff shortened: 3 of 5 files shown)


bbedward:
This is fine I think, but would it be better to just expand the scope of the entire dms notification daemon to allow per-app or per-notification rules? I'd probably rather go "all out" so to speak than doing a partial solution if that makes more sense

tkroll-ionos:
Thanks for the feedback! I actually dug into the notification rules system (`notificationRules` in `SettingsData.qml`) to see how big a lift the "all out" version would be, and I think there's a good case for merging this small toggle now regardless of what happens with per-app rules later.

A couple of things that came out of that:

- The existing `notificationRules` mechanism already supports per-app policy (`action`, `urgency`) via `desktopEntry`/`appName` matching, with full CRUD helpers and a settings UI. Adding a per-rule `timeout` override there would be a natural, fairly contained extension — I agree it's worth doing.
- But that per-app override would need a fallback for notifications that don't match any rule. Right now, unmatched notifications default to *always* honoring the app's `expireTimeout`, which is exactly the behavior some users want to opt out of globally, independent of any per-app rule they've set up.
- So even with full per-app rules in place, `notificationIgnoreAppTimeout` still earns its place as the base-layer default: it decides what happens when no rule fires, rather than being made redundant by rules that specifically override it.

Given that, I'd suggest merging this toggle as-is now — it's a small, self-contained change with no risk to the rules feature — and treating the per-app timeout override as a separate, additive PR on top of the existing `notificationRules` schema (adding a `timeout` field, consulted before this global setting in the timer logic). Happy to take that follow-up on too if you're on board with the scope split.

PR #2915, opened 2026-07-23
feat(settings): add option to suppress config reload toast (#2730)

### LINKED ISSUE #2730: Add an option to remove the 'config reloaded' info toast
### Feature Description

In Niri whenever you change the Niri config file the  'niri: config reloaded' info toast pops up. That's nice if I edit the config manually, but not if the config is changed programmatically and often, in that case the toast becomes just an annoyance.

### Use Case

I alter the Niri config automatically from scripts to change Niri behaviour based on certain circumstances and this can happen very often.

### Compositor(s)

Niri

### If Other, please specify

might apply to others, I cannot tell

### Proposed Solution

Add a config option to disable it.

_No response_

hthienloc:
## Description

Add a toggle in Settings → Personalization → Niri/Mango to disable the "config reloaded" info toast for Niri and MangoWC. When disabled, the toast is suppressed on both manual config edits and programmatic reloads.

image

## Type of change

## Related issues

Closes #2730

## Checklist

### DIFF
--- a/quickshell/Modules/Settings/CompositorLayoutTab.qml
+++ b/quickshell/Modules/Settings/CompositorLayoutTab.qml
@@ -270,8 +270,8 @@ awk '$1 == "xray" { print FILENAME ":" FNR; exit }' $files 2>/dev/null`;
             SettingsCard {
                 width: parent.width
                 tags: ["compositor", "toast"]
-                title: I18n.tr("Compositor Toasts")
-                settingKey: "compositorToasts"
+                title: I18n.tr("Notifications")
+                settingKey: "compositorNotifications"
                 iconName: "notifications"
                 visible: CompositorService.isNiri || CompositorService.isMango
 
--- a/quickshell/Modules/Settings/CompositorLayoutTab.qml
+++ b/quickshell/Modules/Settings/CompositorLayoutTab.qml
@@ -269,6 +269,24 @@ awk '$1 == "xray" { print FILENAME ":" FNR; exit }' $files 2>/dev/null`;
 
             SettingsCard {
                 width: parent.width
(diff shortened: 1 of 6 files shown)


quickshell/Modules/Settings/CompositorLayoutTab.qml
```diff
@@ -267,6 +267,24 @@ awk '$1 == "xray" { print FILENAME ":" FNR; exit }' $files 2>/dev/null`;
                 }
             }
 
+            SettingsCard {
+                width: parent.width
+                tags: ["compositor", "toast"]
+                title: I18n.tr("Compositor Toasts")
```
bbedward: Can we squish this translation so it re-uses an already existing one, maybe just I18n.tr("Notifications")

bbedward (changes requested):

quickshell/Modules/Settings/CompositorLayoutTab.qml
hthienloc: I thought about that too, but isn't a toast different from a notification? "Notifications" already refers to the system notification settings tab, while this card controls the compositor config reload toasts, which is a different feature. But I can make that change if it makes the translations more consistent.

bbedward: My thinking is that the end state might be Compositor -> Notifications (or Alerts), since its under compositor tab it can already be clear that the settings are related to that.

Then maybe

```
Notification <-> Toast <-> None
Error Only T/F
```

I actually think the default state should probably be to only show loading errors and show it as a normal critical notification thats persistent. But we dont have to reach the end state here, I just want to minimize new translator terms mainly.

PR #2928, opened 2026-07-24
refactor(mpris): enrich metadata across equivalent players

arqueon:
## Summary

- keep the active MPRIS source stable and prefer a controllable player over informational mirrors
- group equivalent playing sources and use the richest peer for stable title, artist, and album metadata
- let `TrackArtService` borrow artwork from an equivalent MPRIS source
- show the stable album value in the native media player

## Why

Browsers and desktop integrations can expose multiple MPRIS players for the same playback. One source may provide working transport controls but incomplete metadata, while another exposes the artist, album, or artwork. Selecting only the first playing entry leaves the native DMS media UI with missing information.

The fix is provider-agnostic: it operates only on MPRIS state and capabilities and adds no site-specific lookup or scraping.

## Impact

DMS remains the owner of player selection and transport. The native media surfaces and any plugin consuming `MprisController` or `TrackArtService` can use richer metadata without switching control to a read-only mirror. Single-source players keep their existing behavior.

## Validation

- Qt 6 `qmllint` passed for all three changed QML files
- live reload completed without new QML errors
- verified with simultaneous controllable, metadata-rich, and read-only MPRIS sources: artist and album remained correct and `TrackArtService.artReadyFor()` returned `true`
- confirmed the patch adds no provider, domain, or HTML-scraping logic

### DIFF
--- a/quickshell/Modules/DankDash/MediaPlayerTab.qml
+++ b/quickshell/Modules/DankDash/MediaPlayerTab.qml
@@ -490,7 +490,7 @@ Item {
                     }
 
                     StyledText {
-                        text: activePlayer?.trackAlbum || ""
+                        text: MprisController.stableAlbum
                         font.pixelSize: Theme.fontSizeSmall
                         color: Theme.surfaceTextSecondary
                         width: parent.width
--- a/quickshell/Services/MprisController.qml
+++ b/quickshell/Services/MprisController.qml
@@ -171,14 +171,26 @@ Singleton {
         return player && player.playbackState === MprisPlaybackState.Stopped && !player.trackTitle && !player.trackArtist;
     }
 
-    function normalizedTrackTitle(player: MprisPlayer): string {
-        return displayTrackTitle(player).toLowerCase();
+    // App-name suffixes that browsers/integrations append as "<title> | <App>".
+    // Only these are stripped for track matching; display always keeps the full title.
+    readonly property var _appTitleSuffixes: ["youtube", "youtube music", "soundcloud", "spotify", "chrome", "chromium", "firefox", "brave", "vivaldi", "twitch"]
+
+    function _stripAppTitleSuffix(title: string): string {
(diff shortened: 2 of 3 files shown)


purian23 (changes requested):
Thanks for the PR! There are just a couple of things to tackle and we'll follow up after with our own addition: 

1. **`_bestPlayingPlayer`: unrelated players can steal the active source.**
   When the active player is playing but not controllable, the function falls through to `controllable[0] || playing[0]` without checking whether that candidate is the same track. Any unrelated controllable player that starts playing will take over the active source. 
Let's only auto-switch to a controllable peer when `isSameTrack(activePlayer, candidate)` is true, or when the active player is idle/stopped, otherwise keep the current active player.

2. **`displayTrackTitle`: keep the strip for matching, not for display.**
   Stripping the trailing `" | …"` suffix currently feeds both track matching *and* the `stableTitle` shown in the bar/overview, so users see shortened titles. 
Let's use the stripped title only for `isSameTrack` matching, and keep the full `trackTitle` for display. If the suffix does need to be hidden in the UI, limit the strip to known app-name suffixes rather than any trailing `" | …"`, which can cut into legitimate titles.

arqueon:
Thanks for the review! Both points are addressed in c81179dd:

1. **Source stealing** — when the active player is playing but not controllable, ownership is now only handed to a controllable peer that passes `isSameTrack(activePlayer, candidate)`; otherwise the current active player is kept. The `controllable[0] || playing[0]` fallback now only applies when the active player is not playing (idle/stopped/paused), as before.

2. **Title stripping** — `displayTrackTitle` now always returns the full `trackTitle`, so the bar/overview show the complete title. The suffix strip only feeds the `isSameTrack` matching key (`normalizedTrackTitle`), and it's limited to a known app-name allowlist (YouTube, SoundCloud, browsers, Twitch…), so legitimate titles containing " | " are never cut.

Issue #1206, opened 2025-12-29
openconnect VPN with PKCS11 authentication


lostinthemath:
## Feature Description

Allow VPN connection with NetworkManager openconnect plugin with PKCS11 authentication, i.e. HSM (Smartcard, YubiKey, etc.)

## Use Case

I opened this as a feature request rather than a bug only due to how niche my situation probably is.

My work uses a Cisco AnyConnect VPN, with authentication via a pin-protected smartcard (user certificate and private key). On my end, I have networkmanager (with the openconnect plugin) configured to use my public x509 certificate and associated private key (through p11kit) to authenticate with the VPN server. The x509 cert doesn't require the smartcard pin, only the private key.

As-is, trying to connect through the DMS GUI brings up the standard username/password prompt rather than prompting for the smartcard PIN. 

I have no way of knowing if this issue extends to other VPN plugins or protocols that use hardware tokens.

## Compositor

Is this feature specific to one compositor?

- [x] All compositors
- [ ] niri
- [ ] Hyprland
- [ ] dwl (MangoWC)
- [ ] sway

## Proposed Solution

I'm not entirely sure how exactly to implement this, however for security, the PIN should never be saved, and especially should never be viewable in plaintext. Ideally, the VPN plugin would recognize the authentication method, and prompt for the PIN rather than a username + password.

## Alternatives/Existing Solutions

Most relevant to adding this feature to DMS, I can connect by using:

    nmcli connection up <vpn name> --ask

I can also connect via `nmtui` (both `nmcli` and `nmtui` kick me back to openconnect in the terminal for PIN entry). Finally, I can connect via the nm-applet in the system tray, though I believe it uses the openconnect plugin's own GUI.

If full integration with DMS is unfeasible, custom connection commands would be a very welcome stopgap. In that case, I would spawn a terminal window running `nmtui-connect <vpn name>`. As-is, I'm stuck running nm-applet or connecting in the terminal. Mildly annoying, at worst.

bbedward:
Could you share the information from nmcli connection show (vpn name) , and maybe a screenshot of nm-applet or nmtui prompt. You can redact anything identifiable I just need to see the network manager connection flags for how it identifies the connection.

lostinthemath:
Sure, but first, I did some testing. A lot of this is probably obvious, but hopefuly some of it is helpful:

[...]

PR #2970, opened 2026-07-31
feat(network): persist NM secrets to D-Bus Secret Service keyring

Mikilio:
#### Problem
PR #2402 introduced keyring retrieval, but I never added the logic to actually store secrets in the keyring or remove them when connections are deleted. So the retrieval path had nothing to retrieve. I only noticed after clearing my old keyring and realizing there was nothing left to retrieve. This PR closes that gap.

#### Changes
- **`secret_service.go`**: added `store()` (CreateItem with `replace=true`), `deleteByUuid()` (search by connection-uuid, then Item.Delete each match), `ensureUnlocked()` (reads the `Locked` property and unlocks the default collection if needed), and `searchItems()` (which now correctly merges locked and unlocked results instead of discarding one set). `trySecretService` return type changed from `nmSettingMap` to `map[string]string`, and the switch statement that restricted keyring lookup to specific setting types was removed.
- **`agent_networkmanager.go`**:
  - Added `formatKeyringSecrets` helper to format raw secrets into NM response format (handles VPN `secrets` dict, 802-1x key filtering).
  - `GetSecrets` now persists secrets to the keyring when the user opts to save (`reply.Save`). Explicit cases for `802-11-wireless-security`, `802-1x`, and `vpn`/`wireguard`. Unknown setting types log a Warn with key names (never values) rather than storing blindly.
  - `DeleteSecrets` and `DeleteSecrets2` now remove keyring items keyed by connection UUID. `DeleteSecrets2`'s `GetSettings` failure log is bumped from silent to Warn.
  - `fieldsNeeded` default returns `hints` instead of `[]string{}` so every NM setting type can participate.
  - **Phase 2 cleanup**: the password-flags decision chain was extracted into `readPasswordFlags` and `handlePasswordFlags` helpers, and field resolution now groups by setting type (vpn infers fields directly; wifi/802-1x delegate to the flag helpers). This also fixed a bug where VPN requests with empty hints always hit the missing-flags error path.
  - **Dispatch point logging**: every setting-type switch now logs when it falls through to generic handling (Warn for response shaping, Debug for field inference, caching, and field metadata), so connection types we don't handle explicitly are easy to spot in logs.
- **Tests**: added `TestDeleteSecretsByConn_EmptyConnUuid`, `TestDeleteSecrets_ExtractsConnUuid`, and `TestDeleteSecrets2_ExtractsSettings` to verify UUID extraction and the nil-conn guard.

#### Current state
Working. Secrets are stored when the user saves, retrieved from the keyring on subsequent connections, and cleaned up when connections are deleted. The keyring unlock dialog continues to appear when needed.

#### Future work
This PR includes some minor refactoring alongside the new functionality. Given permission, I'd like to slowly improve the readability and maintainability of this code over time. The new dispatch-point logging should help surface which connection types future refactors need to account for.

As a side note, may I be tagged on PRs that touch these files just so I can stay informed?

### DIFF
--- a/core/internal/server/network/agent_networkmanager.go
+++ b/core/internal/server/network/agent_networkmanager.go
@@ -171,79 +171,34 @@ func (a *SecretAgent) GetSecrets(
 
 	// Phase 2: Resolve fields from hints or password-flags.
 	if len(fields) == 0 {
-		if settingName == "vpn" {
-			if a.backend != nil {
-				a.backend.stateMutex.RLock()
-				isConnectingVPN := a.backend.state.IsConnectingVPN
-				a.backend.stateMutex.RUnlock()
-
-				if !isConnectingVPN {
-					log.Infof("[SecretAgent] VPN with empty hints - deferring to other agents for %s", vpnSvc)
-					return nil, dbus.NewError("org.freedesktop.NetworkManager.SecretAgent.Error.NoSecrets", nil)
-				}
-
-				fields = inferVPNFields(conn, vpnSvc)
-				log.Infof("[SecretAgent] VPN with empty hints but we're connecting - inferred fields: %v", fields)
-			} else {
+		switch settingName {
+		case "vpn":
+			if a.backend == nil {
 				log.Infof("[SecretAgent] VPN with empty hints - deferring to other agents for %s", vpnSvc)
 				return nil, dbus.NewError("org.freedesktop.NetworkManager.SecretAgent.Error.NoSecrets", nil)
 			}
-		}
 
-		if len(fields) == 0 {
-			const (
-				NM_SETTING_SECRET_FLAG_NONE         = 0
-				NM_SETTING_SECRET_FLAG_AGENT_OWNED  = 1
-				NM_SETTING_SECRET_FLAG_NOT_SAVED    = 2
-				NM_SETTING_SECRET_FLAG_NOT_REQUIRED = 4
-			)
+			a.backend.stateMutex.RLock()
+			isConnectingVPN := a.backend.state.IsConnectingVPN
+			a.backend.stateMutex.RUnlock()
 
-			var passwordFlags uint32 = 0xFFFF
-			switch settingName {
-			case "802-11-wireless-security":
-				if wifiSecSettings, ok := conn["802-11-wireless-security"]; ok {
-					if flagsVariant, ok := wifiSecSettings["psk-flags"]; ok {
-						if pwdFlags, ok := flagsVariant.Value().(uint32); ok {
-							passwordFlags = pwdFlags
-						}
-					}
-				}
-			case "802-1x":
-				if dot1xSettings, ok := conn["802-1x"]; ok {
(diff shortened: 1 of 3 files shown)


Mikilio:
#1206 is making me rethink the UX for this. Reverting to Draft

core/internal/server/network/agent_networkmanager.go
```diff
@@ -966,13 +1080,40 @@ func reasonFromFlags(flags uint32) string {
 	return "required"
 }
 
+func formatKeyringSecrets(settingName string, secrets map[string]string) nmSettingMap {
+	switch settingName {
+	case "vpn", "wireguard":
```
bbedward, line 1085: I think vpn and wireguard shouldnt be grouped here, wireguard can fall back to default case probably it uses `private-key`, `preshared-key`

core/internal/server/network/agent_networkmanager.go
```diff
@@ -178,79 +171,34 @@ func (a *SecretAgent) GetSecrets(
...
-					fields = infer8021xFields(conn)
-				default:
-					log.Warnf("[SecretAgent] Agent-owned secrets for unhandled setting %s (flags=%d)", settingName, passwordFlags)
-					return nil, dbus.NewError("org.freedesktop.NetworkManager.SecretAgent.Error.NoSecrets", nil)
-				}
-				log.Infof("[SecretAgent] Agent-owned secrets, inferred fields: %v", fields)
-			} else if passwordFlags&NM_SETTING_SECRET_FLAG_NOT_SAVED != 0 {
-				log.Infof("[SecretAgent] Secrets not saved, will need to prompt (flags=%d)", passwordFlags)
-				// Fall through — fields remain empty, prompt will be required.
-			} else {
```
bbedward, line 250: Should probably keep the fallback for any cases we dont handle, to defer to other agents

core/internal/server/network/secret_service.go
```diff
@@ -116,47 +151,56 @@ func (s *secretServiceSession) unlock(items []dbus.ObjectPath) error {
...
+	if !locked {
+		return nil
+	}
+
+	log.Debugf("[SecretAgent] Default collection is locked, unlocking...")
+	return s.unlock([]dbus.ObjectPath{dbus.ObjectPath(secretDefaultCollection)})
+}
+
 func (s *secretServiceSession) lookup(connUuid, settingName, settingKey string) string {
+	if err := s.ensureUnlocked(); err != nil {
```
bbedward, line 179: I dont think we should ensureUnlocked every lookup, maybe just if the key exists - which I think we can tell without unlocking the keyring

bbedward (changes requested):

Mikilio:
@bbedward I am sorry this took up your time. This is still going to undergo massive changes, as my previous attempt was still quite sloppy and I had not done enough research. When I mark this as ready for review again the logic will be closer to what happens in nm-applet.

Your feedback was still helpful, however.

PR #2972, opened 2026-07-31
feat(registry): support multiple plugin and theme registries via env vars

### LINKED ISSUE #2763: Support custom plugin and theme registries
### Feature Description

Currently, the registry repository is hardcoded in `core/internal/{plugins,themes}/registry.go`. I propose that we allow users to configure which registries are used for plugins and themes in addition to the official registry.

### Use Case

Personal registries for private plugins, testing without having to build DMS from source, collections of plugins for niche uses, and the ability to pin a registry to a specific revision for Nix users.

### Compositor(s)

All compositors

### If Other, please specify

_No response_

### Proposed Solution

In `settings.json`, there should be a new option, for example:

```jsonc
"registries": [
  // if the array is empty, auto-fill with the official registry
  { "name": "Official", "url": "https://github.com/AvengeMedia/dms-plugin-registry.git" }, // implicitly default branch
  { "name": "Personal", "url": "https://github.com/wongcallum/dms-plugins.git", "branch": "main" },
  { "name": "Pinned",   "url": "https://github.com/wongcallum/dms-plugins.git", "rev": "a1b2c3d" } // commit or tag
]
```

`core/internal/{plugins,themes}/registry.go` should accept a list of registries, which is read from the config. 

We will have to clone each registry into its own cacheDir, which I propose to be subdirectories in `filepath.Join(os.TempDir(), "dms-registries")` instead of the old `dankdots-plugin-registry`.

In the UI, plugins and themes should be aggregated from all configured registries, and to extend upon #1884 there should be an option to filter between registries. If there are multiple visible plugins with the same `id`, they can both be shown with text to indicate which registry the plugin comes from. There is already a warning prompt for 3rd party plugins from the official registry, this could be reused for custom registries. We should also add markers throughout the plugin browsing, installation and management UI to clearly indicate if a plugin is not from the official registry.

In the CLI, the `dms plugins` command should have an additional flag `--registry <name or url>` for all subcommands. Without the flag, `list, search, browse` aggregate across all configured registries, and `install` should pick the first registry in the `registries` array that contains the plugin `id`. We could also add a new `dms registry` management command for adding, listing, and removing registries.

Internally, we will have to track which registry each plugin/theme came from. This is probably the most difficult part. I propose that we give each registry an ID based on a content-independent hash derived from the normalised URL and branch/revision, so that the ID is stable even if the registry is updated. This ID should be used as the clone directory and set as the source of each plugin/theme.

On disk, the `.meta` sidecar files should now be written for all installed plugins (currently only done for monorepo plugins) to have an authoritative `registryId`, alongside with a `registryUrl` and `registryBranch` for updating. To remain backwards-compatible, a missing `.meta` file or `registryId` should be mapped to the official registry.

Themes should have their `registryId` recorded in `settings.json`.

Currently, I just manually symlink individual plugins into `~/.config/DankMaterialShell/plugins`. AFAIK that's the only way to add custom plugins right now.

Alternatively, we could just allow the user to override the default registry url through an environment variable or config option. This is a much simpler code change but is less useful.

louzt:
# feat(registry): support multiple plugin and theme registries via env vars

Closes AvengeMedia/DankMaterialShell#2763.

The plugin and theme registries currently hardcode a single git repo
(`registryRepo` const, cloned into a single cache dir). This blocks the
common case of users who maintain a personal or community registry
alongside the official one. This PR adds multi-registry support driven by
two env vars and keeps the default behavior identical for users who don't
opt in.

## What this PR does

- Both `core/internal/plugins/registry.go` and `core/internal/themes/registry.go`
  now hold `[]RegistryConfig{Name, URL}` instead of a single hardcoded URL.
- `ParseRegistriesFromEnv()` reads `DMS_PLUGIN_REGISTRIES` /
  `DMS_THEME_REGISTRIES` (comma-separated git URLs, trimmed, empty entries
  skipped). Defaults to the official `AvengeMedia/dms-plugin-registry.git`
  when unset, whitespace-only, or all-empty.
- `Update()` iterates over registries, clones/pulls each into its own
  `<cacheBase>/<name>/` subdir, and aggregates results in declaration order.
  **First occurrence of each plugin/theme ID wins** — duplicates from
  later registries are silently dropped, so a personal registry cannot
  shadow the official one by accident.
- `migrateLegacyCache()` runs once per `Update()` to move the pre-refactor
  flat `<base>/plugins/` (and `<base>/themes/`) into the new
  `<base>/official/` layout. Idempotent. Without it, every existing
  user re-clones the official registry on first Update after upgrade.
- `loadPluginsFrom(dir)` / `loadThemesFrom(dir)` extracted from the
  monolithic loader so each registry's cache is read independently.
- `GetThemeSourcePath` / `GetThemeDir` now search all registries (first hit
  wins, fallback to first registry) so multi-registry themes resolve to the
  right path.
- Per-registry errors are wrapped with the registry name
  (`fmt.Errorf("registry %s: %w", cfg.Name, err)`) so triage is one grep away.

## Decisions taken

- **Env vars, not config file.** Matches the existing `DMS_*` override
  pattern (`DMS_SOCKET`, `DMS_DEBUG`); avoids introducing a new on-disk
  format for a feature that may not need persistence yet. A future PR
  could add `~/.config/dms/registries.toml` for users who want named
  registries (`louzt`, `nix-community`) with the same semantics.
- **Auto-naming `r0/r1/...` from index.** Keeps the env var syntax to URLs
  only. If the maintainer prefers explicit names, a one-line parser change
  (`name|url`) is enough — happy to ship that as a follow-up.
- **Declaration order = priority.** First hit wins in both `Update()`
  aggregation and `GetThemeSourcePath` resolution. No `disabled: bool`
  field yet — adding it later is additive (defaults to false).
- **`official` is the only stable name.** When users set
  `DMS_PLUGIN_REGISTRIES=https://github.com/foo/bar.git`, the cache dir is
  `<base>/r0`. The `official` name only applies to the default fallback so
  users can layer a personal registry on top without conflicting with the
  default's stable name.

## Scope boundary

**This PR does NOT:**

- Add a per-user config file (`registries.toml`) — see Decisions, follow-up.
- Add a `disabled` / per-registry priority flag — additive, follow-up.
- Change how plugins/themes are *installed* (still per-plugin JSON /
  theme.json discovery) — only how the *source list* is composed.
- Touch the IPC layer, the widget system, or the settings UI — those are
  separate concerns.

## Validation

```
$ go build ./...
(no output)

$ go test ./internal/plugins/... ./internal/themes/...
ok      github.com/AvengeMedia/DankMaterialShell/core/internal/plugins   0.006s
ok      github.com/AvengeMedia/DankMaterialShell/core/internal/themes    0.004s

$ go vet ./internal/plugins/... ./internal/themes/...
(no output)

$ go test -run TestParseRegistriesFromEnv -v ./internal/plugins/...
=== RUN   TestParseRegistriesFromEnv
=== RUN   TestParseRegistriesFromEnv/defaults_to_official_when_env_unset
=== RUN   TestParseRegistriesFromEnv/defaults_to_official_when_env_is_whitespace
=== RUN   TestParseRegistriesFromEnv/parses_comma-separated_URLs
=== RUN   TestParseRegistriesFromEnv/trims_whitespace_around_URLs
=== RUN   TestParseRegistriesFromEnv/skips_empty_entries
=== RUN   TestParseRegistriesFromEnv/falls_back_to_official_when_all_entries_empty
--- PASS: TestParseRegistriesFromEnv (0.00s)
    --- PASS: TestParseRegistriesFromEnv/defaults_to_official_when_env_unset (0.00s)
    --- PASS: TestParseRegistriesFromEnv/defaults_to_official_when_env_is_whitespace (0.00s)
    --- PASS: TestParseRegistriesFromEnv/parses_comma-separated_URLs (0.00s)
    --- PASS: TestParseRegistriesFromEnv/trims_whitespace_around_URLs (0.00s)
    --- PASS: TestParseRegistriesFromEnv/skips_empty_entries (0.00s)
    --- PASS: TestParseRegistriesFromEnv/falls_back_to_official_when_all_entries_empty (0.00s)
PASS
ok      github.com/AvengeMedia/DankMaterialShell/core/internal/plugins   0.005s
```

Themes have parallel coverage (`TestParseRegistriesFromEnv` with three
subtests: unset, comma-separated, all-empty fallback).

Existing tests for `TestUpdate`, `TestList`, and `TestLoadPlugins` were
updated to the new method signatures; new subtests exercise the
multi-registry path end-to-end with a mock git client:

- `TestUpdate/aggregates_from_multiple_registries` — official + louzt, both contribute
- `TestUpdate/dedupes_by_ID_with_declaration_order_priority` — first registry wins
- `TestUpdate/migrates_legacy_cache_directory` — `<base>/plugins/` → `<base>/official/plugins/`
- `TestUpdate/no-op_migration_when_legacy_cache_absent` — fresh install skips migration
- `TestUpdateMigration/migrates_legacy_themes_cache` — themes parallel coverage

## Diffstat

```
 core/internal/plugins/manager.go        |  17 ++
 core/internal/plugins/registry.go       | 161 ++++++++++++++++++++-----
 core/internal/plugins/registry_test.go  | 212 ++++++++++++++++++++++++++++----
 core/internal/themes/registry.go        | 213 ++++++++++++++++++++++++++++-----
 core/internal/themes/registry_test.go   | 151 ++++++++++++++++++++++++
 5 files changed, 667 insertions(+), 87 deletions(-)
```

Three logical commits on `feat/registry-list-config` (master..HEAD):

1. `86f7550b` — feat(registry): support multiple plugin/theme registries via env
2. `de79ae8f` — feat(registry): dedupe by ID + migrate legacy cache
3. `ed7511ec` — test(themes): cover legacy cache migration (parallel coverage)

Two packages, one logical change. Plugins and themes share no code; this
PR keeps that property (no shared `registries` helper package introduced
yet — easy to extract later if a third caller appears).

## Backwards compatibility

`DMS_PLUGIN_REGISTRIES` and `DMS_THEME_REGISTRIES` unset → single
official registry, identical to the previous behavior. Existing CLI
flags, config keys, and widget bindings are unchanged.

Cache dir moves from `<base>/plugins/` to `<base>/official/plugins/`
on first `Update()` after this PR. The migration helper
(`migrateLegacyCache`) renames the existing flat layout in place; if no
legacy dir is present, it's a no-op. This is invisible for users on a
fresh install and recovers a fast-forward pull for users on an upgrade.

## Why this is the right shape for #2763

Issue #2763 asks for "support custom plugin and theme registries". The
env-var approach:

- Ships in one PR instead of three (config-file, naming, priority).
- Doesn't require a settings-UI change to be useful.
- Lets power users script it (`DMS_PLUGIN_REGISTRIES=$WORK_REGISTRY dms`).
- Composable with future PRs: a config file can override the env var,
  a `disabled` flag is additive, named registries are a parser tweak.

### DIFF
--- a/core/cmd/dms/commands_common.go
+++ b/core/cmd/dms/commands_common.go
@@ -704,6 +704,7 @@ func getCommonCommands() []*cobra.Command {
 		ipcCmd,
 		debugSrvCmd,
 		pluginsCmd,
+		registryCmd,
 		dank16Cmd,
 		brightnessCmd,
 		dpmsCmd,
--- a/core/cmd/dms/commands_registry.go
+++ b/core/cmd/dms/commands_registry.go
@@ -1,0 +1,67 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/AvengeMedia/DankMaterialShell/core/internal/log"
+	"github.com/AvengeMedia/DankMaterialShell/core/internal/registries"
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+)
+
+var registryCmd = &cobra.Command{
+	Use:   "registry",
+	Short: "Manage plugin and theme registries",
+	Long:  "Manage the registries DMS fetches plugins and themes from. The official registry is always active; additional registries can be added by name and git URL.",
+}
+
+var registryListCmd = &cobra.Command{
+	Use:   "list",
+	Short: "List configured registries",
+	Run: func(cmd *cobra.Command, args []string) {
+		for _, s := range registries.Load(afero.NewOsFs()) {
+			suffix := ""
+			if s.Official() {
+				suffix = " (official)"
+			}
+			fmt.Printf("%s%s\n    %s\n", s.Name, suffix, s.URL)
+		}
+	},
+}
+
+var registryAddCmd = &cobra.Command{
+	Use:   "add <name> <url>",
+	Short: "Add a registry",
+	Long:  "Add a registry by name and git URL. The repository must contain a plugins/ or themes/ directory in the registry format.",
+	Args:  cobra.ExactArgs(2),
+	Run: func(cmd *cobra.Command, args []string) {
+		if err := registries.Add(afero.NewOsFs(), args[0], args[1]); err != nil {
+			log.Fatalf("Error adding registry: %v", err)
+		}
+		fmt.Printf("Registry added: %s\n", args[0])
+	},
+}
+
+var registryRemoveCmd = &cobra.Command{
+	Use:   "remove <name>",
+	Short: "Remove a registry",
+	Args:  cobra.ExactArgs(1),
+	ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		if len(args) != 0 {
+			return nil, cobra.ShellCompDirectiveNoFileComp
+		}
+		var names []string
+		for _, s := range registries.Load(afero.NewOsFs()) {
+			if !s.Official() {
+				names = append(names, s.Name)
+			}
+		}
+		return names, cobra.ShellCompDirectiveNoFileComp
+	},
+	Run: func(cmd *cobra.Command, args []string) {
+		if err := registries.Remove(afero.NewOsFs(), args[0]); err != nil {
+			log.Fatalf("Error removing registry: %v", err)
+		}
+		fmt.Printf("Registry removed: %s\n", args[0])
+	},
+}
--- a/core/cmd/dms/main.go
+++ b/core/cmd/dms/main.go
@@ -16,6 +16,7 @@ func init() {
 	setupCmd.AddCommand(setupBindsCmd, setupLayoutCmd, setupColorsCmd, setupAlttabCmd, setupOutputsCmd, setupCursorCmd, setupWindowrulesCmd)
 	updateCmd.AddCommand(updateCheckCmd)
 	pluginsCmd.AddCommand(pluginsBrowseCmd, pluginsListCmd, pluginsInstallCmd, pluginsUninstallCmd, pluginsUpdateCmd)
+	registryCmd.AddCommand(registryListCmd, registryAddCmd, registryRemoveCmd)
 	rootCmd.AddCommand(getCommonCommands()...)
 
 	rootCmd.AddCommand(authCmd)
--- a/core/internal/plugins/registry.go
+++ b/core/internal/plugins/registry.go
@@ -11,7 +11,49 @@ import (
 	"github.com/spf13/afero"
 )
 
-const registryRepo = "https://github.com/AvengeMedia/dms-plugin-registry.git"
+const defaultRegistryURL = "https://github.com/AvengeMedia/dms-plugin-registry.git"
+
+// envRegistriesKey is the env var consulted at registry construction time.
+// Comma-separated list of git URLs. Empty/unset → defaultRegistryURL.
+const envRegistriesKey = "DMS_PLUGIN_REGISTRIES"
+
+// RegistryConfig identifies a single registry source.
+type RegistryConfig struct {
+	// Name is a short identifier used for the per-registry cache subdir
+	// (e.g. "official", "louzt"). Defaults to "r<N>" when only URL given.
+	Name string
+	// URL is the git URL of the registry repository.
+	URL string
+}
+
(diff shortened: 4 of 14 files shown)


bbedward:
I expanded on this to have CLI + GUI integration, since it was a 1.6 milestone issue

louzt:
Good work wiring up those @bbedward. Glad it’s merged for 1.6.

PR #3010, opened 2026-08-07
feat(settings): add location search to theme & gamma coordinates

Curious-r:
## Description

Adds a searchable location picker to the automatic location sections of Theme & Colors (Color Mode) and Gamma Control, so users can search for a city and auto-fill the latitude/longitude coordinates (like the weather setting do) instead of typing them by hand.

Extracts the previously duplicated "Use IP Location + manual coordinates" UI into a new shared `SettingsLocationSection` component used by both tabs. The existing weather location search is left unchanged.

Tested on [hardware](https://github.com/Curious-r/nix-config/commit/f9f1cf8bb520c203b1efc060044b82c799fe6c69): search → auto-fill coordinates, cross-tab sync (both tabs share the same coordinates), and the "Share Gamma Control Settings" toggle all work as before. No backend changes. 

Follow-ups: While working on this, I spotted two minor things to address later: remembering and displaying the user's last selection in the search field, and fixing a visual glitch where the error icon flashes post-selection. Since both touch the shared DankLocationSearch component, I'll handle them in a separate PR over at dank-qml-common(maybe later).

## Type of change

- [ ] Documentation
- [ ] Other

## Checklist

### DIFF
--- a/quickshell/Modules/Settings/GammaControlTab.qml
+++ b/quickshell/Modules/Settings/GammaControlTab.qml
@@ -320,100 +320,10 @@ Item {
                             }
                         }
 
-                        Column {
-                            property bool isLocationMode: SessionData.nightModeAutoMode === "location"
-                            visible: isLocationMode
-                            spacing: Theme.spacingM
+                        SettingsLocationSection {
                             width: parent.width
-
-                            DankToggle {
-                                id: ipLocationToggle
-                                width: parent.width
-                                text: I18n.tr("Use IP Location")
-                                description: I18n.tr("Automatically detect location based on IP address")
-                                checked: SessionData.nightModeUseIPLocation || false
-                                onToggled: checked => {
-                                    SessionData.setNightModeUseIPLocation(checked);
-                                }
-
-                                Connections {
-                                    target: SessionData
-                                    function onNightModeUseIPLocationChanged() {
-                                        ipLocationToggle.checked = SessionData.nightModeUseIPLocation;
-                                    }
-                                }
-                            }
-
-                            Column {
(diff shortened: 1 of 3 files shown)


purian23 (changes requested):
Thanks for the PR, and good idea! 

One ask before merge: after the lat/lon fields are edited, the `SessionData` binding can break so search/cross-tab updates don’t refresh the visible text. Matching weather’s pattern (set `.text` on select + `Connections`) in `SettingsLocationSection` would fix that.

Happy to re-review once that’s in and the `dank-qml-common` follow-ups are fine separately.

Curious-r:
Hi @Purian23 , thanks for your time!
Ready for re-review — addressed in `c33fe031`.

`SettingsLocationSection` now follows the weather tab pattern: the coordinate fields initialize from SessionData in `Component.onCompleted`, refresh through `Connections` on SessionData changes, and `onLocationSelected` sets `.text` explicitly, so manual edits no longer break the visible text when search results or other tabs update the values.

[Tested on hardware](https://github.com/Curious-r/nix-config/commit/88574964958f3c450e36bcb5463a8635ac50b939):

- Manually editing lat/lon applies immediately (the day/night switch reacts as expected).
- After a manual edit, re-searching and re-selecting the same city refreshes the coordinate fields and re-applies the location.
- Cross-tab sync works in both directions (Gamma ⇄ Theme), including after manual edits.

One more thing: the "Share Gamma Control Settings" toggle only affects the time schedule — coordinates and the IP-location flag are always shared by design (there are no theme-specific coordinate fields). I kept that behavior unchanged.

PR #3020, opened 2026-08-08
feat(theme): add a selectable source color mode for wallpaper palettes

Conava:
## Description

DMS builds your whole color scheme out of a single color picked from the wallpaper. Today that's always the color covering the most pixels. Often this is a background color and not the color of the main subject of the image, for example a blue background from the sky. This can lead to many images resulting in a blue color palette.

There's no way to change that pick. This PR adds a **Source Color** setting in Theme & Colors that lets the user decide. DMS hands the picked color to [matugen](https://github.com/InioX/matugen), the external tool that generates the actual scheme. Five of the options below are matugen's own, Colorful is not.

| Option | Picks |
|---|---|
| **Dominant** | The most common color. Current behavior, and still the default |
| **Colorful** | The most colorful prominent color, rather than the largest one |
| Darkest / Lightest | The darkest or lightest of matugen's candidates |
| Most Saturated / Least Saturated | Most or least saturated |
| Most Vivid | The boldest of matugen's candidates |

On one of the wallpapers below that's the difference between a `#8ad0ee` cold cyan accent and a `#ffb690` warm peach one, from the same image.

The five matugen options map to its `--prefer` flag, through an exact-match allowlist so a hand-edited `settings.json` can't inject arbitrary arguments. Colorful has no matugen equivalent, so DMS computes it: the image is reduced to 128 representative colors, then each is scored on how much of the image it covers against how colorful it is. That scoring is inspired by [caelestia-cli](https://github.com/caelestia-dots/cli), which produces noticeably warmer palettes than DMS on the same images. Either way matugen receives a single color, so nothing downstream changes.

### Notes for review

- **Nothing changes for existing users.** Dominant is the default and produces byte-identical output to before. The flag isn't even added to matugen's command line unless the user picks something else
- **It can't break theming.** If the image can't be read (DMS accepts jxl/avif/heif/exr, which Go can't decode), it logs a warning and falls back to today's behavior
- `--prefer` arrived in matugen 4.1. On older versions those five options fall back to Dominant instead of failing on an unknown flag. Colorful works on any matugen version, since matugen only ever sees a single color
- One new Go dependency, github.com/Nadim147c/material/v3, used only by core/internal/matugen/sourcecolor.go, for the color-space math and image quantization Colorful needs. Apache-2.0, and its own go.mod has no requires at all, so it pulls in nothing else.

**Why this doesn't just call the library's own image quantizer:** its `QuantizeCelebi` is broken: it stops after one unconverged pass, and its distance function returns a dot product rather than a distance, so "nearest color" is meaningless. On a 3840x2160 wallpaper it grouped 7,979,139 of 8,292,604 pixels into a single cluster. This PR does that step itself instead, with the reasoning in a comment at the call site.

## Type of change

- [ ] Documentation
- [ ] Other

## Related issues

None.

## Screenshots / video

screenshot_2026-08-08_18-15-20
screenshot_2026-08-08_18-15-08

## Validation

54 theme generations through the built binary against a scratch config, matugen 4.1.0.

**What each option produces**, across six wallpapers. Accent color, dark mode, `scheme-tonal-spot`:

| Wallpaper | Dominant | Colorful | Darkest | Lightest | Most Sat. | Least Sat. | Most Vivid |
|---|---|---|---|---|---|---|---|
| 1 | `#8ad0ee` | `#ffb690` | `#8ad0ee` | `#ffb690` | `#ffb690` | `#95cdf7` | `#8ad0ee` |
| 2 | `#81d3df` | `#faba73` | `#cdca75` | `#fdb976` | `#fdb976` | `#91d5ac` | `#cdca75` |
| 3 | `#edb4eb` | `#bec2ff` | `#edb4eb` | `#90cef4` | `#b8c4ff` | `#edb4eb` | `#edb4eb` |
| 4 | `#b9c3ff` | `#bdc2ff` | `#b9c3ff` | `#9acbfa` | `#9acbfa` | `#f5b2e1` | `#b9c3ff` |
| 5 | `#edc06c` | `#e7c26c` | `#edc06c` | `#89d6b8` | `#edc06c` | `#a2d399` | `#a2d399` |
| 6 | `#feb877` | `#fcb974` | `#ffb692` | `#f1be6d` | `#ffb692` | `#f1be6d` | `#ffb692` |

Every option produced a complete palette on every wallpaper, nothing failed or came back empty. The seven collapse to 4-5 distinct results per image, because matugen's candidate list is short and several `--prefer` values land on the same candidate.

Wallpapers 1 and 2 are the case this PR is for: Dominant picks a cold cyan, Colorful a warm peach or amber. On 4 to 6 nothing much moves. Across my own wallpaper collection of 207 images, which is whatever I happen to have rather than a set curated for this, the median hue shift between Dominant and Colorful is 7° on a 0-180° scale, with 14 wallpapers above 160°. It's a targeted fix for a specific failure, not a global repaint.

Which `--prefer` value lands closest to Colorful also changes per wallpaper (lightness, lightness, saturation, darkness, saturation, lightness), so it works as a per-image knob, not as one setting the user leaves alone.

**The default is inert.** Five generations of one wallpaper, differing only in the Source Color setting, all produced a byte-identical colors file (sha256 `d615235985ca`): setting absent, `dominant`, `""`, `nonsense`, and `--exec`. The last two confirm a hand-edited settings value can't inject arguments into matugen's command line.

**Deterministic.** The same image always produces the same color. This matters because DMS compares generated colors to decide whether to re-theme; a color that drifted between runs would re-theme the desktop on every wallpaper event.

Commands run:

```
go test ./...                             # 2169 passed, 88 packages
make fmt && go mod tidy                   # clean
GOOS=freebsd GOARCH=amd64 go build ./...  # clean
make lint-qml                             # PASS (2 entrypoints)
make build
prek run --files <changed files>
```

## Checklist

- [x] I have opened a corresponding pull request in dlx-docs to document any new behaviors: https://github.com/AvengeMedia/DankLinux-Docs/pull/129


quickshell/Common/SettingsData.qml
```diff
@@ -2599,6 +2600,16 @@ Singleton {
...
 
+    function setMatugenSourceMode(mode) {
+        var normalized = mode || "dominant";
+        if (matugenSourceMode === normalized)
+            return;
+        set("matugenSourceMode", normalized);
+        if (typeof Theme !== "undefined") {
+            Theme.generateSystemThemesFromCurrentTheme();
+        }
+    }
```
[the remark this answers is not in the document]
Conava: Explicit call redundant, removed.

core/internal/matugen/matugen.go
```diff
@@ -301,6 +303,22 @@ func buildOnce(opts *Options) (bool, error) {
 	var dank16JSON string
 	var importArgs []string
 
+	// Colorful mode resolves the seed here, before matugen is invoked at all,
+	// by rewriting the source to the extracted hex. Both the dry-run and the
```
bbedward, line 352: This does break some downstream templates if they use the matugen key {{image}} , the builtin pywalfox one does and some other user-templates may. With matugen though we should be able to inject the wallpaper path via a custom keyword, similar to how we do with dank16 to keep those working.

quickshell/Common/Theme.qml
```diff
@@ -372,6 +372,45 @@ Singleton {
         return schemes[0];
     }
 
+    readonly property var availableSourceModes: [({
+                "value": "dominant",
+                "label": I18n.tr("Dominant", "matugen source color option"),
+                "description": I18n.tr("Uses the wallpaper's most dominant color as the seed (default).")
```
bbedward: I'd like to cut down the amount of terms here, this adds ~16 translator terms for the dropdown

bbedward (changes requested):

core/internal/matugen/matugen.go
Conava: Confirmed, fixed in [146cbc2](https://github.com/AvengeMedia/DankMaterialShell/pull/3020/commits/146cbc21f3965b8a80ae3a9a2533806401747636).

Colorful calls matugen with a hex source, so matugen never sees the wallpaper and leaves {{image}} unset. The builtin pywalfox template then writes "wallpaper": "Null", and any user template using the keyword does the same. Only colorful is affected: the --prefer modes still pass the image through.

I used matugen's own image key in the existing --import-json-string rather than a new keyword, so templates that already use {{image}} keep working without an edit. The path is made absolute, since that is what matugen renders for {{image}} in the other modes.

Verified end to end against the pywalfox template and a user template, matugen 4.1.0:

before: "wallpaper": "Null"
after:  "wallpaper": "/path/to/wall.png"

The generated palette is byte-identical before and after.

quickshell/Common/Theme.qml
Conava: Fixed in [029b3ca](https://github.com/AvengeMedia/DankMaterialShell/pull/3020/commits/029b3ca1ef6482aa135ba8d9ab15ce6db569caa9)

Cut to 8 new terms, down from 15.

I dropped the per-mode descriptions. Five of the seven differed by a single adjective, and the dropdown row's own description already says what the setting does. The docs PR has a table with each mode and the matugen flag behind it, which should be sufficient if anyone needs those details.

What is left is six one- or two-word labels plus the row title and description.
"Colorful" already exists and does not need to be translated, it just appends the translation comment.

If you want we could lower the count one more by replacing "Least Saturated" with "Muted", which is already a term. I left it alone because it breaks the pair with "Most Saturated" and would need the docs table changed as well.

real-or-random:
This seems a nice addition. If we do this, can we also do something much simpler, namely give the user a color picker to set a fixed color? Though this probably shouldn't be in the "auto" tab.

Sorry for sneaking in a feature request here. I can open a separate issue if desired.

Conava:
> This seems a nice addition. If we do this, can we also do something much simpler, namely give the user a color picker to set a fixed color? Though this probably shouldn't be in the "auto" tab.
> 
> Sorry for sneaking in a feature request here. I can open a separate issue if desired.

There already is a very similar feature. You can choose between having your color theme generated automatically (the behavior this PR extends), Custom, Browse (do download custom presets) and Generic, which already has I believe 8 or 10 colors to pick from. You might be happy with creating a custom color scheme or implementing a color picker as an option for the generic mode.
I often see this where instead of 8 available colors we have 7 and one rainbow icon and when you click on that a color picker opens. I would support this change, but it belongs in its own issue/PR. @real-or-random

real-or-random:
> You can choose between having your color theme generated automatically (the behavior this PR extends), Custom, Browse (do download custom presets) and Generic, which already has I believe 8 or 10 colors to pick from. You might be happy with creating a custom color scheme or implementing a color picker as an option for the generic mode.

Yes, I was aware. I opened #3206 as a proper feature request.

purian23:
Hey @Conava, we plan on pushing out a stable release tomorrow if you want to get this rebased and merged. I think it's a solid feature folks will enjoy!

Conava:
> Hey @Conava, we plan on pushing out a stable release tomorrow if you want to get this rebased and merged. I think it's a solid feature folks will enjoy! 

Yes for sure, that was my goal :)
What do I need to do, just rebase now?

PR #3023, opened 2026-08-08
feat(dock): cycle grouped app windows

### LINKED ISSUE #2035: Feat(dock): Cycle through app instances with sequential clicks on group icon
### Feature Description

Currently the Dock navigates instantly to a window for a non-grouped indicator and displays a context window when selecting a multi-instance/grouped indicator (when "Group by App" is enabled). It would enhance usability and consistency by extending app group indicators to navigate to the closest instance of the given app and cycle through the rest of the instances with subsequent clicks.

### Use Case

The current behavior produces inconsistent results: some dock indicators directly facilitate navigation, whereas others prompt a context menu. The context menu is potentially useful but requires more of a tedious user interaction. The click-to-cycle user interaction pattern is established on other platforms and need not detract from existing functionality.

### Compositor(s)

All compositors

### If Other, please specify

_No response_

### Proposed Solution

Sequential clicks on a grouped dock indicator should navigate through all instances of a given app and cycle indefinitely. The context menu trigger could either be retained or reverted to a right-click alone for consistency with the rest of the indicators.

The Niri overview is pretty cool, but it is not a direct replacement. I would love a macOS App Exposé-like app navigation feature, but I'm not sure where that would fit in the stack of DMS/QuickShell/Niri.

hthienloc:
## Description

Grouped dock icons currently open the context menu on left click when multiple windows are present, while right click already opens the same menu. This makes left and right click overlap in behavior.

Change left click for grouped apps to cycle through the app's windows instead, matching the Running Apps widget behavior and common dock interaction patterns.

## Type of change

- [x] Breaking change (fix or feature that changes existing behavior)
- [ ] Documentation
- [ ] Other

## Related issues

Closes #2035

### DIFF
--- a/quickshell/Modules/Dock/DockAppButton.qml
+++ b/quickshell/Modules/Dock/DockAppButton.qml
@@ -168,6 +168,25 @@ Item {
         Qt.callLater(() => waylandToplevel.activate());
         return true;
     }
+
+    function cycleGroupedToplevels() {
+        const toplevels = getGroupedToplevels();
+        if (toplevels.length === 0)
+            return;
+
+        let currentIndex = -1;
+        for (let i = 0; i < toplevels.length; i++) {
+            if (toplevels[i].activated) {
+                currentIndex = i;
+                break;
+            }
+        }
+
+        const nextToplevel = toplevels[(currentIndex + 1) % toplevels.length];
+        if (restoreSpecialWorkspaceWindow(nextToplevel))
+            return;
+        nextToplevel.activate();
+    }
     onIsHoveredChanged: {
         if (mouseArea.pressed || dragging)
             return;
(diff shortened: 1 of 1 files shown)


korbash:
With this configuration:

showDock: false
showDockOnOverview: true

the dock used to be hidden normally, but still appear in Overview mode (niri). After this change, the dock no longer appears in Overview either.

hthienloc:
@korbash I think the regression might come from this PR instead: https://github.com/AvengeMedia/DankMaterialShell/pull/3014. It skips creating the dock layer when the dock is disabled, which may affect the "show dock on overview" option.

I'll take a look. Thanks for the report.

korbash:
I definitely need more time to make this PR clearer.

PR #3045, opened 2026-08-11
fix(notifications): do not show battery notification when already charging

marcinjahn:
## Description

I had a case where:

1. Laptop wasn't charging, it was working on battery
2. Battery fell to 20%
3. I got the notification about battery being low, asking me to charge (good)
4. I connected a charger
5. After short time I got the same notification again (bad)

This should fix it.

## Type of change

- [ ] Documentation
- [ ] Other

## Checklist

- [ ] I have tested my changes locally

### DIFF
--- a/quickshell/Services/BatteryService.qml
+++ b/quickshell/Services/BatteryService.qml
@@ -131,6 +131,16 @@ Singleton {
     property bool _hasNotifiedCriticalBattery: false
     property bool _hasNotifiedChargeLimit: false
 
+    // AC plug detection (onBattery) can precede the battery device reporting
+    // Charging by a few seconds. Timestamp plug-ins so alerts can be
+    // suppressed only for a short grace period instead of whenever on AC.
+    property var _lastPluggedInTime: 0
+    readonly property int _pluggedInGracePeriod: 5000
+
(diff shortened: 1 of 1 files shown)


purian23:
The race diagnosis makes sense, though this introduces a five-second grace window around the AC transition. We generally prefer reacting to the underlying power/UPower state changes rather than relying on timing where possible. Is the grace period needed because there isn’t a reliable follow-up event we can key off here, or could this be resolved when the relevant battery/charging state update arrives instead? 

I've personally not ran into this one before. Once I plug in AC charging the status is satisfied.

marcinjahn:
> The race diagnosis makes sense, though this introduces a five-second grace window around the AC transition. We generally prefer reacting to the underlying power/UPower state changes rather than relying on timing where possible. Is the grace period needed because there isn’t a reliable follow-up event we can key off here, or could this be resolved when the relevant battery/charging state update arrives instead?
> 
> I've personally not ran into this one before. Once I plug in AC charging the status is satisfied.

It happened to me just once, I think it might be difficult to reproduce consistently. I tried solving it the way I did, because I didn’t know of a better way.
And yes, it is hacky, up to you whether it’s worth merging in.

PR #3078, opened 2026-08-15
feat(lock): run face auth concurrently with the password field

RakaiSeto:
## Summary

With a concurrent face module (`pam_gaze_grosshack.so`) the PAM password prompt fires while the face scan is still running — but the lock screen swallowed all input during `passwd.active` and auto-responded with the buffer the instant the prompt arrived, so typing a password during the scan was impossible and partial text was submitted as a password.

This PR changes the lock screen so the scan never interferes with the password input:

- **Input stays live during `passwd.active`** (`handleKey` + `imeCommitSink`): the buffer keeps accepting keystrokes for the whole scan.
- **Respond on explicit Enter only** (`Pam.qml` `submitPassword`): the auto-respond in `onResponseRequiredChanged` is gone. A typed password is submitted intact when the scan settles; if the face matched first, it's discarded with the unlock.
- **Empty Enter during an attempt is a no-op** — a reflex Enter can't kill a face match in progress.
- **Esc aborts the attempt** (`cancelActiveAuth`) without clearing the buffer.
- **Response phase is untimed**: the 15 s stall guard covers the scan phase only; once the password module asks, the stack waits on the user's Enter indefinitely (Esc is the escape). The guard re-arms after a response so a hung module can't hold the stack forever.
- **Polkit modal**: keeps input typed while the scan runs (it already deferred submits until `responseRequired`).
- **Lock feedback line**: Gaze's in-process phase strings now render in state colors (waiting = primary, matched = success, failed = error) instead of blanket error red.

Second commit: `gaze-observe` helper + `AuthHudPill` — a top-center pill for terminal sudo/polkit prompts that DMS can't see in-process, fed by the daemon's scrubbed `auth_phase` broadcast ([Gaze PR](https://github.com/GunduLabs/gaze/pull/491)). Elevation surfaces only; hides on `idle`; self-healing observer. The helper is a new Go binary in the existing module (godbus already a dependency) — no vendorHash change.

## Behaviour

- Enter with the right password during the scan → unlocks, face flow dropped.
- Face match during typing → unlocks, pending password discarded.
- Empty Enter → nothing; Esc → abort, buffer kept.

## Testing

- `go test ./internal/qmlchecks/` passes.
- Go helper: `go build` + `go vet` clean.
- QML reviewed manually; no qmllint available in this environment.
- Needs a real session for the interaction matrix (type-during-scan, Esc, wrong-password-during-race, faillock) and the pill's layer placement.

Depends on the Gaze gate from [GunduLabs/gaze#491](https://github.com/GunduLabs/gaze/pull/491) for the actual race; the input semantics ship standalone with the plain module.

### DIFF
--- a/core/cmd/gaze-observe/main.go
+++ b/core/cmd/gaze-observe/main.go
@@ -1,0 +1,139 @@
+// gaze-observe subscribes to the Gaze daemon's scrubbed auth-phase broadcast
+// (com.gundulabs.Gaze.AuthPhase) and prints one key=value record per change
+// to stdout. It is the read-only status channel for the shell's auth HUD:
+// the daemon fan-out is uid-checked on registration, and the payload contains
+// phase and camera status codes plus the surface class — never face names or
+// match scores.
+//
+// Protocol (stdout, one record per line, fields space-separated):
+//
+//	ready=1                                   after successful registration
+//	phase=waiting rgb=no-face ir=unused surface=elevation
+//	phase=matched rgb=ready ir=unused surface=elevation
+//	phase=not-recognized rgb=no-face ir=unused surface=elevation
+//
+// phase values: waiting | matched | not-recognized | unavailable | idle
+// rgb/ir values: unused | no-face | too-dark | clipped | not-centered |
+//
+//	too-far | too-close | ready | usable
+//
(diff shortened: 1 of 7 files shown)


RakaiSeto:
Closing — opened without authorization.

Issue #2060, opened 2026-03-23
IPC latency is very high


AveryanAlex:
### Compositor

Niri

### Distribution

NixOS

### Select your Installation Method

Distro Packaging

### Was this your original Installation method?

Yes

### dms doctor -vC

<details>
<summary>Click to expand</summary>
## DMS Doctor Report

**System**
- [ok] Operating System: NixOS 26.05 (Yarara)
  - Supported for runtime (install via NixOS module or Flake)
- [ok] Architecture: amd64
- [ok] Display Server: Wayland
  - WAYLAND_DISPLAY=wayland-1

**Versions**
- [ok] DMS CLI: dms 1.4.4+date=2026-03-20_1f64bb8
  - /nix/store/srrs2h9rvaibfvx7mr4ahv0c12shvdsg-dms-shell-1.4.4+date=2026-03-20_1f64bb8/bin/.dms-wrapped
- [ok] Quickshell: quickshell 0.2.1, revision 41828c4180fb921df7992a5405f5ff05d2ac2fff, distributed by: Official-Nix-Flake
  - /etc/profiles/per-user/alex/bin/qs
- [ok] DMS Shell: 1.4.4+date=2026-03-20_1f64bb8
  - /nix/store/srrs2h9rvaibfvx7mr4ahv0c12shvdsg-dms-shell-1.4.4+date=2026-03-20_1f64bb8/share/quickshell/dms

**Installation**
- [ok] DMS Configuration: Found
  - /nix/store/srrs2h9rvaibfvx7mr4ahv0c12shvdsg-dms-shell-1.4.4+date=2026-03-20_1f64bb8/share/quickshell/dms
- [ok] shell.qml: Present
  - /nix/store/srrs2h9rvaibfvx7mr4ahv0c12shvdsg-dms-shell-1.4.4+date=2026-03-20_1f64bb8/share/quickshell/dms/shell.qml
- [info] Install Type: Nix store
  - /nix/store/srrs2h9rvaibfvx7mr4ahv0c12shvdsg-dms-shell-1.4.4+date=2026-03-20_1f64bb8/share/quickshell/dms

**Compositor**
- [ok] niri: niri unstable 2026-03-10 (commit b07bde3ee82dd73115e6b949e4f3f63695da35ea)
  - /run/current-system/sw/bin/niri
- [info] Active: niri

**Quickshell Features**
- [ok] Polkit: Available
  - Authentication prompts
[...]

PR #3081, opened 2026-08-15
perf(ipc): call Quickshell directly

hthienloc:
## Description

Reduce `dms ipc` latency by calling the running Quickshell instance directly over its Unix IPC socket instead of spawning a `qs` process for every command.

This keeps the existing `dms ipc call <target> <function>` syntax and falls back to the existing path when no running session is available.

Benchmarking `spotlight toggle` locally reduced latency from approximately `336 ms ± 130ms` to `210.5 ms ± 88.7ms` by removing the intermediate `qs` process.

A separate lightweight IPC binary could reduce startup latency further, but is intentionally left out of this PR to preserve the existing command interface and keep the change focused.

## Type of change

## Related issues

#2060

## Screenshots / video

Not applicable.

## Checklist

### DIFF
--- a/core/internal/qsipc/client.go
+++ b/core/internal/qsipc/client.go
@@ -27,9 +27,9 @@ const (
 )
 
 var (
-	ErrTargetNotFound      = errors.New("Quickshell IPC target not found")
-	ErrFunctionNotFound    = errors.New("Quickshell IPC function not found")
-	ErrArgumentMismatch    = errors.New("Quickshell IPC argument mismatch")
+	ErrTargetNotFound      = errors.New("quickshell IPC target not found")
+	ErrFunctionNotFound    = errors.New("quickshell IPC function not found")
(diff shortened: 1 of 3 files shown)


core/internal/qsipc/client.go
```diff
@@ -0,0 +1,132 @@
...
+	"net"
+	"os"
+	"path/filepath"
+	"time"
+	"unicode/utf16"
+)
+
+const (
+	commandStringCall = 3
+	responseCompleted = 5
```
bbedward: We s hould add support for indexed responses 2 (target not found), 3 (function not found), 4 (argument mismatch)

bbedward: And maybe add a comment just with a link to where they come from in case they ever change in QS https://github.com/quickshell-mirror/quickshell/blob/59e9c47b0eb48a9e4bcf9631fa062ee939bd2e83/src/io/ipccomm.cpp#L126

bbedward (changes requested):

PR #3082, opened 2026-08-15
control-center: route Bluetooth power toggles through DMS backend

alamin147:
## Description

The Control Center previously toggled `BluetoothService.adapter.enabled` directly, bypassing DMS’s Bluetooth backend.

This change routes Bluetooth power toggles through `bluetooth.setPowered` whenever the DMS Bluetooth capability is available. The backend already attempts soft-rfkill unblocking before enabling the BlueZ adapter, making recovery automatic and invisible to the user.

If the backend is unavailable, it falls back to the Quickshell Bluetooth adapter.

## What changed

- Added `BluetoothService.setBluetoothEnabled()`
- Routed the existing Control Center toggle through the DMS backend
- Logs backend failures
- Preserves the direct adapter fallback
- Adds no prompts, external commands, or visual changes

## Testing

- `make lint-qml` passes
- `git diff --check` passes
- Quickshell configuration loads without new errors
- Tested in the active DMS installation

### DIFF
--- a/quickshell/Modules/ControlCenter/Components/DragDropGrid.qml
+++ b/quickshell/Modules/ControlCenter/Components/DragDropGrid.qml
@@ -191,9 +191,7 @@ Column {
             }
         case "bluetooth":
             {
-                if (BluetoothService.available && BluetoothService.adapter) {
-                    BluetoothService.adapter.enabled = !BluetoothService.adapter.enabled;
-                }
+                BluetoothService.setBluetoothEnabled(!BluetoothService.enabled);
                 break;
             }
         case "audioOutput":
--- a/quickshell/Services/BluetoothService.qml
+++ b/quickshell/Services/BluetoothService.qml
@@ -11,6 +11,7 @@ import qs.Services
 Singleton {
     id: root
 
+    readonly property var log: Log.scoped("BluetoothService")
     readonly property BluetoothAdapter adapter: Bluetooth.defaultAdapter
     readonly property bool available: adapter !== null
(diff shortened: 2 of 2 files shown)


bbedward:
We have soft rfkill unblocking built in, it can be instantiated when toggling device maybe instead

```js
function setBluetoothEnabled(enabled) {
    if (DMSService.isConnected && DMSService.capabilities.includes("bluetooth")) {
        DMSService.sendRequest("bluetooth.setPowered", { powered: enabled }, ...)
        return
    }
    if (adapter)
        adapter.enabled = enabled
}
```

It may require udev rules since its unprivileged (or just being in bluetooth group on some distros), but maybe we can build escalation into the existing infra instead of executing external commands like pkexec. Kinda similar to CUPS service but not exactly since cups itself is the one that raises privilege escalation.

I'd rather make rfkill unblocking a behind the scenes event than a user-prompted event though, that can fail and log if it doesnt succeed. Since from my POV at least, if I want to use a bluetooth adapter rfkill unblocking is something I just expect to be done implicitly.

alamin147:
Yeah I understand. I will look into it and implement built in system

alamin147:
Thanks, agreed. I updated the PR to route the normal Bluetooth toggle through bluetooth.setPowered, with the Quickshell adapter as fallback. This uses the existing implicit soft-rfkill handling and removes the button and pkexec command. The branch is rebased and lint passes. @bbedward

marcusramberg (approved):
Lgtm. Was actually just chasing why control center wasn't unblocking rfkill, so very happy to see someone else already chasing it. 😀

PR #3094, opened 2026-08-17
feat(chooser): edit and copy target

### LINKED ISSUE #3021: DMS Chooser / Add Ability to Copy Path
### Feature Description

Would it be possible to add the ability to copy the path from the DMS Chooser panel?

Image

### Use Case

When clicking on links, sometimes you realise it needs to be opened in a container tab in "Firefox", rather than close the DMS chooser it would be handy to be able to copy the link from the DMS Chooser itself.

GNOME Junction has this ability and its surprisingly useful.

### Compositor(s)

All compositors

### If Other, please specify

_No response_

### Proposed Solution

Make the path field in DMS Chooser a text field that can be selected by double clicking or add a quick "copy to clipboard" icon. 

Junction has this feature - https://apps.gnome.org/Junction/ - https://github.com/sonnyp/Junction

Image

hthienloc:
## Description

Make the DMS Chooser target editable and add quick copy support for file paths and URLs.

- Reuse the existing `DankTextField` and clipboard command.
- Add `Ctrl+Shift+C` support and temporary copy confirmation feedback.
- Preserve the original target binding while editing the current chooser request.

## Type of change

## Related issues

Closes #3021

## Video

https://github.com/user-attachments/assets/dac1e3c0-db22-474c-a4bd-8cd5ab7689e4

### DIFF
--- a/quickshell/Modals/AppPickerModal.qml
+++ b/quickshell/Modals/AppPickerModal.qml
@@ -564,8 +564,9 @@ DankModal {
 
                         Keys.onPressed: event => {
                             const hasCtrl = (event.modifiers & Qt.ControlModifier) !== 0;
+                            const hasShift = (event.modifiers & Qt.ShiftModifier) !== 0;
                             const hasOtherModifier = (event.modifiers & (Qt.AltModifier | Qt.MetaModifier)) !== 0;
-                            if (hasCtrl && !hasOtherModifier && event.key === Qt.Key_C) {
+                            if (hasCtrl && hasShift && !hasOtherModifier && event.key === Qt.Key_C) {
                                 root.copyEditableTarget();
                                 event.accepted = true;
                             }
@@ -581,7 +582,7 @@ DankModal {
                         iconName: root.targetCopied ? "check" : "content_copy"
                         iconSize: Theme.iconSize - 6
                         iconColor: root.targetCopied ? Theme.primary : Theme.surfaceText
-                        tooltipText: I18n.tr("Copy target (Ctrl+C)")
+                        tooltipText: I18n.tr("Copy target (Ctrl+Shift+C)")
                         onClicked: root.copyEditableTarget()
                     }
                 }
--- a/quickshell/Modals/AppPickerModal.qml
+++ b/quickshell/Modals/AppPickerModal.qml
@@ -229,6 +229,15 @@ DankModal {
             }
 
             Keys.onPressed: event => {
+                const hasCtrl = (event.modifiers & Qt.ControlModifier) !== 0;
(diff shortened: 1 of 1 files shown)


quickshell/Modals/AppPickerModal.qml
```diff
@@ -501,25 +519,71 @@ DankModal {
...
+                        Keys.onRightPressed: event => {
+                            if (cursorPosition === text.length)
+                                event.accepted = true;
+                        }
+                    }
+
+                    Item {
+                        id: targetCopyShortcutHandler
+
+                        Keys.onPressed: event => {
```
bbedward: Seems like it'd be better to move this up around line ~200-ish to the existing onPressed handler `appContent`

quickshell/Modals/AppPickerModal.qml
```diff
@@ -501,25 +519,71 @@ DankModal {
...
+                    DankActionButton {
+                        id: copyTargetButton
+                        anchors.right: parent.right
+                        anchors.rightMargin: Theme.spacingXS
+                        anchors.verticalCenter: parent.verticalCenter
+                        buttonSize: 36
+                        iconName: root.targetCopied ? "check" : "content_copy"
+                        iconSize: Theme.iconSize - 6
+                        iconColor: root.targetCopied ? Theme.primary : Theme.surfaceText
+                        tooltipText: I18n.tr("Copy target (Ctrl+Shift+C)")
```
bbedward: Use a parameter for the key binding in the trnaslation `I18n.tr("Copy target (%1)", "app picker copy target button tooltip").arg("Ctrl+Shift+C")`

quickshell/Modals/AppPickerModal.qml
```diff
@@ -501,25 +519,71 @@ DankModal {
...
-                    height: 36
+                    height: 52
                     radius: Theme.cornerRadius
                     color: Theme.withAlpha(Theme.surfaceContainerHigh, 0.5)
                     border.color: Theme.outlineMedium
                     border.width: 1
                     visible: root.showTargetData && root.targetData.length > 0
 
                     StyledText {
+                        id: targetDataLabel
```
bbedward: Best to rename since `targetDataLabel` exists as a property with the same name

bbedward (changes requested):

quickshell/Modals/AppPickerModal.qml
hthienloc: Using appContent navigation would make arrow keys move the selected app instead of the text cursor while typing in the text field. Is this your intention?

PR #2765, opened 2026-07-06
add: Cellular Network support

yechielw:
## Description
This PR was writin completly by codex
I have bean using it for a couple of weeks and it works perfect
it includs celluler options in quick settings, celluler indicator in the widget and celluler tab in the settings app
image
image
image

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that changes existing behavior)
- [ ] Refactor / internal cleanup
- [ ] Documentation
- [ ] Other

## Checklist

- [x] My code follows the conventions in CONTRIBUTING.md
- [x] I have tested my changes locally
- [x] New user-facing strings are wrapped in `I18n.tr()` with translator context, reusing existing terms where possible
- [x] Go changes: ran `make fmt`, added/updated tests, `make test` passes, and `go mod tidy` is clean
- [x] QML changes: ran `make lint-qml` with no new warnings
- [ ] I have opened a corresponding pull request in dlx-docs to document any new behaviors: https://github.com/AvengeMedia/DankLinux-Docs

PR #3096, opened 2026-08-17
add: Cellular Network support (rebase of #2765)

kmf:
## Description

Rebased and review-fixed version of #2765 (cellular network support).

Original work by @yechielw. Rebased onto current `master` and tested on a machine with a Fibocom L831-EAU-00 (Vodacom LTE).

Review from @bbedward on #2765:

- Rebased onto `master` (`term_freeze.json` is gone)
- Incremented `APIVersion` 30 → 31
- Regenerated `settings_search_index.json` via `extract_settings_index.py`
- Reused existing translation terms where they already cover the new UI
- Moved the Cellular settings tab to index 46 (avoids Battery / Mouse & Touchpad collisions)
- Added `wpa_supplicant` cellular stubs so the `Backend` interface still compiles
- Kept the newer hotspot Control Center UI while adding cellular

Fixes found while testing:

- Do not assign the readonly `userPreference` alias (Cellular chip was a no-op)
- Make the modem row actually clickable
- If no GSM profile exists, `AddAndActivateConnection` like Ethernet connect

## Type of change

## Related issues

Continues / supersedes #2765

## Checklist

- [x] Code follows CONTRIBUTING.md
- [x] New user-facing strings use `I18n.tr()`, reusing existing terms where possible
- [x] Go: `go test ./internal/server/...` passed
- [x] Tested on hardware with a WWAN modem


core/internal/server/network/backend_networkmanager.go
```diff
@@ -212,6 +223,36 @@ func (b *NetworkManagerBackend) Initialize() error {
...
+			hwAddr := ""
+			description := "Mobile broadband"
+			if g != nil {
+				hwAddr, _ = g.GetPropertyHwAddress()
+				if desc, err := g.GetPropertyTypeDescription(); err == nil && desc != "" {
+					description = desc
+				}
+			}
+
+			b.cellularDevices[iface] = &cellularDeviceInfo{
```
bbedward: This map access needs to be protected. Either uses the [syncmap from dankgo ](https://github.com/AvengeMedia/dankgo/tree/master/syncmap) or protect with RW mutex

core/internal/mocks/network/mock_Backend.go
```diff
@@ -87,9 +87,81 @@ func (_c *MockBackend_ActivateWiredConnection_Call) RunAndReturn(run func(uuid s
 	return _c
 }
 
-// CancelCredentials provides a mock function for the type MockBackend
-func (_mock *MockBackend) CancelCredentials(token string) error {
-	ret := _mock.Called(token)
+// ActivateCellularConnection provides a mock function with given fields: uuid
```
bbedward: This file shouldnt be hand-edited its generated. Re-generate mocks with mockery v3 just by running `mockery` e.g. `go run github.com/vektra/mockery/v3@latest`

This info about mockery was missing from CONTRIBUTING.md but I just added it on master

quickshell/Modules/ControlCenter/Details/NetworkDetail.qml
```diff
@@ -454,6 +489,336 @@ Rectangle {
...
+                            event.accepted = true;
+                        }
+                    }
+                }
+            }
+
+            StyledText {
+                width: parent.width
+                visible: (NetworkService.cellularDevices?.length ?? 0) === 0 && cellularConnectionsModel.values.length === 0
+                text: I18n.tr("No devices found")
```
bbedward, line 805: I feel like this text could render twice when theres no modem and no saved config (line 632 also)

core/internal/server/network/backend_networkmanager_cellular.go
```diff
@@ -0,0 +1,387 @@
...
+			return fmt.Errorf("failed to create and activate cellular connection: %w", err)
+		}
+	}
+
+	b.refreshCellularState()
+	return nil
+}
+
+func (b *NetworkManagerBackend) refreshCellularState() {
+	b.updateAllCellularDevices()
```
bbedward, line 125: Re-use this method in more places, some places are repeating this logic

core/internal/server/network/backend_networkmanager_cellular.go
```diff
@@ -0,0 +1,387 @@
...
+		},
+		"ipv6": {
+			"method": "auto",
+		},
+	}
+
+	if _, err := nm.AddAndActivateConnection(settings, dev); err != nil {
+		delete(settings["gsm"], "auto-config")
+		if _, retryErr := nm.AddAndActivateConnection(settings, dev); retryErr != nil {
+			return fmt.Errorf("failed to create and activate cellular connection: %w", err)
```
bbedward: This should wrap retryErr, seems like it'd be more indicative of the real error

quickshell/Modules/ControlCenter/Details/NetworkDetail.qml
```diff
@@ -454,6 +489,336 @@ Rectangle {
...
+                font.weight: Font.Medium
+                horizontalAlignment: Text.AlignHCenter
+            }
+
+            Rectangle {
+                anchors.horizontalCenter: parent.horizontalCenter
+                width: enableCellularLabel.implicitWidth + Theme.spacingL * 2
+                height: enableCellularLabel.implicitHeight + Theme.spacingM * 2
+                radius: height / 2
+                color: enableCellularButton.containsMouse ? Qt.rgba(Theme.primary.r, Theme.primary.g, Theme.primary.b, 0.12) : Qt.rgba(Theme.primary.r, Theme.primary.g, Theme.primary.b, 0.08)
```
bbedward: Instead of hand rolling Qt.rgba re-use the appropriate tokens like wifiOffContent does `Theme.surfaceTextSecondary/primaryHover/primaryHoverLight`

bbedward (changes requested):

core/internal/server/network/backend_networkmanager.go
kmf: Done in e922cc98. Added `cellularDevicesSnapshot()` / `cellularDeviceByIface()` / `setCellularDeviceInfo()` / `removeCellularDeviceByPath()` guarded by the existing `devMutex`, mirroring the ethernet and wifi accessors already in this file — `cellularDevices` is now covered by the same pattern rather than a second mechanism. Every raw access is converted: `Initialize`, both signal-pump loops, `handleDeviceAdded`, `handleDeviceRemoved`, `updateCellularState`, `updateAllCellularDevices` and `DisconnectCellularDevice`. Also updated the `devMutex` doc comment to mention the map.

Happy to switch to `syncmap` instead if you would rather the new map not extend `devMutex`.

core/internal/mocks/network/mock_Backend.go
kmf: Done in e922cc98 — regenerated with `go run github.com/vektra/mockery/v3@latest` (v3.7.3) instead of hand-editing. The file is now purely additive against master (395 insertions, 0 deletions); the hand-edited version had drifted from the generator output.

One note: mockery also regenerates `internal/mocks/internal/plugins/mock_GitClient.go`, which is stale on master and unrelated to cellular. I reverted that one to keep this PR focused — worth a separate regen commit on master.

quickshell/Modules/ControlCenter/Details/NetworkDetail.qml
kmf: Good catch, confirmed. The label at the top of `cellularColumn` was gated only on `cellularDevices.length === 0`, while this one also checks the connections model — so with no modem and no saved config both were visible. Dropped the top one in e922cc98; this one's condition was already the correct combined check.

core/internal/server/network/backend_networkmanager_cellular.go
kmf: Done in e922cc98. `refreshCellularState()` is now used in `SetCellularEnabled`, `DisconnectCellular`, `DisconnectCellularDevice`, `ActivateCellularConnection` and the device-removal path in `backend_networkmanager_signals.go`.

I left the remaining sites in `signals.go` alone on purpose — those interleave ethernet/wifi/hotspot updates and fire a single `onStateChange` at the end, so folding the helper in would emit an extra mid-sequence broadcast.

core/internal/server/network/backend_networkmanager_cellular.go
kmf: Agreed — fixed in e922cc98, it now wraps `retryErr`. Also added a short comment noting the retry exists because older NetworkManager/ModemManager stacks reject `gsm.auto-config`.

quickshell/Modules/ControlCenter/Details/NetworkDetail.qml
kmf: Done in e922cc98 — the icon now uses `Theme.surfaceTextSecondary` and the button uses `Theme.primaryHover` / `Theme.primaryHoverLight`, matching `wifiOffContent`.

kmf:
Still testing

kmf:
## Hardware testing done

Re-tested on the Fibocom L831-EAU-00 (Vodacom SA, `cdc-wdm0`) after the review fixes in e922cc98.

### Build / unit

- `go build ./...` clean
- `go test -race ./internal/server/network/...` passes
- `go vet ./internal/server/network/...` fails on `manager_test.go:389` (`assert.NotNil` copies a `syncmap.Map` lock). This reproduces on a clean `origin/master` worktree too, so it is pre-existing and not from this PR — flagging in case you want it fixed separately.

### Cellular against the real modem

Driven over the backend socket (API v31, NetworkManager backend), starting from a **deleted** GSM profile so the create-from-scratch path was exercised:

| Operation | Result |
| --- | --- |
| `connect`, no saved profile | `AddAndActivateConnection` created + activated "Mobile broadband", IP `100.65.208.90` |
| state after connect | device `activated`, connection listed `isActive`, `cellularConnectionUuid` set, preference → cellular |
| `disconnect` | `disconnected`, IP/uuid cleared, profile retained |
| `connect`, existing profile | reconnected via `ActivateConnection`, IP `10.68.140.18` |
| radio `disable` / `enable` | `unavailable` → `activated`, IP `100.88.195.33` |
| radio `toggle` | back to `unavailable` |

All four `refreshCellularState()` steps (devices, state, connections, primary connection) propagate correctly in every case, so the de-duplication in that refactor is verified end to end.

### UI

Control Center cellular panel renders correctly in both states:

- **Radio off** — muted `Theme.surfaceTextSecondary` icon and the "Enable Cellular" button in `Theme.primaryHoverLight`, consistent with the rest of the theme (this is the hand-rolled `Qt.rgba` fix).
- **Connected** — single "Mobile broadband / Connected" row, active border, disconnect action, and no stray "No devices found" label.

### Two things I could not prove on this hardware

1. **The `auto-config` retry branch never executed.** The first `AddAndActivateConnection` succeeds with `auto-config` on this modem, so the `retryErr` wrap was only observed on an unrelated polkit-denied attempt (`Not authorized to control networking.` — my own fault for launching the backend from an ssh session rather than the graphical one). The wrap is correct by inspection, but a stack that genuinely rejects `gsm.auto-config` would be needed to exercise it for real.
2. **The exact duplicate-label scenario was not reproduced.** It requires *zero* modems present; unbinding the USB device seemed more invasive than warranted. With a modem attached only the correct single empty state is reachable, so that specific double-render rests on code inspection.

quickshell/Services/DMSNetworkService.qml
```diff
@@ -716,6 +782,22 @@ Singleton {
...
 
+    function toggleCellularRadio() {
+        if (!networkAvailable || cellularToggling)
+            return;
+        cellularToggling = true;
+        DMSService.sendRequest("network.cellular.toggle", null, response => {
+            cellularToggling = false;
+
+            if (response.error) {
+                ToastService.showError(I18n.tr("Failed to enable cellular"), response.error);
```
bbedward: The error is always "failed to enable cellular" even if its a failed to disable. I would just replace the string with "Failed to toggle cellular"

quickshell/Services/NetworkService.qml
```diff
@@ -221,12 +231,24 @@ Singleton {
...
+        }
+    }
+
     function enableWifiDevice() {
         if (activeService && activeService.enableWifiDevice) {
             activeService.enableWifiDevice();
         }
     }
 
+    function enableCellularRadio() {
```
bbedward: Not used so I'd just delete it, in addition to the function in DMSNetworkService

quickshell/Modules/ControlCenter/Details/NetworkDetail.qml
```diff
@@ -39,12 +45,28 @@ Rectangle {
...
+        const types = [];
+        if (hasEthernetAvailable)
+            types.push("ethernet");
+        if (hasWifiAvailable)
+            types.push("wifi");
+        if (hasCellularAvailable)
+            types.push("cellular");
+        return types.length > 0 ? types : ["wifi"];
+    }
+    property int selectedTypeIndex: -1
```
bbedward: Think it would be more robust to store the selected value instead of index

```js
property string selectedType: ""
readonly property string currentConnectionType: {
    if (selectedType && connectionTypes.includes(selectedType))
        return selectedType;
    return connectionTypes[Math.max(0, currentPreferenceIndex)] || "wifi";
}
```

quickshell/Modules/ControlCenter/Details/NetworkDetail.qml
```diff
@@ -153,20 +168,35 @@ Rectangle {
...
+
+                visible: connectionTypes.length > 1 && NetworkService.backend === "networkmanager" && DMSService.apiVersion > 10
+                model: connectionTypes.map(t => labelsByType[t] || t)
+                currentIndex: selectedTypeIndex >= 0 ? selectedTypeIndex : currentPreferenceIndex
                 selectionMode: "single"
                 onSelectionChanged: (index, selected) => {
                     if (!selected)
                         return;
-                    NetworkService.setNetworkPreference(index === 0 ? "ethernet" : "wifi");
+                    selectedTypeIndex = index;
```
bbedward: And here instead of index:

```js
currentIndex: Math.max(0, connectionTypes.indexOf(currentConnectionType))
onSelectionChanged: (index, selected) => {
    if (!selected)
        return;
    selectedType = connectionTypes[index] || "wifi";
    NetworkService.setNetworkPreference(selectedType);
}
```

bbedward (changes requested):

core/internal/server/network/backend_networkmanager_cellular.go
```diff
@@ -0,0 +1,362 @@
...
+
+	for _, conn := range connections {
+		connSettings, err := conn.GetSettings()
+		if err != nil {
+			continue
+		}
+
+		if connMeta, ok := connSettings["connection"]; ok {
+			if connType, ok := connMeta["type"].(string); ok && isCellularConnectionType(connType) {
+				if _, err := nm.ActivateConnection(conn, dev, nil); err != nil {
```
kmf: **[bug]** `ConnectCellular` activates the first `gsm`/`cdma` profile from `ListConnections()` and returns immediately on failure. NetworkManager/ModemManager commonly auto-creates several GSM profiles (internet, MMS, dun) and list order is not priority order, so a generic Connect can attach the MMS APN — or fail and never try the internet profile or the `AddAndActivateConnection` fallback below.

**Suggestion:** Skip non-default/MMS-style profiles (e.g. `ipv4.method=disabled`, no default route), prefer autoconnect-priority / last-used UUID, and on `ActivateConnection` failure continue to the next cellular profile before creating a new one.

quickshell/Services/DMSNetworkService.qml
kmf: **[suggestion]** `toggleCellularRadio()` always toasts "Failed to enable cellular" on error, including when the user is turning WWAN off. The success path already branches on `cellularEnabled`.

**Suggestion:** Use `I18n.tr("Failed to toggle cellular")` (or branch on the pre-toggle `cellularEnabled` value), matching the existing review comment.

quickshell/Services/NetworkService.qml
kmf: **[suggestion]** `enableCellularRadio()` is still unused, as is `DMSNetworkService.enableCellularRadio()` (line 813). The "Enable Cellular" button calls `toggleCellularRadio()`. The backend `network.cellular.enable` handler can stay for IPC; the QML wrappers are dead.

**Suggestion:** Delete both QML functions unless a caller is added.

quickshell/Modules/ControlCenter/Details/NetworkDetail.qml
kmf: **[suggestion]** `selectedTypeIndex` stores a position in the dynamic `connectionTypes` list (`ethernet`, then `wifi`, then `cellular` as each appears). If a NIC is added/removed while the panel is open, that index maps to a different medium — e.g. viewing WiFi at index 1 with `[ethernet, wifi, cellular]`, then unplugging ethernet, yields `[wifi, cellular]` and the pane jumps to Cellular. `onSelectionChanged` (line 187) writes the same index.

**Suggestion:** Store `selectedType` as the string value (`"wifi"` / `"ethernet"` / `"cellular"`) and derive `currentIndex` via `connectionTypes.indexOf(...)`, as previously suggested.

quickshell/Modules/Settings/NetworkStatusTab.qml
```diff
@@ -161,13 +164,15 @@ Item {
 
                         DankButtonGroup {
                             id: preferenceButtons
-                            model: [I18n.tr("Auto"), I18n.tr("Ethernet"), I18n.tr("WiFi")]
+                            model: [I18n.tr("Auto"), I18n.tr("Ethernet"), I18n.tr("WiFi"), I18n.tr("Cellular")]
```
kmf: **[suggestion]** The preference row is now shown whenever any two of ethernet/wifi/cellular are connected, but the button model is still the fixed `[Auto, Ethernet, WiFi, Cellular]`. Ethernet+WiFi users get a Cellular choice that deprioritizes both real uplinks even when no modem exists; ethernet+cellular users likewise get a WiFi choice.

**Suggestion:** Build the model from available/connected types (plus Auto), the same way `NetworkDetail.connectionTypes` does, and map selection back from the chosen value rather than a hard-coded index.

quickshell/Modules/Settings/NetworkCellularTab.qml
```diff
@@ -0,0 +1,331 @@
...
+                                        MouseArea {
+                                            id: modemActionBtn
+                                            anchors.fill: parent
+                                            hoverEnabled: true
+                                            cursorShape: Qt.PointingHandCursor
+                                            onClicked: {
+                                                if (modemDelegate.isConnected)
+                                                    NetworkService.disconnectCellularDevice(modelData.name);
+                                                else
+                                                    NetworkService.connectCellular();
```
kmf: **[suggestion]** Each adapter row has its own connect control, but the disconnected path calls `NetworkService.connectCellular()` with no device. The backend always activates `b.cellularDevice` (first modem seen). Disconnect correctly uses `disconnectCellularDevice(modelData.name)`, so on a multi-modem machine the connect button on modem B will enable modem A.

**Suggestion:** Add a device argument to connect (mirroring disconnect) and pass `modelData.name` from both the action button and the row `MouseArea` (line 209).

quickshell/Services/DMSNetworkService.qml
kmf: Done in 29bc88dd — the error toast is now `Failed to toggle cellular`.

quickshell/Services/NetworkService.qml
kmf: Done in 29bc88dd — removed `enableCellularRadio()` from both `NetworkService.qml` and `DMSNetworkService.qml`. The backend `network.cellular.enable` IPC handler is still there.

quickshell/Modules/ControlCenter/Details/NetworkDetail.qml
kmf: Done in 29bc88dd — `selectedType` is now a string, and `currentConnectionType` uses it when it is still in `connectionTypes`, otherwise it falls back to preference. If a NIC disappears while the panel is open the pane no longer remaps to a different medium via a stale index.

quickshell/Modules/ControlCenter/Details/NetworkDetail.qml
kmf: Done in 29bc88dd — `currentIndex` is `connectionTypes.indexOf(currentConnectionType)`, and `onSelectionChanged` writes `selectedType` from the chosen value.

core/internal/server/network/priority.go
```diff
@@ -161,7 +208,23 @@ func (m *Manager) setConnectionPriority(connType string, autoconnectPriority int
...
+
+		connRef := connName
+		if connUUID != "" {
+			connRef = "uuid:" + connUUID
+		}
+
+		if err := exec.Command("nmcli", "con", "mod", connRef,
 			"connection.autoconnect-priority", fmt.Sprintf("%d", autoconnectPriority),
 			"ipv4.route-metric", fmt.Sprintf("%d", routeMetric),
 			"ipv6.route-metric", fmt.Sprintf("%d", routeMetric)).Run(); err != nil {
```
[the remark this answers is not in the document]
kmf: Good catch, and it's worse than a cellular bug — confirmed and fixed in 3efc98ca.

Verified empirically on my test box using `con show` (same selector parsing as `con mod`, but read-only):

```
$ nmcli con show uuid:9f010098-7fb0-341d-8859-034f060aeb39
Error: uuid:9f010098-7fb0-341d-8859-034f060aeb39 - no such connection profile.
$ nmcli con show uuid 9f010098-7fb0-341d-8859-034f060aeb39
OK
$ nmcli con show "Wired connection 1"
OK
```

And you're right that this was a regression I introduced, not just a cellular gap: `master` passed `connName` here, and since practically every profile has a UUID, `connRef` was almost always the broken form — so autoconnect-priority and route-metric updates were silently no-ops for ethernet and WiFi too. The failure only reached `log.Warnf` and the caller ignores the error, which is why nothing surfaced.

Applied your suggestion as-is (arg-list construction, `uuid <UUID>` when present, bare name otherwise). `gofmt` clean, `go build ./...` and `go test ./internal/server/...` pass.

quickshell/Modules/Settings/NetworkCellularTab.qml
```diff
@@ -0,0 +1,331 @@
...
+                    Row {
+                        width: parent.width
+                        spacing: Theme.spacingM
+
+                        StyledText {
+                            text: {
+                                if (!NetworkService.cellularHardwareEnabled)
+                                    return I18n.tr("Unavailable");
+                                if (NetworkService.cellularToggling)
+                                    return I18n.tr("Connecting...");
```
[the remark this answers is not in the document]
kmf: Fixed in 3efc98ca — applied your suggestion.

While doing it I found the terms you pointed at weren't actually in the catalog: `en.json`/`template.json` were never regenerated for this PR, so 7 of the 23 new `I18n.tr()` terms were missing, including both `Enabling cellular...` and `Disabling cellular...` (already used at `NetworkDetail.qml:542`). Added them.

I did *not* just run `extract_translations.py` — it rewrites the catalog wholesale and came out at −2599/+970, dropping 107 existing terms while adding 8. Given `i18nsync.py` warns that `--prune` deletes every POEditor term missing from the local `en.json` along with its translations, that looked destructive and well out of scope here. So I spliced in only the 7 missing entries, generating them through the script's own `create_poeditor_json`/`create_template_json` so the context/reference/comment/tags fields match its output exactly. Diff is additive only: `en.json +66`, `template.json +49`, both still sorted.

Worth flagging separately: the committed catalog and a fresh extraction disagree by ~100 terms on `master` too, so something is stale there independent of this PR.

kmf:
@bbedward anything else needed for this?

PR #3103, opened 2026-08-18
feat(settings): improve battery status layout

hthienloc:
## Type of change

## Screenshots / video

dms_capture_1787184832770
dms_capture_1787184923595


hthienloc:
@bbedward @Purian23 My initial intention is to redesign the battery info tab in the settings, but there is a question I want to clarify before going any further:

Should we repeat the information that's already available in the battery widget?

Because I'd like to remove the duplicate information and replace it with something more useful, like battery history, so users can track their battery level over an extended period of time. But the edge case is some user might not use the battery widget so this is still holding some use case.

purian23:
Hey @hthienloc, I'm fairly open on this one with no strong opinions either way. It should repeat maybe the basics of info, but otherwise open to the options you listed for sure. Thanks!

PR #3108, opened 2026-08-19
fix(dgop): stop the process list opening blank

cwelsys:
## Description
 
Renders the first process sample instead of discarding it.

Queues a refetch when updateAllStats is called during an in-flight request instead of dropping it.

Also restores mergeChildren behavior which was dropped in da664565 

## Type of change

- [ ] Documentation
- [ ] Other

## Checklist


quickshell/Services/DgopService.qml
```diff
@@ -206,10 +205,13 @@ Singleton {
...
             return;
         }
-        if (isUpdating)
+        if (isUpdating) {
+            pendingUpdate = true;
             return;
+        }
 
         const params = buildMetaParams();
         if (!params)
```
bbedward: should we clear pendingUpdate here ?

quickshell/Services/DgopService.qml
```diff
@@ -220,9 +222,14 @@ Singleton {
...
                 log.warn("dgop.meta failed:", response.error || "empty result");
                 isUpdating = false;
-                return;
+            } else {
+                parseData(response.result);
+            }
+
+            if (pendingUpdate) {
+                pendingUpdate = false;
+                updateAllStats();
```
bbedward: I think this should run through the `primeTimer` so the cursor gets a real interval

bbedward (changes requested):

quickshell/Services/DgopService.qml
cwelsys: Requeue waits the 1s prime now instead of firing immediately

quickshell/Services/DgopService.qml
cwelsys: Yeah, buildMetaParams returns null there and the length guard above doesn't catch it

PR #3128, opened 2026-08-20
feat(greeter): auto-start external auth for howdy face recognition

yuhaoxin:
## Problem

The greeter auto-starts PAM external authentication (via `maybeAutoStartExternalAuth`) only for fingerprint (`pam_fprintd`) and security keys (`pam_u2f`). When [howdy](https://github.com/boltgolt/howdy) face recognition (`pam_howdy.so`) is in the greetd PAM stack, the camera stays off until the user presses Enter once — defeating the purpose of face login.

## Change

Treat `pam_howdy` like the other external auth modules:

- `greeterPamHasHowdy` detects `pam_howdy` in the effective PAM stack (reusing `greeterPamStackHasModule`)
- Included in `greeterExternalAuthAvailable` / `greeterPamHasExternalAuth`, so `maybeAutoStartExternalAuth()` fires and the camera activates as soon as the password view is shown
- Status message "Awaiting face authentication" while howdy runs

No settings toggle is added: presence of `pam_howdy.so` in the PAM stack is an explicit admin choice, same as the module ordering. Unlike fprint, howdy has no D-Bus probe equivalent, so PAM-stack detection is authoritative (howdy exits quickly with `PAM_AUTHINFO_UNAVAIL` when disabled via its own config).

## Testing

- Arch Linux, greetd + niri, [howdy-next](https://codeberg.org/nathawat/howdy) 3.4.0 (`pam_howdy.so`), `auth sufficient pam_howdy.so` in `/etc/pam.d/greetd`
- Before: camera only activated after pressing Enter
- After: camera LED turns on automatically when the password view appears; face match completes login without keyboard input; password entry still works as fallback; `qmllint` clean

### DIFF
--- a/quickshell/Modules/Greetd/GreeterContent.qml
+++ b/quickshell/Modules/Greetd/GreeterContent.qml
@@ -72,8 +72,9 @@ Item {
     // Falls back to PAM-only detection until the fprintd D-Bus probe completes.
     readonly property bool greeterPamHasFprint: greeterPamStackHasFprint && (!fprintdProbeComplete || fprintdHasDevice)
     readonly property bool greeterPamHasU2f: greeterPamStackHasModule("pam_u2f")
-    readonly property bool greeterExternalAuthAvailable: (greeterPamHasFprint && GreetdSettings.greeterEnableFprint) || (greeterPamHasU2f && GreetdSettings.greeterEnableU2f)
-    readonly property bool greeterPamHasExternalAuth: greeterPamHasFprint || greeterPamHasU2f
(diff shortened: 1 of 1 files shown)


bbedward:
I'm ok to merge this to the stable branch, but it needs to go to the dank-greeter repo since 1.6+ will be using that one https://github.com/AvengeMedia/dank-greeter

bbedward:
https://github.com/AvengeMedia/dank-greeter/commit/692171e730966c79fc141b1979785d7bd0d90a7b

PR #3149, opened 2026-08-23
feat(matugen): add standalone Spicetify theme

hthienloc:
Adds a standalone DMS Spicetify theme generated from the current Matugen colors. Disabling the template restores the `marketplace` theme.

Due to a Spicetify limitation, theme changes cannot be applied on the fly and only take effect after Spotify is reloaded or restarted.


hthienloc:
Closed since this is not working as intended due to a spicetify limitation

PR #3155, opened 2026-08-23
feat(screenshot): use native cursor shapes

hthienloc:
Port `wp_cursor_shape_manager_v1` to the Go screenshot backend and use compositor-native cursor shapes for the region selector.

This establishes the cursor protocol layer for follow-up PRs that add Ctrl-drag selection behavior and transition between `Crosshair`, `Grab`, and `Grabbing`. This PR intentionally does not include the Ctrl-drag interaction itself.


core/internal/proto/wp_cursor_shape/cursor_shape.go
```diff
@@ -0,0 +1,64 @@
+package wp_cursor_shape
```
bbedward, line 29: I'd like to see the XML spec included in the proto/xml directory and generate this file instead of hand writing it.

The README command is outdated and I'll fix that, but the command would look like this  once the XML is added

```bash
go run github.com/AvengeMedia/dankgo/cmd/go-wayland-scanner@latest \
  -i internal/proto/xml/cursor-shape-v1.xml \
  -pkg wlr_gamma_control -o internal/proto/wp_cursor_shape/cursor_shape.go
```

bbedward (changes requested):

PR #3160, opened 2026-08-23
feat(control-center): show Bluetooth battery in quick tile

0xdeafbeef:
## Description

Shows the connected Bluetooth device battery level in the Control Center tile when available.

## Type of change

- [x] New feature

## Related issues

None.

## Screenshots / video

Screenshot from 2026-08-23 20-33-48

## Checklist

### DIFF
--- a/quickshell/Modules/ControlCenter/Components/DragDropGrid.qml
+++ b/quickshell/Modules/ControlCenter/Components/DragDropGrid.qml
@@ -559,8 +559,18 @@ Column {
                                 }
                                 return null;
                             })();
-                        if (primaryDevice)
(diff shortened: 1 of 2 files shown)


0xdeafbeef:
I can make it toggleable from the settings if you want. But it utilizes free space, so I see 0 problems, only value :)

0xdeafbeef:
@bbedward I've added a comment with links to the `Quickshell` and `BlueZ` docs. I actually saw 0% reported while testing with connected headphones, so I kept it as a valid value instead of treating it as unavailable. I also updated the existing bluetooth details check to follow the same behavior

PR #3161, opened 2026-08-23
fix(typography): use monospace for dynamic numeric values

0xdeafbeef:
## Description

The CPU temperature kept changing width and making the bar jump around, which was pretty distracting.
This switches dynamic numbers in DankBar to the configured monospace font so the layout stays stable.

## Type of change

## Related issues

None.

## Checklist

### DIFF
--- a/quickshell/Modules/DankBar/Widgets/Battery.qml
+++ b/quickshell/Modules/DankBar/Widgets/Battery.qml
@@ -151,6 +151,7 @@ BasePill {
 
                 StyledText {
                     text: battery.verticalDisplayText
+                    isMonospace: true
                     font.pixelSize: Theme.barTextSize(battery.barThickness, battery.barConfig?.fontScale, battery.barConfig?.maximizeWidgetText)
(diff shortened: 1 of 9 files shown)


bbedward:
We dont like this approach as much because it makes the fonts inconsistent, the intended behavior that may not be working correctly is to use TextMetrics to basically keep a minimum width for the largest possible value, that way the text changes but the overall widget size stays the same

0xdeafbeef:
@bbedward 

non-mono

mono


TextMetrics keeps the widget’s outer width stable, but the proportional digits still visibly shift within that fixed space. 

Mono keeps both the widget width and the digit positions stable :)

BTW


Qt supports `font.features: { "tnum": 1 }`, and the bundled Inter font has tabular numerals, so that might make both of us happy :) We can keep the intended font consistent while preventing the digits from visually jumping around.

Should i open an issue? I can make a pr  latter

0xdeafbeef:
@bbedward thank you :)

PR #3234, opened 2026-08-30
fix(audio): show the muted icon when volume reaches 0%

### LINKED ISSUE #3050: UX: Show muted audio icon when volume reaches 0%
Currently, when the system volume is lowered to 0% using keyboard media keys, mouse wheel scrolling on the volume control, or by dragging the volume slider in Settings, the audio icon continues to show the normal speaker icon with 0%.

https://github.com/user-attachments/assets/c55c73a8-2598-468d-9d63-32d21131b1d2

Image

Image

Mario-Mohar:
## Description

At zero volume `AudioService.volumeIconName()` returned `volume_mute` — the speaker glyph *without* sound waves. At bar size that reads as the ordinary speaker, which is why lowering the volume to 0% looks like nothing happened (the screenshots in #3050 show exactly this glyph). Zero is now treated like muted and returns `volume_off`, the crossed-out speaker.

Two other places never used the service helper at all:

- `Modules/OSD/VolumeOSD.qml` looked only at `audio.muted`, so the OSD showed `volume_up` while the level read 0 — that is the case in the issue's video. Both the horizontal and the vertical layout now go through the service.
- `Modules/Greetd/GreeterContent.qml` carried a third copy of the same branching, including a dimming condition that already treated "muted or zero" as one state.

To express that shared question once, the service gains `isSilent(node)` and a `sinkSilent` property; the icon helper is written in terms of it. The zero comparison is `<= 0` rather than `=== 0` so a slider drag that lands just under zero cannot slip through. `volume_mute` is no longer referenced anywhere in the tree.

## Type of change

- [ ] Documentation
- [ ] Other

## Related issues

Fixes #3050

## Screenshots / video

Bar, same session, before and after setting the sink to 0% (`dms ipc call audio setvolume 0`). At 100% the speaker keeps its waves; at 0% it is crossed out.

Volume 100%:

volume at 100 percent

Volume 0%:

volume at 0 percent

## Checklist

- [x] New user-facing strings are wrapped in `I18n.tr()` with translator context, reusing existing terms where possible — no new strings in this PR
- [x] Go changes: none in this PR
- [ ] I have opened a corresponding pull request in dlx-docs — not applicable: this corrects an icon, it documents no new behavior

### How this was verified

Ran the built `dms` against this working tree inside a nested Hyprland session on its own `XDG_CONFIG_HOME`/`XDG_STATE_HOME`/`XDG_DATA_HOME`, drove the volume through `dms ipc call audio setvolume`, and compared the bar before and after.

`make lint-qml` needs the Quickshell tooling VFS (`.qmlls.ini`), which this machine does not generate outside a full DMS install, so the three touched files were checked with Qt 6 `qmllint` directly — no errors beyond the pre-existing unresolved-import warnings that every file in the tree produces without the VFS.

### DIFF
--- a/quickshell/Modules/Greetd/GreeterContent.qml
+++ b/quickshell/Modules/Greetd/GreeterContent.qml
@@ -1579,17 +1579,15 @@ Item {
                         if (!AudioService.sink?.audio) {
                             return "volume_up";
                         }
-                        if (AudioService.sink.audio.muted)
+                        if (AudioService.isSilent(AudioService.sink))
                             return "volume_off";
-                        if (AudioService.sink.audio.volume === 0)
-                            return "volume_mute";
                         if (AudioService.sink.audio.volume * 100 < 33) {
                             return "volume_down";
                         }
                         return "volume_up";
                     }
                     size: Theme.iconSize - 2
-                    color: (AudioService.sink && AudioService.sink.audio && (AudioService.sink.audio.muted || AudioService.sink.audio.volume === 0)) ? Qt.rgba(255, 255, 255, 0.5) : "white"
+                    color: AudioService.sinkSilent ? Qt.rgba(255, 255, 255, 0.5) : "white"
                     anchors.verticalCenter: parent.verticalCenter
                     visible: AudioService.sink && AudioService.sink.audio
                 }
--- a/quickshell/Modules/OSD/VolumeOSD.qml
+++ b/quickshell/Modules/OSD/VolumeOSD.qml
@@ -65,7 +65,7 @@ DankOSD {
 
             OsdLevelRow {
                 anchors.fill: parent
-                iconName: AudioService.sink?.audio?.muted ? "volume_off" : "volume_up"
+                iconName: AudioService.sinkSilent ? "volume_off" : "volume_up"
                 iconInteractive: true
                 value: root._displayVolume
                 minimum: 0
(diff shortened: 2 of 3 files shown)


bbedward:
I sort of get the complaint although I'm not sure I agree with this approach exactly. Because "muted" and "volume 0" are two separate states and this would make them appears to be the same. 

Think originally we had the icons set this way then changed it to the current behavior because of that.

See if 5c552f97e4553947b17534ffbe091dc9ef958efe is a sufficient alternative, adding a new inactive icon color to distinguish the state

Mario-Mohar:
Fair point, and looking at master you've already covered the actual complaint in 5c552f97.

The bug I was after was `VolumeOSD` hardcoding `muted ? "volume_off" : "volume_up"`, so at 0% unmuted the OSD showed a volume_up icon next to a slider sitting at zero. Both call sites (`VolumeOSD.qml:68` and `:111`) now go through `sinkVolumeIconName`, so that's fixed — and keeping `volume_mute` distinct plus the new `widgetInactiveIconColor` is the better call than collapsing the two states the way I did. Agreed on that; you're right that they're separate states.

The `isSilent`/`sinkSilent` extraction landed with your commit too, so there's nothing left in this branch except the icon collapse you don't want — which is also what it now conflicts on.

Closing this one — it's fully covered by 5c552f97.

Issue #3232, opened 2026-08-30
A corrupted clipboard DB crashes the entire dms process on the next copy — `storeEntry`/`deduplicateInTx` lack nil/error checks


wxmyyds:
### Compositor

Niri

### Distribution

Arch Linux

### Select your Installation Method

Distro Packaging

### Was this your original Installation method?

Yes

### dms doctor -vC

<details>
<summary>Click to expand</summary>

## DMS Doctor Report

**System**
- [ok] Operating System: Arch Linux
  - ID: arch, Version: rolling, Arch: amd64
- [ok] Architecture: amd64
- [ok] Display Server: Wayland
  - WAYLAND_DISPLAY=wayland-1

**Versions**
- [ok] DMS CLI: dms v1.5.3
  - /usr/bin/dms
- [ok] Quickshell: Quickshell 0.3.1 (revision , distributed by Arch Linux)
  - /usr/bin/qs
- [ok] DMS Shell: v1.5.3
  - /usr/share/quickshell/dms

**Installation**
- [ok] DMS Configuration: Found
  - /usr/share/quickshell/dms
- [ok] shell.qml: Present
  - /usr/share/quickshell/dms/shell.qml
- [info] Install Type: System package
  - /usr/share/quickshell/dms

**Compositor**
- [ok] niri: 26.04
  - /usr/bin/niri
- [info] Active: niri
- [ok] Background Blur: Supported
  - Compositor supports ext-background-effect-v1

**Quickshell Features**
- [ok] Polkit: Available
  - Authentication prompts
- [ok] IdleMonitor: Available
  - Idle detection
- [ok] IdleInhibitor: Available
  - Prevent idle/sleep
- [ok] ShortcutInhibitor: Available
  - Allow shortcut management (niri)
- [ok] BackgroundBlur: Available
  - Background blur API support in Quickshell

**Optional Features**
- [ok] accountsservice: Available
  - User accounts
- [ok] power-profiles-daemon: Available
  - Power profile management
- [ok] logind: Available
  - Session management
- [ok] cups-pk-helper: Available
  - Printer management
- [info] I2C/DDC: No monitors detected
  - External monitor brightness control
- [ok] qt6-imageformats: Installed (1 formats)
  - Formats: GIF (/usr/lib/qt6/plugins/imageformats:/usr/lib64/qt6/plugins/imageformats)
- [ok] kimageformats: Installed (4 formats)
[...]

PR #3235, opened 2026-08-30
feat(clipboard): clear only the filtered entries

### LINKED ISSUE #3198: Clear filtered clipboard entries (not just all)
### Feature Description

The clipboard manager has a "Clear All" button that deletes **all unpinned entries** regardless of the active search query or type filter. There is no way to delete only the entries that match the current filter.

### Use Case

When you have a long clipboard history and want to clean up only specific entries (e.g., all image entries, or all entries matching a search term), you currently have to delete them one by one via the context menu. Being able to clear only the filtered results would save significant time.

### Compositor(s)

All compositors

### Proposed Solution

When a search query or type filter is active, the "Clear All" button should delete only the **currently filtered** entries (pinned entries are still kept). When no filter is active, behavior stays as-is (clear all unpinned).

Clipboard managers like CopyQ and Klipper allow clearing only the visible/filtered entries.

Mario-Mohar:
## Description

The clear button in the clipboard modal deleted every unpinned entry regardless of the active search query or type filter, so pruning a subset meant deleting entries one at a time from the context menu (#3198).

**Behaviour**

- When the recents list is filtered — search text, or a type filter other than `all` — the button clears exactly the entries the list is showing, and the modal stays open on that filter so the next one can be worked through. `Shift+Del` takes the same path.
- With no filter, nothing changes: all unpinned entries, pinned kept, modal closes.
- A filter that matches nothing does **nothing**. Without that guard, filtering to zero results and hitting clear would have fallen back to wiping the whole history — the sharpest edge in this change.
- The tooltip reads "Clear Filtered" while a filter is active, and the confirmation names the count.
- The saved tab and the bar widget's context-menu entry are untouched; both still clear everything, since neither is showing a filtered list.

**Why a new IPC method.** Deleting the filtered ids one at a time would cost a round trip *and* a full clipboard state broadcast per entry. `clipboard.deleteEntries` does it in one bolt transaction with a single `updateState`/`notifySubscribers`, and returns how many rows it removed. It skips pinned entries server-side: the UI never sends a pinned id, but a bulk delete should not be *able* to take out saved items.

New strings: `Clear Filtered` and the confirmation sentence. The term freeze is currently inactive (`term_freeze.json` absent) and `check_term_variants.py` reports no case or punctuation duplicates against the existing catalog. Catalogs are left alone per CONTRIBUTING.

## Type of change

- [ ] Documentation
- [ ] Other

## Related issues

Fixes #3198

## Screenshots / video

Six entries, one pinned.

Filtered to `TREFFER` — two of six match:

filtered list

The confirmation names the count:

confirmation dialog

After confirming, the modal stays open on the filter and the other four entries survive:

after clearing

With no filter the dialog is unchanged:

unfiltered dialog

## Checklist

- [x] New user-facing strings are wrapped in `I18n.tr()`, reusing existing terms where possible (the dialog title reuses `Clear History?`)
- [x] Go changes: ran `make fmt`, added tests, `make test` passes, `go mod tidy` is clean
- [x] QML changes: linted, no new warnings
- [ ] I have opened a corresponding pull request in dlx-docs — happy to, if you want it: the clipboard page in dlx-docs documents the `dms cl` CLI, which this PR does not change, and the modal's clear button is not described there today. Say the word and I will add a section.

### How this was verified

Six new Go tests in `internal/server/clipboard`: deletes the requested ids and keeps the rest, skips pinned, ignores unknown ids, empty list is a no-op, the handler reports the deleted count, and five malformed `ids` shapes are rejected without touching the store. `go test ./...` passes for the whole core.

End to end: the built `dms` running against this working tree in a nested Hyprland session on its own XDG dirs. Populated six entries, pinned one, filtered, cleared, and checked the result with `dms cl history` — the two matching entries gone, the other four including the pinned one intact. Then filtered to a term that only the pinned entry matches and pressed `Shift+Del`: no dialog, nothing deleted. Then cleared with no filter: only the pinned entry left, modal closed, as before.

`make lint-qml` needs the Quickshell tooling VFS (`.qmlls.ini`), which this machine does not generate outside a full DMS install, so the touched QML was checked with Qt 6 `qmllint` directly — no errors beyond the pre-existing unresolved-import warnings.

### DIFF
--- a/core/internal/server/clipboard/handlers.go
+++ b/core/internal/server/clipboard/handlers.go
@@ -1,8 +1,10 @@
 package clipboard
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
+	"math"
 
 	clipboardstore "github.com/AvengeMedia/DankMaterialShell/core/internal/clipboard"
 	"github.com/AvengeMedia/DankMaterialShell/core/internal/server/models"
@@ -19,6 +21,8 @@ func HandleRequest(conn *models.Conn, req models.Request, m *Manager) {
 		handleGetEntry(conn, req, m)
 	case "clipboard.deleteEntry":
 		handleDeleteEntry(conn, req, m)
+	case "clipboard.deleteEntries":
+		handleDeleteEntries(conn, req, m)
 	case "clipboard.clearHistory":
 		handleClearHistory(conn, req, m)
 	case "clipboard.copy":
@@ -101,6 +105,71 @@ func handleDeleteEntry(conn *models.Conn, req models.Request, m *Manager) {
 	}
 
 	models.Respond(conn, req.ID, models.SuccessResult{Success: true, Message: "entry deleted"})
+}
+
+func handleDeleteEntries(conn *models.Conn, req models.Request, m *Manager) {
+	raw, ok := params.Any(req.Params, "ids")
+	if !ok {
+		models.RespondError(conn, req.ID, "missing 'ids' parameter")
+		return
+	}
+
+	list, ok := raw.([]any)
+	if !ok {
+		models.RespondError(conn, req.ID, "'ids' must be an array")
+		return
+	}
+
+	ids := make([]uint64, 0, len(list))
+	for _, item := range list {
+		id, err := toEntryID(item)
+		if err != nil {
+			models.RespondError(conn, req.ID, err.Error())
+			return
+		}
+		ids = append(ids, id)
+	}
+
+	deleted, err := m.DeleteEntries(ids)
+	if err != nil {
+		models.RespondError(conn, req.ID, err.Error())
+		return
+	}
+
+	models.Respond(conn, req.ID, map[string]int{"deleted": deleted})
+}
+
+// toEntryID accepts the shapes a clipboard entry id can arrive in: JSON decodes
+// numbers as float64, while Go callers and tests pass the integer types
+// directly.
+func toEntryID(value any) (uint64, error) {
+	switch v := value.(type) {
+	case float64:
+		if v < 0 || v != math.Trunc(v) {
+			return 0, fmt.Errorf("invalid entry id: %v", v)
+		}
+		return uint64(v), nil
+	case json.Number:
+		id, err := v.Int64()
+		if err != nil || id < 0 {
+			return 0, fmt.Errorf("invalid entry id: %v", v)
+		}
+		return uint64(id), nil
+	case int:
+		if v < 0 {
+			return 0, fmt.Errorf("invalid entry id: %v", v)
+		}
+		return uint64(v), nil
+	case int64:
+		if v < 0 {
+			return 0, fmt.Errorf("invalid entry id: %v", v)
+		}
+		return uint64(v), nil
+	case uint64:
+		return v, nil
+	default:
+		return 0, fmt.Errorf("invalid entry id: %v", value)
+	}
 }
 
 func handleClearHistory(conn *models.Conn, req models.Request, m *Manager) {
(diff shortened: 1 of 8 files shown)


Mario-Mohar:
Both points are addressed in `21cefbf`.

**Translator context** — added to the two strings this PR introduces, in the style CONTRIBUTING documents:

- `ClipboardHeader.qml:71` → `I18n.tr("Clear Filtered", "clipboard modal: clear button tooltip while a search filter is active")`
- `ClipboardHistoryContent.qml:150` → `I18n.tr("This will delete the %1 entries matching the current filter. Pinned entries are kept.", "clipboard modal: clear confirmation while a search filter is active, %1 is the number of matching entries")`

I left `"Clear All"` on that same ternary without a context on purpose: it is an existing shared term with four other call sites (`ClipboardHistoryModal`, `ClipboardHistoryPopout`, `ClipboardButton`, and `Settings/PrinterTab`), so a clipboard-specific comment there would end up attached to the term the printer tab uses too. Happy to add a neutral one if you would rather every term in the file carry a context.

**IPC help listing** — `clipboard.deleteEntries` is now in the listing in `server.go`, next to `deleteEntry`:

```
 clipboard.deleteEntries               - Delete entries by ID, skipping pinned ones (params: ids)
```

That is the only static method listing I found; `printIPCHelp()` in `cmd/dms/shell.go` enumerates Quickshell IPC targets at runtime and does not need an entry.

Verified: `gofmt` clean, `go build ./...`, `go vet ./internal/server/...`, `go test -count=1 ./internal/server/clipboard/...` pass, `go mod tidy` is a no-op, and the `i18n-term-freeze` / `i18n-term-variants` / `no-console-in-qml` hooks pass. I ran `extract_translations.py` only to confirm both new terms pick up their comment and then reverted the catalogs, so `en.json` and `template.json` are untouched.

Mario-Mohar:
@bbedward heads up on the red `lint-and-test` here — and a correction to my own `chore: re-trigger CI` commit message, where I called it an unrelated flake. That was too quick. It reproduced on `0fb9cc1e`, and in the last 30 Go CI runs the only two failures are both on this branch. So the link to this PR is real, even though the clipboard code isn't what's broken.

What dies is `TestStoreEntry_CorruptDBReturnsErrorNotPanic` (added in c68b7e5), and it dies with SIGBUS rather than a recovered panic:

```
=== RUN   TestStoreEntry_CorruptDBReturnsErrorNotPanic
unexpected fault address 0x7efcd0d48000
fatal error: fault
[signal SIGBUS: bus error code=0x2 addr=0x7efcd0d48000]
```

`corruptInlineBucketRoot` deliberately writes an out-of-range pgid — bbolt reports it as `Page expected to be: 16, but self identifies as 282584257676671`. bbolt recovers from its own assertion panics, but it can't recover from a fault when that offset lands on a page inside the mapping that has no file behind it. That's a SIGBUS, and the Go runtime turns it into a fatal error that takes the whole test binary with it.

Which means whether it faults is a property of the process memory layout, not of the code under test. Two things point that way: the bogus pgid is byte-identical across runs (`282584257676671`), while the fault address is not — `0x7f56ceb73000` on the first run, `0x7efcd0d48000` on the second. It tracks the mmap base, i.e. ASLR.

This PR inserts six test functions into `manager_test.go` at line 345, ahead of the corrupt-DB test in the same binary, which is enough to shift that layout. So I don't think this is my branch being special — the test looks latently unsafe on master, and any PR that adds tests to this package before it can trip the same way.

For what it's worth, the tests this PR adds all pass; the binary dies further down the run, after they've gone green.

Happy to send a separate PR against master for it. Options as I see them:

- have `corruptInlineBucketRoot` write a bogus-but-in-bounds pgid, so the failure stays an assertion panic bbolt can actually recover from
- run that one test in a subprocess and assert on the exit status
- drop it and lean on the `TestOpenDB_Heals*` cases, which already cover the salvage path without dereferencing a wild pointer

First option is the smallest and still covers what #3232 was about. Tell me which you'd rather have and I'll open it.

Issue #2982, opened 2026-08-02
Lock screen: pressing Enter never tries the security key, even with u2fMode: "or"


0ldMonk:
### Compositor

Niri

### Distribution

Fedora

### Select your Installation Method

Distro Packaging

### Was this your original Installation method?

Yes

### dms doctor -vC

<details>
<summary>Click to expand</summary>
## DMS Doctor Report

**System**
- [ok] Operating System: Fedora Linux 44.20260801.0 (Forty Four)
  - ID: fedora, Version: 44, Arch: amd64
- [ok] Architecture: amd64
- [ok] Display Server: Wayland
  - WAYLAND_DISPLAY=wayland-1

**Versions**
- [ok] DMS CLI: dms v1.5.3
  - /usr/bin/dms
- [ok] Quickshell: Quickshell 0.3.0 (revision 43d4fa9e883cb03239b3d578c9c57070f4fbd281, distributed by Fedora COPR (avengemedia/quickshell))
  - /usr/bin/qs
- [ok] DMS Shell: 1.5.3
  - /usr/share/quickshell/dms

**Installation**
- [ok] DMS Configuration: Found
  - /usr/share/quickshell/dms
- [ok] shell.qml: Present
  - /usr/share/quickshell/dms/shell.qml
- [info] Install Type: System package
  - /usr/share/quickshell/dms

**Compositor**
- [ok] niri: 26.04
  - /usr/bin/niri
- [info] Active: niri
- [ok] Background Blur: Supported
  - Compositor supports ext-background-effect-v1

**Quickshell Features**
- [ok] Polkit: Available
  - Authentication prompts
- [ok] IdleMonitor: Available
  - Idle detection
- [ok] IdleInhibitor: Available
  - Prevent idle/sleep
- [ok] ShortcutInhibitor: Available
  - Allow shortcut management (niri)
- [ok] BackgroundBlur: Available
  - Background blur API support in Quickshell

**Optional Features**
- [ok] accountsservice: Available
  - User accounts
- [ok] power-profiles-daemon: Available
  - Power profile management
- [ok] logind: Available
  - Session management
- [warn] cups-pk-helper: Not available (install cups-pk-helper)
  - Printer management
- [info] I2C/DDC: No monitors detected
  - External monitor brightness control
- [ok] qt6-imageformats: Installed (1 formats)
  - Formats: GIF (/usr/lib64/qt6/plugins/imageformats)
- [warn] kimageformats: Not installed
[...]

PR #3247, opened 2026-09-01
fix(lock): do not spend a PAM attempt on an empty password submit

fromelicks:
## Description

`Enter` button, on an empty password field calls `pam.passwd.start()`. The attempt cannot succeed, but it still consumes a `pam_faillock` attempt, so a stray `Enter` on the lock screen moves the session towards a real lockout for nothing.

Guard both submit paths with `canSubmitPassword()`. The guard is not unconditional: an empty submit is still allowed when the PAM stack is not DMS's own (`lockPamExternallyManaged`, or a custom `lockPamPath`), because there the conversation itself may prompt for an inline u2f or fingerprint touch and Enter is how the user starts it. The DMS-managed u2f path is unaffected, since #2982 was resolved with a dedicated `lockScreenSecurityKeyShortcut` rather than bare `Enter`.

## Type of change

- [x] Breaking change (fix or feature that changes existing behavior)
- [ ] Documentation
- [ ] Other

I'm not sure yet, whether this is a breaking change or not.

## Checklist

- [x] I have tested my changes locally (I've made them and use them for myself)
- [x] I have opened a corresponding pull request in dlx-docs to document any new behaviors: https://github.com/AvengeMedia/DankLinux-Docs

### DIFF
--- a/quickshell/Modules/Lock/LockScreenContent.qml
+++ b/quickshell/Modules/Lock/LockScreenContent.qml
@@ -106,6 +106,15 @@ Item {
         pam.u2f.startForAlternativeAuth();
     }
 
+    function canSubmitPassword() {
+        // Submitting an empty buffer spends a PAM attempt that cannot succeed,
+        // pushing the session towards a faillock lockout for nothing. The
+        // exception is a PAM stack DMS does not own: there the conversation
+        // itself may prompt (inline u2f/fprint), so an empty response is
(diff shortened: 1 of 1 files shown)


bbedward:
I feel like this was a deliberate action so empty password could trigger something else in the pam stack, like Howdy. Not sure if its needed though or if theres a more conventional way to trigger howdy in a pam config

PR #3294, opened 2026-09-04
fix(dash): dim today button when on current month

hthienloc:


hthienloc:
Depend on https://github.com/AvengeMedia/dank-qml-common/pull/5 to handle tooltip auto-hide on click/disabled, allowing us to use standard `DankActionButton.tooltipText` without a custom tooltip instance.

hthienloc:
Turned out the stuck tooltip was just caused by dynamically toggling tooltipText to null on disable which broke StateLayer.onExited. Keeping tooltipText static works fine without changes to dank-qml-common.

PR #3316, opened 2026-09-05
notifications: auto-enable do not disturb while screen sharing

### LINKED ISSUE #1482: Do not disturb while screen sharing
### Feature Description

I would like to be able to configure to automatically have "Do not disturb" enabled while I am screen sharing.

### Use Case

While I'm sharing my screen, I would like to avoid notifications to appear which may contain sensitive information (i.e from chats).

Only the small dot on the notification widget should appear.

Perhaps, if possible, it would be neat to have the notifications popup when I stop screen sharing.

### Compositor(s)

- [x] All compositors
- [ ] Niri
- [ ] Hyprland
- [ ] MangoWC (dwl)
- [ ] Sway
- [ ] Other (specify below)

### Proposed Solution

A toggle under the "Do Not Disturb" section in DMS settings which reads something like "Automatically enable while screen sharing".

Not sure if it's possible to implement, but Niri has the ability to block certain apps from screen sharing. If it was possible to block out the notification popups from screen sharing that would also be cool. Though I'm not sure if the other compositors have this ability, so a more universal approach would be to just (temporarily if possible) suppress the notifications.

avesst:
## Description

Automatically enable Do Not Disturb while a screen share is active, so notification popups (and their contents) do not appear on the shared screen. The unread dot on the notification widget still appears. Existing DND bypasses (critical priority, per-app allow list) keep working.

The setting is under **Settings → Notifications → Do Not Disturb**: **Automatically enable while screen sharing**. It is on by default.

Detection is the same as the existing screen-sharing privacy indicator (Niri cast IPC, otherwise Pipewire / xdg-desktop-portal), so this applies on every compositor DMS already supports.

### Behavior

Auto-DND turns the normal DND switch on for the share, then turns it off shortly after the share ends — unless the user already owned DND. The release is deliberately delayed by about four seconds; the reasoning is below the table.

| Case | What happens |
| --- | --- |
| Share starts, DND was off | DND turns on for the share. Popups are suppressed. The unread dot still shows. |
| Share ends, we turned DND on | DND turns off about four seconds later. |
| DND already on (indefinite) before the share | Left alone. Stays on after the share ends. |
| Timed DND already on before the share | Left alone. The timer keeps running after the share ends. |
| User picks **Until I turn it off** during a share | User owns DND. It stays on after the share ends. |
| User sets a duration during a share | User owns DND. The timer survives the end of the share. |
| User turns DND off during a share | Treated as a dismiss. DND stays off for the rest of that share, and across a shell reload. A share started after that one has ended auto-enables again. |
| Timed DND expires while still sharing | Auto-DND takes over for the rest of the share, then releases about four seconds after it ends. |
| Setting is off, then a share starts | DND is not auto-enabled. |
| Setting turned off while auto-DND is holding | Hold is released and DND turns off immediately, with no delay. |
| Setting turned on while already sharing (DND off) | DND turns on for the rest of the share. |
| Restart mid-share | DND stays on if the share is still active. If the share ended while the shell was down, DND is released about four seconds after start. |
| Capture stopped and started again quickly | DND stays on. The short gap is ignored. |
| Critical / per-app DND allow list | Unchanged. Those popups still appear during a share. |

### Why the delay on release

Detection is asynchronous and neither source announces when it has finished reporting. Quickshell's Pipewire API exposes only `ready`, which means the connection is up, not that nodes have been enumerated; niri cast state cannot arrive until compositor detection has resolved. So a share reading as gone is only acted on once it stays gone for four seconds. Turning DND on stays immediate, and turning the setting off releases immediately, because that is a user action rather than a detection guess.

Measured here, the niri path (compositor detection, socket connect, first cast snapshot) takes about 150 ms. Four seconds also clears the 3 s timeout `detectCompositor` uses before it falls back to env probing.

A manual DND-off during a share is remembered against the sources that were sharing at the time (niri `stream_id`, Pipewire node id), so it holds for that share and survives a shell reload, while a newly started share gets a fresh decision.

### Cost

With the setting on, `SessionData` instantiates `PrivacyService` at startup, which opens a `PwObjectTracker` over every non-stream Pipewire node. Until now `PrivacyService` was only created when the privacy indicator bar widget was enabled. That is the cost of the feature being active, and it is only paid when the setting is on.

## Type of change

**NOTE:** Default on might be considered breaking here, I don't know. I opted for on by default because of privacy reasons, I suspect most users would benefit from this, though we can change if maintainers feel like it. The idle cost above is the other side of that decision.

## Related issues

Closes #1482

## Screenshots / video

image

## Checklist

- [ ] Go changes: n/a
- [x] QML changes: `make lint-qml`

### DIFF
--- a/quickshell/Common/SessionData.qml
+++ b/quickshell/Common/SessionData.qml
@@ -36,6 +36,9 @@ Singleton {
     property bool isLightMode: false
     property bool doNotDisturb: false
     property real doNotDisturbUntil: 0
+    property bool doNotDisturbHeldByScreenShare: false
+    property bool screenShareDndDismissed: false
+    property var screenShareDismissedIds: []
     property bool idleInhibited: false
     property string terminalOverride: ""
     property bool isSwitchingMode: false
@@ -71,7 +74,15 @@ Singleton {
         id: dndExpireTimer
         repeat: false
         running: false
-        onTriggered: root.setDoNotDisturb(false)
+        onTriggered: root.setDoNotDisturb(false, 0, "timer")
+    }
+
+    Timer {
+        id: screenShareReleaseTimer
+        interval: 4000
+        repeat: false
+        running: false
+        onTriggered: root._endScreenShareDnd()
     }
 
     function _armDndExpireTimer() {
@@ -80,7 +91,7 @@ Singleton {
             return;
         const remaining = doNotDisturbUntil - Date.now();
         if (remaining <= 0) {
-            setDoNotDisturb(false);
+            setDoNotDisturb(false, 0, "timer");
             return;
         }
         dndExpireTimer.interval = remaining;
@@ -108,6 +119,21 @@ Singleton {
             root.suppressOSD = true;
             osdSuppressTimer.restart();
             root._applyDndExpirySanity();
+            Qt.callLater(root._syncScreenShareDnd);
+        }
+    }
+
+    Connections {
+        target: (!root.isGreeterMode && typeof SettingsData !== "undefined" && SettingsData.notificationDndWhileScreenSharing && typeof PrivacyService !== "undefined") ? PrivacyService : null
+        function onScreensharingActiveChanged() {
+            root._syncScreenShareDnd();
+        }
+    }
+
+    Connections {
+        target: typeof SettingsData !== "undefined" ? SettingsData : null
+        function onNotificationDndWhileScreenSharingChanged() {
+            root._syncScreenShareDnd();
         }
     }
 
@@ -309,6 +335,7 @@ Singleton {
                 Theme.generateSystemThemesFromCurrentTheme();
 
             loaded();
+            Qt.callLater(root._syncScreenShareDnd);
 
             _checkSessionWritable();
         } catch (e) {
@@ -390,6 +417,7 @@ Singleton {
                 Theme.generateSystemThemesFromCurrentTheme();
 
             loaded();
+            Qt.callLater(root._syncScreenShareDnd);
         } catch (e) {
             _parseError = true;
             const msg = e.message;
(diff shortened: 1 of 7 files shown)


avesst:
Rewrote the release path and squashed the branch, so the earlier inline review comments point at code that no longer exists. Summary of what changed and why.

Both review findings were valid. I fixed them first by tightening `screensharingReady`, but every version of that predicate leaked somewhere:

- `Pipewire.ready` only means the connection is up, not that nodes have been enumerated, so on Hyprland/sway a reload mid-share could still release the hold before the scan saw anything.
- On niri, cast state cannot arrive before compositor detection resolves, and older niri builds never send cast events at all, so requiring them left DND stuck on with no way out.
- Quickshell's Pipewire API exposes `nodes`, `links`, `linkGroups`, the defaults, `ready` and the add/remove signals. There is no "initial enumeration complete" signal, and `onNodeAdded` cannot tell a pre-existing node from a new one. The question a readiness flag has to answer is not answerable through that API.

So readiness is gone. A share reading as gone is now acted on only after it stays gone for four seconds. Engaging is still immediate, and turning the setting off releases immediately, since that is a user action and not a detection guess. `NiriService` and `CompositorService` are untouched by this PR again; `PrivacyService` gains one function.

One other change came out of it: a manual DND-off during a share is now persisted and scoped to the sources that were sharing at the time. Before, a shell reload dropped it and auto-DND re-armed against the user.

Retested: share start/end, restart mid-share with the share live and with it already ended, setting toggled on and off mid-share, timed DND expiring during a share, manual dismiss surviving a reload, dismiss clearing for the next share, and a capture stopped and restarted inside the grace window.

The default-on question and the idle cost it implies are described in the updated PR body. Happy to flip the default if maintainers prefer opt-in.

avesst:
Do you need anything more from me on this?

PR #3335, opened 2026-09-06
dankbar: add opt-in fullscreen auto-hide on niri

steven12138:
## Description

Add a per-bar **Hide Bar on Fullscreen** option for niri overlay bars. It is disabled by default and never changes the bar's layer. With it enabled, fullscreen entry and exit use the existing auto-hide animation, including the configured theme duration and disabled-animation behavior.

Fullscreen detection follows the active window of each output's visible workspace, so moving keyboard focus to another output does not by itself reveal the bar. Edge hover, popout pinning (including strict auto-hide), IPC reveal/tuck, and overview retain their existing policies. Fullscreen hiding preserves the configured exclusive zone rather than resizing background windows. Frame-hosted/non-overlay bars and other compositors do not opt into this behavior.

The first commit fixes prerequisite niri workspace-state bugs: numeric workspace IDs were compared to string object keys, fresh active-window snapshots could be overwritten with stale values, and active-window events on an unfocused output could change global focus flags.

Niri IPC and Wayland toplevels have no shared window ID. The matcher uses exact app ID/title on the output, with keyboard focus as a disambiguator. If identical windows remain ambiguous, it leaves the bar available instead of guessing. The existing dock detector is unchanged. No polling, extra processes, or new timers are added.

## Type of change

- [x] Bug fix
- [x] New feature

## Related issues

Supersedes #3334, which is closed.

## Validation

- 15 passing regression tests: per-output focus, workspace switching, window removal, fullscreen changes, duplicate titles, missing data, and workspace event handling. Run with `node --test quickshell/scripts/test-niri-fullscreen.mjs quickshell/scripts/test-niri-workspace-state.mjs` (no npm dependencies).
- Live niri 26.04 / DMS 1.6.0 session: default settings retain the Top layer. With both options enabled, verified intermediate slide positions on fullscreen entry/exit, IPC reveal/tuck while global auto-hide is off, overview entry/exit, and an unchanged exclusive zone. Temporary diagnostic instrumentation was removed afterward.
- `make lint-qml` reports the same existing `DMSShell.qml` warnings on the base and patched trees; normalized output is identical. `git diff --check` and translation term-variant checks pass. Settings search index regenerated.
- Physical multi-monitor, pointer-edge interaction and browser F11/video requests have not been separately exercised. Multi-monitor matching has regression-fixture coverage.

## Screenshots / video

Live transition progress was checked through temporary instrumentation; no desktop screenshots are attached.

## Checklist

- [x] Code follows CONTRIBUTING.md conventions.
- [x] Tested locally as described above.
- [x] New user-facing strings use `I18n.tr()`; translation catalogs were not edited.
- [x] Ran `make lint-qml` with no new warnings compared with the base.
- [x] Behavior, configuration and matching limits documented in `docs/fullscreen-bar.md`.
- [ ] Separate dlx-docs PR (documentation is included here).


quickshell/scripts/test-niri-fullscreen.mjs
```diff
@@ -0,0 +1,119 @@
+import assert from 'node:assert/strict';
```
bbedward, line 1: These should be under `quickshell/tests`

I know test fixutres dont exist yet but they exist in the upcoming feature branch so that'll be the location they go in eventually.

bbedward (changes requested):

steven12138:
Addressed the requested changes and rebased onto the current `master`:

- Moved the Node tests to `quickshell/tests/*.test.mjs`; the current `sync-shell` staging excludes the entire tests directory, and the embed exclusion test passes.
- Fullscreen-only hiding now uses the effective `topBarCore.autoHide` state for edge hover and click-through input, so the pointer edge can reveal the bar when global auto-hide is off.
- Updated the bar IPC regression fixture and added coverage for `toggleReveal` during fullscreen auto-hide.
- Resolved the upstream settings, bar, compositor, and niri service conflicts while preserving the current APIs.

Validation: 19 focused tests pass, the full logic suite passes, Qt tests pass (26/26), settings search index and i18n variant checks pass, `git diff --check` passes, embed staging contains no tests, and `make build` succeeds.

Issue #1756, opened 2026-02-20
Add an OSD for workspace change


ernestowgg:
### Feature Description

Just that: an optional OSD showing the workspace id / name when switching to another workspace. 

### Use Case

This would be good for those of us who keep the Workspace Switcher clean (no icons, no id), as a simple reminder or confirmation of which workspace we are now on. 

### Compositor(s)

- [x] All compositors
- [ ] Niri
- [ ] Hyprland
- [ ] MangoWC (dwl)
- [ ] Sway
- [ ] Other (specify below)

purian23:
Completed and closed out by last commit d0f47bb316a9392c7422b2731fd71c9a76af42cb

PR #3337, opened 2026-09-06
feat(osd): add workspace switch OSD for niri (#1756)

hthienloc:
Ref #1756

https://github.com/user-attachments/assets/0b7c0168-8e27-4e4b-acc2-e552610eb274

### DIFF
--- a/quickshell/Common/SettingsData.qml
+++ b/quickshell/Common/SettingsData.qml
@@ -1155,6 +1155,7 @@ Singleton {
     property bool osdCapsLockEnabled: true
     property bool osdPowerProfileEnabled: true
     property bool osdAudioOutputEnabled: true
+    property bool osdWorkspaceEnabled: false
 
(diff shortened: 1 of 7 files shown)


purian23:
Cool feature! One thing I'd like to see is full parity support for all 3 first party compositors. Niri, Hyprland and Mango. Doing it initially is much easier than back-porting later on when someone notices it's on one, but not the other. 

Niri will naturally have that bit of extra as our comp of choice, but for the most part keeping each in mind as we go is less work in the future. Totally understandable if you don't have each of them installed, we can get these added after the merge, but again, this is mostly for awareness.

hthienloc:
I've never used Mango or Hyprland before, so I can't test them as thoroughly as I can with Niri. I'd prefer someone with experience using those two compositors to handle the backporting and make sure everything works as expected.

PR #3384, opened 2026-09-10
network: support fortinet vpn with saml and password auth

### LINKED ISSUE #1614: Support openfortivpn
### Feature Description

I would like to be able to import `openfortivpn` configurations for easy access using DMS, with support for `--saml-login`.

### Use Case

We use FortiVPN at work.

### Compositor(s)

- [x] All compositors
- [ ] Niri
- [ ] Hyprland
- [ ] MangoWC (dwl)
- [ ] Sway
- [ ] Other (specify below)

### Proposed Solution

I dont't have any particular proposals but here is an example of starting `openfortivpn` using `--saml-login`:

```
$ sudo openfortivpn --saml-login -c ~/.config/openfortivpn/myvpn
INFO:   Listening for SAML login on port 8020
INFO:   Authenticate at 'https://vpn.domain.com:443/remote/saml/start?redirect=1'
```

Today I use a custom built script which starts `openfortivpn` and pipes it to `read`, extracts the URL and opens it in my browser.

A similar implementation would be very nice.

Thank you in advance.

_No response_

avesst:
## Description

Adds Fortinet VPN support to the NetworkManager openconnect path, covering both SAML/SSO and password authentication, and lets openfortivpn config files be imported as profiles.

FortiGate gateways configured for SAML can't be used with a username and password — authentication happens in a browser against an external IdP. DMS previously detected that a Fortinet profile needed external browser auth, logged that only GlobalProtect was supported, and fell back to prompting for credentials that don't exist.

**Connecting.** A Fortinet profile with `authtype=saml` now runs the login in core:

1. Probe the gateway certificate and, if it isn't trusted, ask the user once.
2. Listen on loopback (8020 by default, `saml-port` overrides) and open `/remote/saml/start` in the browser.
3. Wait for the gateway to redirect back with a session id.
4. Exchange the id for an `SVPNCOOKIE` over a connection pinned to the probed certificate.

The cookie is handed to the openconnect plugin through the existing secret agent, the same way GlobalProtect already works, so this needs no root and leaves no long-lived process behind. The listener is torn down as soon as the login completes.

A Fortinet profile with `authtype=password` goes to the openconnect password flow that already existed, unchanged.

**Importing.** `ImportVPN` reads openfortivpn configs and builds an openconnect profile from `host`, `port`, `username`, `password`, `realm`, `trusted-cert` and `saml-login`. A `saml-login` line selects the browser login; anything else is imported as a password profile carrying whatever credentials the file had, and the connect prompt asks for whichever of username and password is missing. A password alongside `saml-login` is dropped, since that path never reads one.

openfortivpn configs are conventionally extensionless, so detection reads the contents rather than the extension, and anything that doesn't parse falls through to the existing nmcli importers untouched. For the same reason the two VPN import file browsers now offer `*` alongside the per-plugin extensions.

**Certificates.** Trust is stored as the whole-certificate digest that openfortivpn's `trusted-cert` uses, so an existing config's trust line is honoured on the first connect and a newly approved certificate stays compatible with openfortivpn. Approving a certificate appends to that list rather than replacing it, which keeps gateways that answer from several nodes working. openconnect separately gets the public-key pin it expects for `--servercert`.

**Known limitation.** Certificate trust is only persisted when the connection was started from DMS. The approval is queued into the same pending-save slot the existing credential flow uses, and that slot is drained by a handler which returns early unless DMS initiated the activation. Started from nmcli, the applet, or autoconnect, the user approves a certificate and the trust is discarded, so the prompt returns on the next connect. This gap predates the PR and affects saved VPN credentials the same way, so I've left the shared behavior alone rather than changing it from a Fortinet feature. Happy to fix it here or in a follow-up, whichever you prefer.

**Notes for reviewers.** This path requires the NetworkManager **openconnect** plugin, which the import checks for up front and reports if missing. The tunnel is openconnect's Fortinet implementation, not openfortivpn — the config file is only a settings source and nothing about openfortivpn runs.

The loopback listener accepts the session id from any request on its port, matching what openfortivpn's own `--saml-login` server does. A local page racing it during a login costs a retry: the bogus id goes to the real gateway over the pinned connection, is rejected, and the IdP session is still live. Happy to tighten it if you'd rather not carry that.

Implementation written with Claude Code; I directed the design, tested it end to end against a production FortiGate gateway, and reviewed the changes — but Go and VPN internals aren't my area, so I'd appreciate scrutiny of the certificate handling and the secret-agent path.

## Type of change

## Related issues

Closes #1614

## Screenshots / video

**Certificate trust prompt.** Shown once per gateway certificate, before the browser opens. Fingerprint redacted.

image

**Loopback completion page.** Served by core once the gateway redirects the browser back with the session id. The IdP login in between is whatever the gateway's organisation has configured, so there is nothing DMS-specific to show there.

## Checklist

  - No new QML strings. The one new label is a secret-agent field name in Go, added to the existing table alongside `GlobalProtect SAML/SSO` and following the same pattern.
- [x] I have opened a corresponding pull request in dlx-docs to document any new behaviors: AvengeMedia/DankLinux-Docs#146
  - Adds a VPN page covering the browser login, the openfortivpn import rules, the openconnect plugin requirement and certificate trust, and corrects the note that said native Fortinet authentication applied only to password profiles. It asks not to be merged until this lands.

### DIFF
--- a/core/internal/server/network/agent_networkmanager.go
+++ b/core/internal/server/network/agent_networkmanager.go
@@ -315,6 +315,38 @@ func (a *SecretAgent) GetSecrets(
 		}
 		a.backend.cachedOpenConnectMu.Unlock()
 
+		if len(fields) == 1 && fields[0] == "fortinet-saml" {
+			data := vpnDataFromSettings(conn)
+
+			log.Infof("[SecretAgent] Starting Fortinet SAML authentication for gateway=%s", data["gateway"])
+
+			samlCtx, samlCancel := context.WithTimeout(context.Background(), 5*time.Minute)
+			defer samlCancel()
+
+			authResult, err := a.backend.authenticateFortinetSAML(samlCtx, fortinetAuthTarget{
+				ConnName:       displayName,
+				ConnUUID:       connUuid,
+				ConnectionPath: string(path),
+				VpnService:     vpnSvc,
+				Data:           data,
+			})
+			if err != nil {
+				log.Warnf("[SecretAgent] Fortinet SAML authentication failed: %v", err)
+				return nil, dbus.MakeFailedError(fmt.Errorf("SAML authentication failed for the Fortinet gateway: %w", err))
+			}
+
+			a.backend.cachedOpenConnectMu.Lock()
+			a.backend.cachedOpenConnectAuth = &cachedOpenConnectAuth{
+				ConnectionUUID: connUuid,
+				Cookie:         authResult.Cookie,
+				Host:           authResult.Host,
+				Fingerprint:    authResult.Fingerprint,
+			}
+			a.backend.cachedOpenConnectMu.Unlock()
+
+			return buildOpenConnectSecretsResponse(settingName, authResult.Cookie, authResult.Host, authResult.Fingerprint), nil
+		}
+
 		if len(fields) == 1 && fields[0] == "gp-saml" {
 			gateway := ""
 			protocol := ""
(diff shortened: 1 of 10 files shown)


avesst:
@bbedward Let me know if your content with the implementation, and I'll add a PR for the docs.

bbedward:
Looks good, go ahead and add the docs PR if you think its helpful

avesst:
@bbedward PR opened.

PR #3385, opened 2026-09-10
fix(battery): report when no charge threshold exists instead of faking success

### LINKED ISSUE #3278: Battery charge limit: "Apply to Hardware" falsely reports success when no writable sysfs threshold exists (Lenovo IdeaPad uses charge_types/Long_Life)
### Compositor

Niri

### Distribution

Fedora

### If Other, please specify

_No response_

### Select your Installation Method

Distro Packaging

### Was this your original Installation method?

_No response_

### If no, specify

_No response_

### dms doctor -vC

<details>
<summary>Click to expand</summary>

## DMS Doctor Report

**System**
- [ok] Operating System: Fedora Linux 44 (Forty Four)
  - ID: fedora, Version: 44, Arch: amd64
- [ok] Architecture: amd64
- [ok] Display Server: Wayland
  - WAYLAND_DISPLAY=wayland-1

**Versions**
- [ok] DMS CLI: dms v1.6.0
  - /usr/bin/dms
- [ok] Quickshell: Quickshell 0.3.1 (revision , distributed by Fedora COPR (avengemedia/quickshell))
  - /usr/bin/qs
- [ok] DMS Shell: v1.6.0
  - /run/user/UID/danklinux-shell/HASH

**Installation**
- [ok] DMS Configuration: Found
  - /run/user/UID/danklinux-shell/HASH
- [ok] shell.qml: Present
  - /run/user/UID/danklinux-shell/HASH/shell.qml
- [info] Install Type: Unknown
  - /run/user/UID/danklinux-shell/HASH

**Compositor**
- [ok] niri: 26.04
  - /usr/bin/niri
- [info] Active: niri
- [ok] Background Blur: Supported
  - Compositor supports ext-background-effect-v1

**Quickshell Features**
- [ok] Polkit: Available
  - Authentication prompts
- [ok] IdleMonitor: Available
  - Idle detection
- [ok] IdleInhibitor: Available
  - Prevent idle/sleep
- [ok] ShortcutInhibitor: Available
  - Allow shortcut management (niri)
- [ok] BackgroundBlur: Available
  - Background blur API support in Quickshell

**Optional Features**
- [ok] accountsservice: Available
  - User accounts
- [ok] power-profiles-daemon: Available
  - Power profile management
- [ok] logind: Available
  - Session management
- [ok] cups-pk-helper: Available
  - Printer management
- [info] I2C/DDC: No monitors detected
  - External monitor brightness control
- [ok] qt6-imageformats: Installed (5 formats)
  - Formats: WebP, TIFF, GIF, JP2, ICNS (/usr/lib64/qt6/plugins/imageformats)
- [ok] kimageformats: Installed (4 formats)
  - Formats: AVIF, HEIF, JXL, EXR (/usr/lib64/qt6/plugins/imageformats)
- [ok] Terminal: kitty
- [ok] Network: NetworkManager
  - NetworkManager present. Using NM API.
- [ok] matugen: Installed
  - Dynamic theming
- [ok] cava: Installed
  - Audio visualizer
- [info] khal: Not installed
  - Calendar events
- [ok] danksearch: Installed
  - File search
- [info] dankcalendar: Not installed
  - Calendar app
- [info] fprintd: Not installed
  - Fingerprint auth
- [ok] adw-gtk3: Installed
  - GTK3 dynamic theming (system copy, applied on next theme apply)

**Config Files**
- [ok] settings.json: Present
  - /home/USER/.config/DankMaterialShell/settings.json
- [info] clsettings.json: Not yet created
  - /home/USER/.config/DankMaterialShell/clsettings.json
- [info] plugin_settings.json: Not yet created
  - /home/USER/.config/DankMaterialShell/plugin_settings.json
- [ok] session.json: Present
  - /home/USER/.local/state/DankMaterialShell/session.json
- [ok] dms-colors.json: Present
  - /home/USER/.cache/DankMaterialShell/dms-colors.json

**Services**
- [ok] dms.service: enabled, active
- [ok] greetd: enabled

**Environment**
- [info] QT_QPA_PLATFORMTHEME: qt6ct
- [ok] qt6ct plugin (Qt6): Installed
  - /usr/lib64/qt6/plugins/platformthemes/libqt6ct.so
- [info] QT_QPA_PLATFORMTHEME_QT6: Not set
- [info] QS_ICON_THEME: Not set
- [warn] XDG_MENU_PREFIX: Not set
  - Dolphin file associations and 'Open with…' dialog may be empty. Set XDG_MENU_PREFIX=plasma- in your compositor's environment block.

**Fonts**
- [ok] Normal Font: Rubik
  - Available via fontconfig
- [ok] Monospace Font: JetBrainsMono Nerd Font Mono
  - Available via fontconfig

---
**Summary:** 0 error(s), 1 warning(s), 35 ok

</details>

## Summary

On hardware that exposes **no writable numeric charge-threshold file** in sysfs, Settings → Battery → **Apply to Hardware** writes nothing, yet DMS shows the success toast *"Charge limit applied successfully — Limit set to N%"*. The limit is silently not applied and the user is told that it was.

## Root cause

`quickshell/Modules/Settings/BatteryTab.qml` (lines 19–40 on master @ `1f22ad4`):

```qml
Process {
    id: applyLimitProcess
    command: ["pkexec", "sh", "-c", "
for bat in /sys/class/power_supply/BAT*; do
  if [ -f \"$bat/charge_control_limit_max\" ]; then
    ...
  elif [ -f \"$bat/charge_stop_threshold\" ]; then
    ...
  elif [ -f \"$bat/charge_control_end_threshold\" ]; then
    ...
  fi
done
"]
    running: false
    onExited: exitCode => {
        if (exitCode !== 0) {
            ToastService.showError(...)
        } else {
            ToastService.showInfo(I18n.tr("Charge limit applied successfully"), ...)
        }
    }
}
```

If none of the three `if`/`elif` branches match, the `for` loop completes having written nothing and `sh` exits **0**. `onExited` treats every exit code 0 as success, so "nothing was applied" and "the limit was applied" are indistinguishable to the UI.

## What this machine actually exposes

Lenovo IdeaPad Slim 3 15IRH8, `ideapad_laptop` driver, kernel 7.1.12:

```console
$ ls /sys/class/power_supply/BAT0/
alarm  capacity  capacity_level  charge_types  cycle_count  device  energy_full
energy_full_design  energy_now  extensions  hwmon2  manufacturer  model_name
power  power_now  present  serial_number  status  subsystem  technology  type
uevent  voltage_min_design  voltage_now
```

None of `charge_control_limit_max`, `charge_stop_threshold` or `charge_control_end_threshold` exist. What is exposed instead is the power-supply extension property `charge_types`:

```console
$ cat /sys/class/power_supply/BAT0/charge_types
Fast Standard [Long_Life]

$ grep CHARGE_TYPES /sys/class/power_supply/BAT0/uevent
POWER_SUPPLY_CHARGE_TYPES=Long Life

$ ls -l /sys/class/power_supply/BAT0/extensions/
ideapad_laptop -> ../../../../../../../../../pci0000:00/0000:00:1f.0/PNP0C09:00/VPC2004:00
```

The kernel explicitly points at this as the current interface, superseding the old platform attribute:

```
ideapad_acpi VPC2004:00: conservation_mode attribute has been deprecated, see charge_types.
```

`Long_Life` works correctly when written directly — `echo Long_Life | sudo tee /sys/class/power_supply/BAT0/charge_types` switches the mode, and the deprecated `conservation_mode` mirror flips `0` → `1`. **The resulting cap is firmware/EC-controlled and is not settable as a number**; the exact percentage varies by model. So DMS's 50–100 % slider has no meaningful mapping on this class of hardware — but the underlying protection feature does exist and DMS simply has no support for it.

## Suggested fixes (two separable changes)

**1. Immediate — correctness of the apply path.** When no supported threshold file is found, the script should exit non-zero and DMS should surface an *unsupported hardware* message instead of success. For example, track whether any branch matched and `exit 1` (or a distinct code) otherwise, so the UI can tell "no supported interface on this machine" apart from "write failed". Ideally the Apply button and the slider would also be disabled or hidden when no supported interface is detected, so the setting cannot be configured in a way that can never take effect.

**2. Follow-up — support `charge_types` hardware.** Lenovo IdeaPad/ThinkBook and similar models expose `/sys/class/power_supply/BAT*/charge_types` (`Fast` / `Standard` / `Long_Life`) rather than a numeric threshold. Because this is a *mode selector* and not a percentage, it would need its own control — a three-way selector or a "battery conservation mode" toggle — rather than the percentage slider, with a note that the actual cap is firmware-defined.

## Related, but not a duplicate

- #2867 — that report is about the displayed percentage being rescaled against `batteryChargeLimit` on a machine where the numeric threshold *does* work. This report is about the apply path silently no-op'ing on hardware that has **no** numeric threshold at all.

## Note

The "Charge Limit Reached" notification path in `quickshell/Services/BatteryService.qml` is independent of this and works fine — but it is only a notification and does not stop charging, so it does not compensate for the silent no-op.

### Expected Behavior

DMS detects that no supported charge-threshold interface exists on the machine and reports that clearly — an error/warning toast such as *"Charge limit not supported on this hardware"* — instead of reporting success. The percentage slider and Apply button should ideally be disabled when no supported interface is present.

### Steps to Reproduce

1. Use a machine that exposes **none** of `charge_control_limit_max`, `charge_stop_threshold`, `charge_control_end_threshold` under `/sys/class/power_supply/BAT*/` — e.g. a Lenovo IdeaPad on the `ideapad_laptop` driver, which exposes `charge_types` instead. Verify with `ls /sys/class/power_supply/BAT0/`.
2. Open Settings → Battery → **Protection** and set **Battery Charge Limit** to e.g. `80`.
3. Click **Apply to Hardware** and authenticate at the polkit prompt.
4. Observe the toast: *"Charge limit applied successfully — Limit set to 80%"*.
5. Re-inspect sysfs: nothing has been written, no threshold file was created, and `charge_types` is unchanged.
6. Plug in the charger: the battery charges past the configured limit all the way to 100%.

### Error Messages/Logs

There is no error output — that is precisely the problem. The `pkexec sh -c` process runs, matches none of the three `if`/`elif` branches, writes nothing, produces no stderr, and exits with code `0`, which DMS interprets as success.

For reference, the interface the machine does expose:

```
ideapad_acpi VPC2004:00: conservation_mode attribute has been deprecated, see charge_types.
ACPI: battery: new hook: Ideapad Battery Extension
```

### Screenshots/Recordings

_No response_

Mario-Mohar:
## Description

On hardware that exposes none of the three sysfs charge threshold files, **Apply to Hardware** wrote nothing and still reported success. The script walks every `BAT*`, matches none of the three `if`/`elif` branches, and `sh` exits 0, which `onExited` could not tell apart from a real write.

This is not a rare corner. Lenovo IdeaPad and ThinkBook models on `ideapad_laptop` expose `charge_types` (`Fast` / `Standard` / `Long_Life`) instead, and the kernel says so itself: *"conservation_mode attribute has been deprecated, see charge_types"*. I have one of those machines here, so everything below was checked on the affected hardware rather than reasoned about.

Three changes:

**The script now says whether it wrote anything.** The `if`/`elif` chain became a loop over the same three names in the same order, so the priority does not change, with `found=1` set when a branch matches and `exit 2` when none did. A failed write still exits 1, so "this machine has no interface" and "the write failed" stay apart.

**The toast tells the truth.** Exit code 2 gives *"Charge limit not supported on this hardware"* instead of the success toast.

**The button is not offered when it cannot work.** A probe runs once when the tab loads and hides the Apply button if none of the three files exists, with a line naming the reason in its place. The exit code 2 branch stays as the authority, since the probe runs once and a battery can be removed or a driver reloaded after that.

**What I deliberately left alone:** the slider. The issue suggests disabling it too, but `batteryChargeLimit` also drives the "Charge Limit Reached" notification in `BatteryService.qml`, which works on this hardware and is independent of sysfs. Disabling the slider would break a working feature to tidy up a broken one.

**Not in this PR:** `charge_types` support. That interface is a mode selector, not a percentage, and the cap behind `Long_Life` is firmware defined, so it needs its own control and a decision about how it sits next to a 50-100 slider. That is the follow-up the issue describes as separable, and it seems better to agree on the shape of it first.

## Type of change

- [ ] Documentation
- [ ] Other

## Related issues

Fixes #3278

## Screenshots / video

Taken on the affected machine, a Lenovo IdeaPad whose `BAT0` has `charge_types` and none of the three threshold files, in a nested Hyprland running this branch.

protection card before and after

## Checklist

- [x] I have opened a corresponding pull request in dlx-docs to document any new behaviors: https://github.com/AvengeMedia/DankLinux-Docs

Notes on those boxes. Two new terms, `Charge limit not supported on this hardware` and `No writable charge threshold file was found under /sys/class/power_supply.`, both with a translator comment, and the second is shared by the toast and the inline line rather than duplicated. I ran `extract_translations.py` to confirm they land as expected and then reverted `en.json` and `template.json`, since those are yours to sync. No Go changes. `make lint-qml` wants a `.qmlls.ini` from a full install, so I ran `/usr/lib/qt6/bin/qmllint` on the file instead: nothing beyond the unresolved `qs.*` imports every file produces, and the `multiline-strings` count is unchanged from master, which is why the probe command is written on one line. No docs PR, the change removes a wrong message rather than adding behavior to document.

## How it was verified

The apply script and the probe were run against a fake `power_supply` tree as well as the real one:

| case | apply | probe |
| --- | --- | --- |
| this machine, only `charge_types` | 2, nothing written | 1, button hidden |
| `charge_control_end_threshold` present | 0, value written | 0 |
| two batteries, different file each | 0, both written | 0 |
| file present but not writable | 1, error toast | 0 |
| several files present | only `charge_control_limit_max` written, as before | 0 |
| no `BAT*` at all | 2 | 1 |

### DIFF
--- a/quickshell/Modules/Settings/BatteryTab.qml
+++ b/quickshell/Modules/Settings/BatteryTab.qml
@@ -16,22 +16,43 @@ Item {
         return Theme.surfaceText;
     }
 
+    // The sysfs names DMS can write, in the order they are tried. Hardware that
+    // exposes none of them, a Lenovo IdeaPad on ideapad_laptop for instance,
+    // used to run the apply script to completion without writing anything, and
+    // sh exiting 0 was indistinguishable from a successful write.
+    readonly property string thresholdFileList: "charge_control_limit_max charge_stop_threshold charge_control_end_threshold"
+    readonly property int noThresholdExitCode: 2
+
+    property bool chargeLimitSupported: true
+
+    Process {
+        id: thresholdProbe
+        running: Qt.platform.os === "linux"
+        command: ["sh", "-c", "for bat in /sys/class/power_supply/BAT*; do for file in " + root.thresholdFileList + "; do [ -f \"$bat/$file\" ] && exit 0; done; done; exit 1"]
+        onExited: exitCode => root.chargeLimitSupported = exitCode === 0
+    }
+
     Process {
         id: applyLimitProcess
         command: ["pkexec", "sh", "-c", "
+found=0
 for bat in /sys/class/power_supply/BAT*; do
-  if [ -f \"$bat/charge_control_limit_max\" ]; then
-    echo " + SettingsData.batteryChargeLimit + " > \"$bat/charge_control_limit_max\"
-  elif [ -f \"$bat/charge_stop_threshold\" ]; then
-    echo " + SettingsData.batteryChargeLimit + " > \"$bat/charge_stop_threshold\"
-  elif [ -f \"$bat/charge_control_end_threshold\" ]; then
-    echo " + SettingsData.batteryChargeLimit + " > \"$bat/charge_control_end_threshold\"
-  fi
+  for file in " + root.thresholdFileList + "; do
+    if [ -f \"$bat/$file\" ]; then
+      found=1
+      echo " + SettingsData.batteryChargeLimit + " > \"$bat/$file\" || exit 1
+      break
+    fi
+  done
 done
+[ \"$found\" = 1 ] || exit " + root.noThresholdExitCode + "
 "]
         running: false
         onExited: exitCode => {
-            if (exitCode !== 0) {
+            if (exitCode === root.noThresholdExitCode) {
+                root.chargeLimitSupported = false;
+                ToastService.showError(I18n.tr("Charge limit not supported on this hardware", "battery settings: toast title when sysfs has no charge threshold"), I18n.tr("No writable charge threshold file was found under /sys/class/power_supply.", "battery settings: why the charge limit could not be applied"));
+            } else if (exitCode !== 0) {
                 ToastService.showError(I18n.tr("Failed to apply charge limit to system"), I18n.tr("Process exited with code %1").arg(exitCode));
             } else {
                 ToastService.showInfo(I18n.tr("Charge limit applied successfully"), I18n.tr("Limit set to %1%").arg(SettingsData.batteryChargeLimit));
@@ -214,9 +235,18 @@ done
                     onSliderValueChanged: newValue => SettingsData.set("batteryChargeLimit", newValue)
                 }
 
+                StyledText {
+                    width: parent.width
+                    visible: Qt.platform.os === "linux" && !root.chargeLimitSupported
+                    text: I18n.tr("No writable charge threshold file was found under /sys/class/power_supply.", "battery settings: why the charge limit could not be applied")
+                    wrapMode: Text.WordWrap
+                    color: Theme.surfaceVariantText
+                    font.pixelSize: Theme.fontSizeMedium
+                }
+
                 Row {
                     // charge_control_* live in Linux sysfs; no BSD equivalent
-                    visible: Qt.platform.os === "linux"
+                    visible: Qt.platform.os === "linux" && root.chargeLimitSupported
                     width: parent.width
                     height: applyButton.height
                     layoutDirection: I18n.isRtl ? Qt.LeftToRight : Qt.RightToLeft
--- a/quickshell/Modules/Settings/BatteryTab.qml
+++ b/quickshell/Modules/Settings/BatteryTab.qml
@@ -237,7 +237,10 @@ done
 
                 StyledText {
                     width: parent.width
-                    visible: Qt.platform.os === "linux" && !root.chargeLimitSupported
+                    // A machine without a battery has no charge threshold to
+                    // write to either, but saying so there would be noise: the
+                    // limit was never applicable in the first place.
+                    visible: Qt.platform.os === "linux" && BatteryService.batteryAvailable && !root.chargeLimitSupported
                     text: I18n.tr("No writable charge threshold file was found under /sys/class/power_supply.", "battery settings: why the charge limit could not be applied")
                     wrapMode: Text.WordWrap
                     color: Theme.surfaceVariantText
@@ -245,8 +248,11 @@ done
                 }
 
                 Row {
-                    // charge_control_* live in Linux sysfs; no BSD equivalent
-                    visible: Qt.platform.os === "linux" && root.chargeLimitSupported
+                    // charge_control_* live in Linux sysfs; no BSD equivalent.
+                    // Without a battery the apply button cannot do anything,
+                    // and offering it anyway is what produced the false
+                    // success this change is about.
+                    visible: Qt.platform.os === "linux" && BatteryService.batteryAvailable && root.chargeLimitSupported
                     width: parent.width
                     height: applyButton.height
                     layoutDirection: I18n.isRtl ? Qt.LeftToRight : Qt.RightToLeft


Mario-Mohar:
Good catch on the batteryless case, fixed in 3835c2d.

The explanation was gated on `!chargeLimitSupported` alone, which is also false on a desktop or VM that never had a battery. There it reads as a fault report about hardware the machine was never going to have.

Both the explanation and the apply button now also require `BatteryService.batteryAvailable`. Hiding the button there is the same point this PR makes elsewhere: offering a control that cannot work is what produced the false success in the first place, and a machine without a battery has no charge threshold to write to either. `BatteryService.batteryAvailable` is already the gate used twice in this file for the same reason.

`make lint-qml` needs the Quickshell tooling VFS, which I did not want to spin up on this machine, so I ran `qmllint` on the file alone: no syntax errors, and the unqualified-access warning on the new line is the same one the file already produces 24 times for `BatteryService`.

Issue #1542, opened 2026-01-29
tray icons grouped together and hide all at once


grokXcopilot:
### Compositor

- [x] Niri
- [ ] Hyprland
- [ ] MangoWC (dwl)
- [ ] Sway

### Distribution

- [x] Arch Linux
- [ ] CachyOS
- [ ] Fedora
- [ ] NixOS
- [ ] Debian
- [ ] Ubuntu
- [ ] Gentoo
- [ ] OpenSUSE
- [ ] Other (specify below)

### dms doctor -v

Error: unknown command "doctor" for "dms"
Run 'dms --help' for usage.
 FATAL  go: unknown command "doctor" for "dms"

### Description

When multiple Electron apps are running with tray icons (for example 1Password and Vesktop), hiding one tray icon causes all tray icons to be hidden.

This seems to happen because all Electron tray icons use the same name: `chrom_status_icon_1`

### Steps to Reproduce & Installation Method

1. Open two or more Electron apps with tray icons (e.g. 1Password, Vesktop).
2. Verify that all tray icons are visible.
3. Hide the tray icon for one Electron app.
4. All Electron tray icons are hidden.

PR #3386, opened 2026-09-10
fix: stabilize tray item key to prevent position jumps on property changes

Lemon-mon-254:
## Problem

Tray icons (especially Fcitx) jump positions when switching windows because `getTrayItemKey()` uses `id::tooltipTitle` as the key. Fcitx5 updates `tooltipTitle` on input method changes, causing the key to change and fall back to end of list.

## Solution

Use only `id` for tray item key generation. The `StatusNotifierItem` specification guarantees `id` is unique per application, making `tooltipTitle` redundant for key purposes.

## Testing

- Verified with Fcitx5: no position jumps when switching input methods
- Verified with other tray apps: stable positions maintained

### DIFF
--- a/quickshell/Modules/DankBar/Widgets/SystemTrayBar.qml
+++ b/quickshell/Modules/DankBar/Widgets/SystemTrayBar.qml
@@ -48,6 +48,46 @@ BasePill {
         return `${id}::${tooltipTitle}`;
     }
 
+    function trayItemCountForId(id) {
+        if (!id)
+            return 0;
+        return root.allTrayItems.filter(other => (other?.id || "") === id).length;
+    }
+
+    function resolveOrderIndex(key, orderMap) {
+        if (orderMap.has(key))
+            return orderMap.get(key);
+        const id = key.includes("::") ? key.split("::")[0] : key;
+        if (root.trayItemCountForId(id) > 1)
+            return -1;
+        let fallbackIdx = -1;
+        for (const [savedKey, idx] of orderMap) {
+            if (savedKey === id || savedKey.startsWith(id + "::")) {
+                if (fallbackIdx >= 0)
+                    return -1;
+                fallbackIdx = idx;
+            }
+        }
+        return fallbackIdx;
+    }
+
+    function findSavedTrayKey(key, list) {
+        if (!list)
+            return null;
+        if (list.indexOf(key) !== -1)
+            return key;
+        const id = key.includes("::") ? key.split("::")[0] : key;
+        if (root.trayItemCountForId(id) > 1)
+            return null;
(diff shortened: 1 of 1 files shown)


quickshell/Modules/DankBar/Widgets/SystemTrayBar.qml
```diff
@@ -40,12 +40,7 @@ BasePill {
         });
     }
```
bbedward, line 41: Lets remove this function entirely and replace all of the call sites

bbedward (changes requested):

bbedward:
/claude review and identify all of the call sites of the getTrayItemKey function

Lemon-mon-254:
The previous approach (using plain `id` as the key) had clear downsides: it regressed duplicate-id disambiguation (#1542) and invalidated persisted `id::tooltipTitle` entries. This revision **keeps `getTrayItemKey()` unchanged** and only makes persisted-key lookups tolerant of dynamic `tooltipTitle`:

- Added `resolveOrderIndex()` to `sortByPreferredOrder()`: exact key match first; prefix matching against `id::*` entries only when the id is a **single instance** (keeps Fcitx's position stable while its `tooltipTitle` changes — old `Fcitx::…` data still resolves).
- Added `findSavedTrayKey()` / `isTrayIdHidden()` to replace all direct `SessionData.isHiddenTrayId()` calls (`visibleSortedTrayItems`, `manualHiddenBarItems`, `promoteTrayItemToBar`, `isManualHiddenTrayItem`, context-menu visibility toggle), and `showTrayId()` is now called with the actual saved key.
- Prefix fallback is gated by `trayItemCountForId()` to **single-instance ids only**; multiple instances sharing an id always use exact matching, so hiding/ordering one never affects the others.

Existing persisted data keeps matching on read — no migration needed.

Verified locally against the live `session.json`: Fcitx position stays stable across input-method switches, and hiding one of three duplicate-id items leaves the other two untouched.

Lemon-mon-254:
@bbedward Updated. `getTrayItemKey()` is back to untouched — no duplicate-id/key-format regressions. Lookups now exact-match first, prefix-match `id::*` only for single-instance ids, so Fcitx stays put when `tooltipTitle` changes. Would appreciate a fresh look.

Lemon-mon-254:
`getTrayItemKey` now returns the plain `id` for single-instance items, and `id::<occurrenceIndex>` when ids collide.

Verified locally with simulated SNI instances: single-instance key stays stable under changing tooltips; duplicate ids get independent `::0`/`::1`; leftover `duptest` no longer hides a survivor; migration normalizes `Fcitx::…` → `Fcitx`.

Issue #1235, opened 2026-01-01
Gamma settings not applied after wake from sleep [with patch]


philjackson:
### Compositor

- [ ] Niri
- [x] Hyprland
- [ ] MangoWC (dwl)
- [ ] Sway

### Distribution

- [x] Arch Linux
- [ ] CachyOS
- [ ] Fedora
- [ ] NixOS
- [ ] Debian
- [ ] Ubuntu
- [ ] Gentoo
- [ ] OpenSUSE
- [ ] Other (specify below)

### dms version

1.0.3

### Description

When waking from suspend it seems gamma settings aren't (re-)applied. E.G. I sleep my computer when the temp is 4500, seconds later when I wake it, the temp has reverted to 6500k.

Here's a rudimentary patch which seems to resolve the issue for me, not sure if it's a good approach though: [gamma-fix.diff.txt](https://github.com/user-attachments/files/24403273/gamma-fix.diff.txt)

### Expected Behavior

Gamma settings should be applied appropriately.

### Steps to Reproduce & Installation Method

Turn on night mode
Sleep machine
Wake machine
Note wrong temperature

bbedward:
Try 5e111d89a5e8a3aa473bedde642af63a756a96f2

philjackson:
@bbedward Works here!

PR #3388, opened 2026-09-10
feat(display): per-output ICC color profiles and color temperature

KIDult2226:
## Description

Adds per-output ICC color profile support and per-output color temperature to the night light / gamma stack.

**ICC profiles**
- New `core/internal/icc` package: ICC v2/v4 parser (TRC curves, LUT8/LUT16 and parametric curves, `vcgt`, description/version/color space metadata) with unit tests.
- The wayland manager loads `iccProfiles` and `outputTemps` from the DMS config directory of the running compositor (`<config>/<compositor>/dms/wayland.json`, for example `~/.config/niri/dms/wayland.json` on niri), builds gamma ramps from the parsed profile per output, falls back to temperature ramps when a profile is missing or unparsable, and re-applies after resume.
- ICC ramps are applied even when the global night light is off (outputs without a profile keep identity/neutral), so a calibrated profile is not silently dropped when the schedule is disabled.
- The configuration is attached independently of the gamma control lifecycle, and a hotplugged output picks up its profile and temperature as soon as its name is known (or when its control is created).

**Per-output temperature**
- `outputTemps` in the same config; `SetOutputTemp` accepts 1000-10000K and the Display Config slider exposes the full 3000-10000K range.
- Per-output temperature is an independent target: it applies whether or not the night light schedule runs, and a per-output value wins over the schedule, so each display can keep its own white point and its own on/off state (`0` = no override). It is composed on top of an ICC profile ramp, which is treated as measured at 6500K (D65); outputs without a profile are driven with the same value directly. `dms icc set-temp <output> <kelvin>` sets it from the CLI and `dms icc status` reports it.

**IPC / CLI**
- `wayland.icc.getStatus`, `wayland.icc.apply`, `wayland.icc.remove`, `wayland.icc.listOutputs`, `wayland.icc.setTemp`, `wayland.icc.getTemps`.
- `dms icc list | info <file> | apply <output> <file> | remove <output> | status | set-temp <output> <kelvin>`.

**UI**
- Display Config output card gains a **Color Profile** row (browse/apply/remove with description, version, color space and active state) plus a **per-output color temperature** slider, backed by `quickshell/Services/ICCService.qml`.

**Included fix**
- `LoadConfig()` now takes its fallback values from `DefaultConfig()` and covers `Contrast`. `Validate()` rejects `Contrast == 0`, so night light configs written before that field existed aborted manager construction and took gamma/ICC down entirely (`wayland manager not initialized`).

## Type of change

- [ ] Documentation
- [ ] Other

## Screenshots / video

Display Config output cards with the new **Color Profile** row (browse / apply / remove, showing an applied profile with version, color space and active state) and the per-output **Color Temp** slider set to 7000K on the calibrated display:

Display Config: Color Profile row and per-output color temperature

## Checklist

- [x] Go changes: ran `make fmt`, added/updated tests, `make test` passes (`TZ=UTC go test ./...` -> 51 ok, 0 fail), and `go mod tidy` is clean

> `make lint-qml` could not run in my worktree: it needs the Quickshell tooling VFS and the `dank-qml-common` submodule, which is not checked out there; it fails identically on unmodified master and lints only 3 entrypoints. i18n checks (`check_term_variants.py`, `check_term_freeze.py`) and `extract_settings_index.py` were run instead.
- [x] I have opened a corresponding pull request in dlx-docs to document any new behaviors (AvengeMedia/DankLinux-Docs#145)


KIDult2226:
Rebased onto current master (`6b50696b`) and rebuilt so the diff contains only the ICC / per-output-temperature work. Thanks for the review — I reproduced every point before fixing it.

**Reverted upstream fixes (restored)**
- `manager.go:229` — the "output returned, re-establishing controls" branch is back: `addOutputControl` is always posted and `controlsInitialized` is re-marked so the following `gamma_size` drives the reapply.
- `manager.go:341` — `out.lastTemp = 0` restored in the `gamma_size` handler and on the send-failure path.
- `manager.go:542` — restored the midnight-crossing `sunset` adjustment, `activeCycle`/`shiftTimes` (plus their call sites in `getSunPositionNormal`, `getDeadlineNormal`, `updateStateFromSchedule`) and the `m.schedule = sunSchedule{}` reset.
- `manager.go:815` — `applyGamma(neutralTemp)` (fixed 6500) again, not `HighTemp`.
- `DisplayConfigTab.qml:672` — `identifyConfigured`, the `MonitorIdentifyOverlay` loader and `id: monitorCanvas` are back; `Theme.spacingXXS` untouched in both files.

**New-code fixes**
- `manager.go:851` — the manager-level dedupe is gone; `applyGamma` uses the per-output `rampCurrent` check as before, so contrast-only `SetAdjustments` writes and the post-resume forced resend (#1235) work again. ICC/`setTemp` changes now clear the affected output's state (`out.lastTemp = 0`), which the schedule-derived comparison cannot see.
- `needsControls()` now returns true for configured `ICCProfiles`/`OutputTemps`, so disabling the night light no longer calls `destroyControls()` and drops the ICC ramps. Covered by `TestManager_NeedsControlsCoversICCAndOutputTemps` (fails on the old one-liner).
- `types.go:167` / `commands_icc.go:70` — both go through `DMSConfigDir()` / `ICCProfilesDir()`, which map the running compositor to `niri/dms`, `hypr/dms` or `mango/dms` using the same detection the rest of core uses, and resolve via `utils.XDGConfigHome()`. `dms icc` reuses that helper instead of its own copy.
- `parser.go:301` — `desc`/`mluc` offsets and lengths are validated in `int` arithmetic before slicing (`strStart <= strEnd <= len(data)`, `recordStart + 12 <= len(data)`); malformed profiles now return an error. `TestParseDescriptionMalformed` panics on the old code (`slice bounds out of range [12:4]`) and passes on the new one.
- `OutputCard.qml:571` — the `tempLabel.text` assignment is gone, so the existing binding keeps tracking the slider and `ICCService.outputTemps`.
- i18n: new strings carry translator context, existing terms (`Browse`, `Default`, `Active`, `Color Temperature`) are reused, and no catalog was re-extracted.

One small addition beyond the review: `SettingsContent.qml` now passes `parentModal` into `DisplayConfigTab`, so the ICC file browser stacks over the settings modal the way `WallpaperTab`'s does.

Verification: `TZ=UTC go test ./...` (all packages ok), `check_term_variants.py` clean, `extract_settings_index.py` produces no diff. `make lint-qml` fails identically on master in my worktree because the `dank-qml-common` submodule is not checked out; it lints 3 entrypoints only, so it does not cover these files either way.

Diff vs master: 13 files, +2406/−13. The 13 removed lines are `DefaultConfig()`→`LoadConfig()`, the single `GenerateGammaRamp(...)` line replaced by the ICC-aware ramp selection, the `needsControls()` one-liner, the `State` struct field realignment, and the `DisplayConfigTab` instantiation.

KIDult2226:
Post-review updates now on the branch (`033b96b3`), on top of the fixes in the previous comment:

- **`dms icc set-temp <output> <kelvin>`** (alias `setTemp`) sets a per-output value, and **`dms icc status`** reports it, so the override is reachable and checkable without the settings UI.
- **Per-output temperature is an independent target**: it applies whether or not the night light schedule runs, and a per-output value wins over the schedule, so every display keeps its own white point and its own on/off state (`0` = no override, and the range 3000-10000K is not limited by the night light's). It is composed on top of the profile ramp with the profile treated as measured at 6500K (D65). Decision logic lives in `effectiveTempTarget`.
- **Config attachment no longer depends on the control lifecycle**: restoring the "output returned, re-establishing controls" branch made the registry handler establish the controls before the startup post ran, and that post returned early on `controlsInitialized`, so the configured profiles/temperatures were never attached (every output reported `(none)` until re-applied by hand). The startup path now loads the configuration first and only creates the controls when they are missing.
- **Hotplugged outputs** pick up their configured profile and temperature as soon as the name is known, or when the control is created — the description claimed this but the code did not do it.

Tests added/updated: `TestParseDescriptionMalformed` (panics with `slice bounds out of range [12:4]` before the fix), `TestManager_NeedsControlsCoversICCAndOutputTemps`, `TestProfileRampWithTemp`, `TestEffectiveTempTarget`, `TestManager_LoadConfiguredICCWhenControlsAlreadyExist`, `TestManager_AttachConfiguredICCForNamedOutput`.

`TZ=UTC go test ./...` -> 51 packages ok, 0 fail. Verified on a real niri session with three calibrated outputs and per-output 7000K overrides: the ramps shift relative to 6500K as expected, profiles attach on restart, and no ICC/gamma warnings appear in the journal.

KIDult2226:
Third pass pushed (`d09daaf1`). Every point reproduced before fixing.

**ICC tag table (`parser.go:132`)**

The count is now compared against the bytes that are actually left, before it sizes anything. `tagCount*12` wrapped in `uint32`, so `0x15555556` passed the old check, `make([]tagEntry, 357913942)` allocated and the entry loop panicked. `ParseFile` runs on the wayland actor goroutine, which has no `recover()`, so a profile persisted in `wayland.json` took the daemon down at startup. `TestParseBytesTagTableBounds` dies with `runtime: out of memory: cannot allocate 4299161600-byte block` at `parser.go:138` on `4064dec` and passes here; a count that exactly fills the file is still accepted.

**Off-actor `outputState` reads (`manager.go:1037`)**

The manager publishes a snapshot (`m.iccStatus`/`m.iccTemps`) whenever the actor changes a profile or temperature, and `GetICCStatus`/`GetOutputTemps` serve that snapshot under its own mutex instead of ranging `m.outputs` and reading the actor's fields. `updateStateFromSchedule` reads the snapshot, so the per-profile `os.Stat` no longer runs on every 25 K step. A temporary copy of the new getter test against `4064dec` reports `WARNING: DATA RACE` on `manager.go:1547` / `manager.go:1640`; `go test -race` on `internal/icc` and `internal/server/wayland` is clean here.

**`iccProfiles` `omitempty` (`types.go:69`)**

`State.ICCProfiles`/`OutputTemps` no longer omit the key. `dms icc remove` (or removing the last profile from any other surface) now pushes `iccProfiles: {}`, the QML truthiness guard sees a truthy object and `ICCService.status` is cleared — the removed profile no longer stays in Display Config with the green "active" dot. The daemon publishes an empty snapshot at startup, so the key is `{}` rather than `null`.

**`m.outputNames` (`manager.go:1586`)**

The registry remove handler is now `removeOutputByRegistryName`, which deletes the `outputNames`/`outputRegNames` entries together with the output state and publishes the snapshot. A disconnected monitor stops being reported by `dms icc listOutputs`/`status`, and the narrower window is closed too: a rebound `wl_output` reusing the released object ID can no longer attach the previous monitor's profile from the stale name. Covered by `TestManager_RemoveOutputByRegistryNamePrunesNames`.

**Temp slider wheel (`OutputCard.qml:582`)**

`wheelEnabled: false`, matching `SettingsSliderRow:149`, `SettingsSliderCard:135` and `SystemMonitorVariantCard:379`. Confirmed against `dank-qml-common/DankCommon/Widgets/DankSlider.qml:210-222`: the wheel handler assigns `value` and emits `sliderValueChanged` without ever emitting `sliderDragFinished`, so the change was never sent and the binding was destroyed.

Verification: `TZ=UTC go test ./...` -> 51 ok, 0 fail; `go test -race ./internal/icc/ ./internal/server/wayland/` clean; `gofmt`/`go vet` clean; `qmllint` clean on the touched QML; `check_term_variants.py` clean and `extract_settings_index.py` produces no diff.

KIDult2226:
One more commit on the branch (`48866b3e`), beyond the review: `setupOutputControls` stored a fresh `outputState` for every output it touched, and `createControls` calls it whenever the controls are rebuilt, so recreating a control dropped that output's attached `iccPath`/`iccProfile`/`outputTemp`.

It is dormant today (`destroyControls` only runs when nothing is configured, and `needsControls()` now covers `ICCProfiles`/`OutputTemps`), but it would have made the snapshot published after `createControls` report the rebuilt state instead of the configuration.

`controlStateFor` now reuses the state of a known output and resets only what belongs to the new control — the ramp bookkeeping and the failure budget (`rampSize`, `failed`, `retryCount`, `lastFailTime`, `lastTemp`, `lastGamma`, `lastContrast`) — so a control that has not reported `gamma_size` still receives its ramp, while the profile and the per-output temperature survive. An output seen for the first time starts clean. Covered by `TestManager_ControlStateReuseKeepsAttachedICC`.

`TZ=UTC go test ./...` -> 51 ok, 0 fail; `go test -race` on `internal/icc` and `internal/server/wayland` clean.

KIDult2226:
Fourth pass pushed (`91f24693`). All five points reproduced before fixing, and every other occurrence of the same pattern is fixed too.

**`Config` maps aliased across value copies (`types.go:46`)**

Confirmed and fatal: on `48866b3e` a config copy taken under `RLock` and marshalled after the unlock while the IPC goroutine stores a profile dies with `fatal error: concurrent map iteration and map write`. `cloneConfig`/`configSnapshot` now clone `ICCProfiles`/`OutputTemps` while the lock is held, and every copy that outlives it goes through them: `updateStateFromSchedule`, `recalcSchedule`, `getLocation`, `tomorrowDawn`, `SetConfig` (owns the caller's maps), the `savedConfig` handed to `SaveConfig` in `ApplyICC`/`RemoveICC`/`SetOutputTemp`, and `loadConfiguredICC`/`attachConfiguredICC`, which used to index the maps after releasing the lock. The other reference field in `Config`, `Outputs`, is never written by the manager, so a copy of it is already a snapshot. Covered by `TestManager_ConfigSnapshotOwnsItsMaps` and `TestManager_ConfigSnapshotConcurrentMarshal` (marshalling the snapshot in a loop while another goroutine stores profiles; clean under `-race`).

**`SetOutputTemp` never created the controls (`manager.go:1691`)**

Right, and it made the headline control a no-op on a default install. It now persists first, then applies on the actor through `ensureOutputControls` (the block `ApplyICC` had inline, shared by both), and the state push happens in that same actor body after the mutation instead of on the caller goroutine — `ApplyICC` and `RemoveICC` were restructured the same way, so a push can no longer report a pre-mutation snapshot. The published temperature map reports the configured value, falling back to the attached one, so an output whose name is not known yet still shows the value that was set. On `48866b3e` the same scenario reports `0` for the value it just stored; with this commit it reports `7000` and the file on disk agrees. Covered by `TestManager_SetOutputTempWithoutControlsPublishesValue`.

**TRC used as the video card ramp (`parser.go:568`)**

Agreed — only `vcgt` is a video card gamma table. The pre-fix ramp for a gamma-2.2 profile maps mid gray to `47909/65535`, which is the washout you described. `ApplyICC` now refuses a profile without `vcgt`, the attach path skips one with a warning (so `<config>/<compositor>/dms/wayland.json` cannot smuggle one in either), and `GenerateGammaRamp` errors instead of falling back to the TRC tags or to an identity ramp. TRC stays profile metadata: `dms icc info` and the settings dialog still report the tone-curve kind. `TestGenerateGammaRampRequiresVCGT` covers the new contract, `TestParseBytes/IdentityRamp` (which pinned the identity fallback) is gone, and `TestProfileRampWithTemp`'s fixture now carries an identity vcgt table instead of an empty profile. The dlx-docs page documents the rule (KIDult2226/DankLinux-Docs#145, `1434f88`).

**`icc: applied ICC ramp` at Info (`manager.go:960`)**

`log.Debugf` now; a per-25 K-step message has no business in the journal.

**Missing translator context (`ICCProfileInfoModal.qml:201`)**

`I18n.tr("No information available", "No information available")`, and I checked every line this PR adds to QML: that was the only string without context.

Verification: `TZ=UTC go test ./...` -> 51 ok, 0 fail; `go test -race ./internal/icc/ ./internal/server/wayland/ ./cmd/dms/` clean; `gofmt`/`go vet` clean; `qmllint` clean on the touched QML; `check_term_variants.py` clean and `extract_settings_index.py` produces no diff.

KIDult2226:
One more commit (`54d99dd0`), found while sweeping for the same class as the config aliasing: `stateChanged` compared only the schedule fields, so a change that touches nothing else — applying or removing a profile, `dms icc set-temp` — was "no change" for the notifier and never pushed. The settings page only caught up because the UI path calls `requestStatus()` afterwards, and the CLI path did not catch up at all, which is the symptom the `omitempty` fix was supposed to remove.

`iccStateChanged` now compares the profile status and the per-output temperatures, so those changes reach the subscribers on their own. The new `TestStateChanged` cases (profile applied, profile removed, profile description changed, output temp changed, unchanged ICC state) report `stateChanged() = false` on `91f24693`.

`TZ=UTC go test ./...` -> 51 ok, 0 fail; `go test -race` on `internal/icc` and `internal/server/wayland` clean; `go vet`/`gofmt` clean.

KIDult2226:
Fifth pass pushed (`9da245dc`). All three points reproduced before fixing.

**Night light state persisted by the ICC save (`server.go:147`)**

Confirmed with the exact repro: night light on → apply a profile → the daemon writes `"Enabled": true` → turning the night light off only changes the daemon's memory (the shell sends `setEnabled` on toggle, and there is no write) → log out/in and `LoadConfig()` reports `Enabled: true`, so the daemon drives the schedule while the UI shows night mode off.

`SaveICCConfig` replaces `SaveConfig` on the three ICC paths: it reads the file, replaces `iccProfiles`/`outputTemps` and writes it back, so every field the shell owns stays as it is on disk. A config file that does not exist yet is created from the defaults, which is what the daemon would have booted with anyway. Covered by `TestSaveICCConfigKeepsNightLightFields` (the previous commit flips `Enabled` back to `true` on the next load).

**Names not pruned without a gamma control (`manager.go:411`)**

Right, the early return was on the state, and the name maps are populated for every `wl_output` regardless of `needsControls()`. The object ID is resolved from `m.outputRegNames` now, the `outputNames`/`outputRegNames` entries are dropped before the state is looked up, and only the control teardown is skipped when there is no state — a state that is known without a registry name still falls back to matching its own. On `54d99dd0` the same call leaves `ListOutputs() = [HDMI-A-1]`; `TestManager_RemoveOutputByRegistryNameWithoutControl` and the existing `...PrunesNames` both pass here.

**`7000` sentinel (`OutputCard.qml:519`)**

The fallback is `0` now, and the two highlight tests (`DankIcon` colour, `tempLabel.color`) key off `!== 0`, which is what the `Default` label branches were written against. `publishICCState` keeps only `temp != 0` entries, so an unconfigured output now reads "Default" instead of claiming 7000K, and a deliberate 7000K is highlighted. The slider's `value` fallback already keyed off `=== 0` and stays as it is.

Verification: `TZ=UTC go test ./...` -> 51 ok, 0 fail; `go test -race ./internal/icc/ ./internal/server/wayland/` clean; `gofmt`/`go vet` clean; `/usr/lib/qt6/bin/qmllint --bare` reports no syntax error on the touched QML (note: `/usr/bin/qmllint` on this machine is a stub that exits 255 silently — the earlier "qmllint clean" claims of mine were checked against it, so I re-ran them with the real Qt binary; the result is the same, no syntax errors). The docs page gained a note that DMS only writes the two ICC keys (KIDult2226/DankLinux-Docs#145, `2618d1e`).

KIDult2226:
Proactive pass (`cfe93cc0`) over the two surfaces the review has not looked at yet — the `dms icc` CLI and the ICC IPC parameters — so the failures turn up here rather than in the next round.

**`dms icc apply` wrote to the profile directory before the daemon saw the file**

Only the *parse* was checked locally, so a profile the daemon refuses (no `vcgt` table, non-RGB color space) was copied into `<config>/dms/icc/` first and stayed there for `dms icc list` to keep offering. The command now applies the daemon's rules before copying, and drops the copy it just made when the request fails (a file that was already in the profile directory is left alone). Built both binaries and ran them against a temp config directory with the socket isolated:

| case | before (`9da245dc`) | now |
|---|---|---|
| profile without vcgt | copied `nogamma.icc`, went on to the IPC | rejected locally, `residual=[]` |
| valid profile, no daemon | `exit=1`, `residual=[vcgt.icc]` | `exit=1`, `residual=[]` |

`dms icc list` also marks a profile without a `vcgt` table (`(no vcgt table: cannot be applied)`), so the rule that makes `apply` refuse it is visible where the profiles are listed.

**IPC parameters reached the manager unchecked**

An empty (or whitespace-padded) output name was persisted as an entry in `wayland.json` that no `wl_output` can ever match, and a relative profile path was stored to be resolved again on the next start, from a different working directory. `ApplyICC`/`RemoveICC`/`SetOutputTemp` reject both now, before anything is written. Covered by `TestManager_RejectsInvalidICCArguments`.

Verification: `TZ=UTC go test ./...` -> 51 ok, 0 fail; `go test -race ./internal/icc/ ./internal/server/wayland/ ./cmd/dms/` clean; `gofmt`/`go vet` clean.

KIDult2226:
Both findings were already fixed on the branch in `d55d3d1f` while I was working on them, so instead of duplicating that commit I added what it was missing (`657ab160`):

**Regression tests for both fixes**

- `TestManager_RecreateOutputControlFollowsNeedsControls`: night light off with nothing configured → nothing to recreate; night light still off but the output has a profile (or a per-output temperature) → the control has to be recreated. Both configured cases return nil on `9226251c`, which is the `out.failed`-forever path.
- `TestManager_SetOutputTempZeroClearsTheOverride`: the reset button's whole path — `SetOutputTemp(output, 0)` drops the stored value and the published entry, so the row reads "Default" and the output follows the schedule again.

**One layout detail on the reset button**

The button is laid out unconditionally now, with `opacity`/`enabled` following the override instead of `visible`. The temperature row shares its column with the slider directly below it, so with conditional visibility the row grows from ~16 px to 30 px the moment a drag sets a value — and the slider moves 14 px down *while it is being dragged*. Keeping the slot also lets the label width drop the conditional. `qmllint --bare` is clean.

Verification: `TZ=UTC go test ./...` -> 51 ok, 0 fail; `go test -race ./internal/icc/ ./internal/server/wayland/ ./cmd/dms/` clean; `gofmt`/`go vet` clean; `check_term_variants.py` clean, `extract_settings_index.py` produces no diff.

PR #3391, opened 2026-09-10
launcher: add vpn connections as a search provider

### LINKED ISSUE #3352: Toggle VPN from Launcher/Spotlight
### Feature Description

I would like to be able to toggle VPN connection from the Launcher/Spotlight.

### Use Case

Keyboard driven toggle of VPN connection.

### Compositor(s)

All compositors

### If Other, please specify

_No response_

### Proposed Solution

No idea how the launcher/spotlight work but all configured VPN connections should be indexed and searchable.

The entry in the search results should probably look different if connected vs disconnected.

_No response_

avesst:
## Description

Adds a `dms_vpn` built-in launcher plugin that indexes every configured VPN profile and toggles it on enter, so a VPN can be brought up or down without leaving the keyboard.

Rows reflect connection state: connected profiles get a lock icon and a **Connected** badge, in-progress ones show **Connecting...**, the rest get a crossed-out key. The subtitle is the connection type (WireGuard, OpenVPN, and so on), which is also matched when searching.

Profiles come from `DMSNetworkService.profiles`, which is populated from the network state the shell already fetches at startup. No new timer, process, subscription or request is introduced, and nothing runs while the launcher is closed. The provider returns nothing when the backend is not NetworkManager.

Toggling reuses the existing `toggleVpn`, the same call the Control Center VPN list uses, so credential prompts and error toasts behave identically.

Two notes on the design:

- **It ships without a trigger.** A matched trigger prefix replaces the entire result set, so claiming `vpn` would have hidden every application whenever someone typed it to find their VPN client, and would have produced an empty launcher on iwd or wpa_supplicant. Matching on a `vpn` keyword instead keeps applications and profiles in one list. A trigger can still be set per user in Settings.
- **Connected state is read from the same predicate the toggle branches on.** Core reports `activating` and `deactivating` in the active connection list, so testing only for `activated` would have shown a connection still coming up as disconnected while enter tore it down.

## Type of change

- [ ] Documentation
- [ ] Other

## Related issues

Closes #3352

## Screenshots / video

Disconnected state:

Connected state:

## Checklist

- [x] ~~Go changes: ran `make fmt`, added/updated tests, `make test` passes, and `go mod tidy` is clean~~ **N/A** — no Go in this PR, QML only
- [x] ~~I have opened a corresponding pull request in dlx-docs to document any new behaviors: https://github.com/AvengeMedia/DankLinux-Docs~~ **N/A**

### DIFF
--- a/quickshell/Modules/Settings/LauncherTab.qml
+++ b/quickshell/Modules/Settings/LauncherTab.qml
@@ -815,7 +815,7 @@ Item {
                     spacing: Theme.spacingS
 
                     Repeater {
-                        model: ["dms_settings", "dms_notepad", "dms_sysmon", "dms_settings_search", "dms_clipboard_search", "dms_power", "dms_colorpicker", "dms_qr_generator"]
+                        model: ["dms_settings", "dms_notepad", "dms_sysmon", "dms_settings_search", "dms_clipboard_search", "dms_power", "dms_vpn", "dms_colorpicker", "dms_qr_generator"]
 
                         delegate: Rectangle {
                             id: pluginDelegate
--- a/quickshell/Services/AppSearchService.qml
+++ b/quickshell/Services/AppSearchService.qml
@@ -222,6 +222,17 @@ Singleton {
                 viewModeEnforced: true,
                 defaultSectionPriority: 2.3
             },
+            "dms_vpn": {
+                id: "dms_vpn",
+                name: I18n.tr("VPN"),
+                cornerIcon: "vpn_key",
+                comment: "DMS",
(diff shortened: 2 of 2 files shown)


avesst:
Accepted Claude's proposed fix.

avesst:
Pushed two follow-ups:

- **e2fd9d02** Browsing the VPN plugin showed an empty list after the previous fix. It now lists all profiles again, and the home view still hides them.
- **c51cd45c** Searching "connect" also matched connected profiles, since "disconnect" contains "connect". Dropped those keywords.

PR #3407, opened 2026-09-11
fix(audio): apply sink volume via pactl for BlueZ sinks

li-yuanfei:
## Problem

On some Bluetooth (BlueZ/PipeWire) sinks, writing the QML volume property does not
reach the device — the UI slider moves, the OSD changes, but the actual sink volume
stays where it was. Those sinks only respond to `pactl set-sink-volume`.

## Fix

Add `AudioService.setSinkVolume()`, which issues
`pactl set-sink-volume <sink> <n>%` and reports whether it handled the change.
`setVolume()` and `adjustDefaultSinkVolume()` try it first and fall back to the
existing property setter when no sink name is available.

The DankDash media dropdown and media player tab route their volume writes through
`AudioService.setVolume()` so every volume surface shares the same code path.

This follows existing precedent — `AudioService` already shells out to `pactl`
for sink enumeration and port switching (`pactl list sinks`, `pactl set-sink-port`).

## Units

`setSinkVolume()` takes a **percentage** (0..`sinkMaxVolume`), matching the `pactl`
argument. The QML `audio.volume` property keeps its **0..1 fraction** semantics, so
the fallback path still writes `volume / 100`.

## Testing

- `make lint-qml` — PASS (3 entrypoints)
- `make test-qml` fails on this machine, but **also fails on unmodified upstream
  `master`** in the same environment, so it is not a regression from this change.

Verified manually on a BlueZ sink that previously ignored UI volume changes:
volume now tracks the slider.

### DIFF
--- a/quickshell/Modules/DankDash/MediaDropdownOverlay.qml
+++ b/quickshell/Modules/DankDash/MediaDropdownOverlay.qml
@@ -69,7 +69,7 @@ Item {
             return;
         }
         if (AudioService.sink?.audio)
-            AudioService.sink.audio.volume = volume;
+            AudioService.setVolume(Math.round(volume * 100));
     }
 
     function selectPlayer(player) {
--- a/quickshell/Modules/DankDash/MediaPlayerTab.qml
+++ b/quickshell/Modules/DankDash/MediaPlayerTab.qml
@@ -189,7 +189,7 @@ Item {
             return;
         }
         if (AudioService.sink?.audio)
-            AudioService.sink.audio.volume = clamped;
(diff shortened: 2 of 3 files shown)


bbedward:
Pretty sure related to https://github.com/quickshell-mirror/quickshell/pull/808 

https://github.com/quickshell-mirror/quickshell/issues/807

Using pactl as a workaround comes with other problems so  I'd rather not do that

bbedward:
Closing (because quickshell fix is preferred), as a workaround you can change your keybinds to use pactl/wpctl instead of dms volume commands though

PR #3422, opened 2026-09-12
settings: add safe Startup Apps manager

alamin147:
## Summary

Add a Startup Apps page that manages normal per-user login applications without exposing DMS or desktop/session infrastructure as a generic service manager.

- list and toggle XDG autostart entries from user and system configuration directories
- disable system XDG entries through user overrides without modifying `/etc`
- list only conservatively classified user-facing systemd application services
- change systemd enablement without `--now`, so running applications are not started or stopped
- add and remove DMS-created user XDG startup entries
- refresh while the page is open and integrate the page with Settings search

## Safety

Systemd units are excluded by default. A service must be a persistent, loaded `.service`, use an allowed login target, avoid protected identities and critical session dependencies, and match a visible desktop application. Socket, timer, target, template, transient, generated, masked, static, DMS, compositor, audio, bus, portal, keyring, and session infrastructure is rejected.

The backend repeats classification before every systemd mutation. Protected entries therefore cannot be changed by calling the CLI directly. User-authored XDG files can be toggled but only entries created by DMS can be deleted.

## Validation

- `go test ./internal/startup ./cmd/dms`
- `go test -race ./internal/startup`
- `go vet ./internal/startup ./cmd/dms`
- `golangci-lint v2.13.1 run ./...` — 0 issues
- `make build`
- settings search-index generation and validation
- QML parse/syntax checks with Qt 6 `qmlformat` and `qmllint`
- live DMS smoke test: page load, search data, application discovery, and protected direct mutation

The complete `go test ./...` run passes Startup Apps and the rest of the reported packages, but still hits the existing unrelated `internal/screenshot` `TestWaylandOutputBoundsScaled` failure. `make test-qml` could not run locally because `qmltestrunner` is not installed.

### DIFF
--- a/core/cmd/dms/commands_common.go
+++ b/core/cmd/dms/commands_common.go
@@ -793,6 +793,7 @@ func getCommonCommands() []*cobra.Command {
 		blurCmd,
 		trashCmd,
 		systemCmd,
+		startupCmd,
 		switchUserCmd,
 	}...)
 }
--- a/core/cmd/dms/commands_startup.go
+++ b/core/cmd/dms/commands_startup.go
@@ -1,0 +1,94 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/AvengeMedia/DankMaterialShell/core/internal/startup"
+	"github.com/spf13/cobra"
+)
+
+var startupCmd = &cobra.Command{
+	Use:   "startup",
+	Short: "Manage user-facing startup applications",
+	Long:  "List and manage XDG autostart applications and safely classified user systemd application services.",
+}
+
+var startupListJSON bool
+
+var startupListCmd = &cobra.Command{
+	Use:   "list",
+	Short: "List manageable startup applications",
+	Args:  cobra.NoArgs,
+	RunE: func(cmd *cobra.Command, _ []string) error {
+		entries, err := startup.NewManager().List(cmd.Context())
+		if err != nil {
+			return err
+		}
+		if startupListJSON {
+			return json.NewEncoder(os.Stdout).Encode(entries)
+		}
+		for _, entry := range entries {
+			state := "disabled"
+			if entry.Enabled {
+				state = "enabled"
+			}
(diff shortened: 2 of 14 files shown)


bbedward:
Whats the overall rationale of this? Or use case. From what I can tell the main addition is listing system XDG entires (good), and systemd units. Systemd user unit manager is kinda a separate thing and most of it I guess.

One problem claude doesnt catch is that they cant delete files not marked X-DMS-Managed, when technically they should be able to since theyre in the home diectory.

Not sure I like tying it so tightly to CLI calls. We could probably do better with a socket service, fsnotify, and signals. Gated on subscribers and not having a CLI at all (not sure theres much benefit to the CLI or not tbh, maybe for scriptable purposes)

alamin147:
> Whats the overall rationale of this? Or use case. From what I can tell the main addition is listing system XDG entires (good), and systemd units. Systemd user unit manager is kinda a separate thing and most of it I guess.
> 
> One problem claude doesnt catch is that they cant delete files not marked X-DMS-Managed, when technically they should be able to since theyre in the home diectory.
> 
> Not sure I like tying it so tightly to CLI calls. We could probably do better with a socket service, fsnotify, and signals. Gated on subscribers and not having a CLI at all (not sure theres much benefit to the CLI or not tbh, maybe for scriptable purposes)

The main goal was to provide a Startup Apps page for the normal user-facing ways applications start at login, so users don't need to manually edit `~/.config/autostart` or know `systemctl --user`. I included systemd only for units that could be conservatively matched back to desktop applications, rather than trying to make this a general systemd unit manager.

You're right about removal — user-only entries under `~/.config/autostart` should be removable regardless of `X-DMS-Managed`. I'll fix that.

For the CLI, it wasn't intended to be a requirement of the feature; I mainly used it as the QML/backend bridge and it happened to make the functionality scriptable as well. A subscriber-gated service/socket with fsnotify/signals makes sense if that's the preferred DMS architecture.

I'm also fine with narrowing this PR to XDG autostart only and leaving systemd user-unit management separate if that fits the scope better. Would you prefer that direction for this PR?

PR #3425, opened 2026-09-13
settings: add support for multiple config files

### LINKED ISSUE #1732: Support config includes
### Feature Description

Separate device-specific config settings from static ones.

See possible proposed solutions below.

### Use Case

I've multiple systems that use DMS and share common dotfiles via Git.
As I don't want to click through same settings/theme on all my machines, I also store DMS `settings.json` in a Git repo.

The problem is that some machine-specific configs, such as, selected backlight device or at what displays to show the bar, are also saved in the same file.

This makes sharing a mutual DMS config problematic as I have to manually "git stash" the device-specific settings, add the desired shared ones before committing changes.

### Compositor(s)

- [x] All compositors
- [ ] Niri
- [ ] Hyprland
- [ ] MangoWC (dwl)
- [ ] Sway
- [ ] Other (specify below)

### Proposed Solution

One of solutions to the problem can be config include/extend support.

Any parameters declared in the `settings.json` file will override the extended config(s) options. 

#### `"include"` in JSON

Add `"include"` parameter support to include one or more configs.

```json
{
  "include": [
    "file1.json",
    "file2.json"
  ],
  ...
}
```

**Pros:**

File load order is explicitly set by user.

**Cons:**

Unfortunately this will require a one-time manual intervention from a user to add an include to the config file.

#### Autoload directory

Before loading the main `settings.json` config, DMS will autoload all config files from the `~/.config/DankMaterialShell/config.d` directory (if exists).

**Pros:**

Unlike previous option, this doesn't require any user intervention, except dropping shared config files into a designated `config.d` directory.

**Cons:**

When file load order matters (probably very rare case) - user has to explicitly name files in an ordered way.

Many WMs or any other configurable projects support includes in config files.

Directory autoloads (`config.d`) approach is used in udev, NetworkManager, OpenSSH, NVIDIA driver, and another.

ignacypekala:
## Description 

This PR adds support for multiple settings files by sourcing all json files from `config.d/` on top of `settings.json`. As outlined in #1732 this change makes it easier to manage DMS dotfiles across multiple machines.

### New Behaviour

- During startup, if `config.d/` directory exists in the configuration directory, all json files located in it are automatically sourced alongside settings.json. If it doesn't only settings.json is loaded. This check is not repeated so a restart is currently required after creating the directory.

- All the settings files are loaded asynchronously populating their own setting stores. Once all of the files are loaded a startup sequence runs (formerly `Component.onCompleted` signal handler), which, among other things, merges settings from all the files and applies them.

- When settings are being saved the object returned by `Store.toJson` is split between files based on where the settings were loaded from (defaults to `settings.json`).

#### File order

The default `settings.json` file is sourced first, followed by all the files from `config.d` in lexicographic order. A file loaded later can overwrite settings defined by it's precedents. When settings are changed from the UI, only its effective (last sourced) location is changed.

I tried my best to keep compatibility and the behavior should, for the most part, be the same as previously for anyone who doesn't have a `config.d` directory.

### Use of AI 

Everything here has been coded manually and thought out by me. I've occasionally used AI to inquire about specifics of QML which I'm new to and Javascript which I haven't used in ages. Finally, I used an AI to check my code for bugs.

## Type of change

- [ ] Documentation
- [ ] Other

## Related issues

Closes #1732

## Screenshots / video

Not visual

## Checklist

- [x] New user-facing strings are wrapped in `I18n.tr()` with translator
  context, reusing existing terms where possible
- [ ] Go changes: ran `make fmt`, added/updated tests, `make test` passes, and
  `go mod tidy` is clean
- [ ] I have opened a corresponding pull request in dlx-docs to document any new
  behaviors: https://github.com/AvengeMedia/DankLinux-Docs

### Default Value Skip

Formerly Store.toJson would disregard any setting which is set to the default value. This would cause the settings to jump between files. For example:

If you were to set `showSeconds` to true in `config.d/module.json` and then from the UI flick it off and right back on the setting would be removed from `module.json` and re-appear in the default location - `settings.json`

Now `Store.toJson` only ignores default values if they haven't been featured in any of the loaded files.

### Startup sequence 

Previously, the startup sequence would run inside a `Component.onCompleted` signal handler. However with asynchronous file loading that handler fires off before all the `config.d/*.json` files are known. 

This has been replaced by a 3 stage system. 1. **Discovering**: config.d is inspected. If it exists all json files are passed over to the Instantiator. 2. **Loading**: During this stage all the SettingsFile(s) delegated by the Instantiator are created. 3. **Ready**: At the beginning of this stage the startup sequence (formerly `Component.onCompleted`) runs. Settings are ready for saving and reloading.

### FolderListModel Proxy and Debouncing 

FolderListModel will catch all sorts of changes and immediately update its list. The problem, however is that file editors (e.g (neo)vim) commonly create temporary files to which they write the new content and then replace the original file to maintain atomicity of writes. This causes multiple file updates and if `settingsFolderModel` were directly linked to the Instantiator this would remove and re-add the corresponding SettingsFile when a file view was just updated. This would completely bypass the file reload debouncing and immediately reload the file as if had just been created.

This is solved by introducing a proxy `settingsFilesListModel` ListModel which receives debounced updates from the FolderListModel.

### Refactor 

Because I had to pretty much rebuild the settings file management I took this opportunity to cleanup some of the settings management logic:

- centralized settings processing: Previously settings mutiple settings processing checks like migrations, weather location updates, or vpn connections were scattered between loadSettings() which ran during the initial load, and FileView.onLoaded handler which ran on every load. And some logic was even present in both which led to a duplicate check on the initial load.

  Now _loadSettings runs on every load and just adjusts its behaviour based on the _hasLoaded flag. This makes it easy to see what runs when and prevents the duplication. I migrated all these checks to run exactly as they'd formerly do, but I'm pretty sure some of them are not placed correctly (for example clock format check which only runs on the initial load or the vpnLastConnected which doesn't even exist in the SettingsSpec).

- _pendingMigration system removed: The previous system would store a snapshot of settings which would be forcefully saved once a file is known to be writable. This approach successfully ensured that migrations will be preserved, but posed a risk of disregarding user choice (the snapshot could get out of date) and was just quite tricky for me to translate to the multi file system.

  Now the _loadSettings function sequence forces a save if the loaded settings object was modified. Then any affected files have their unsavedChanges flagged and are prevented from reloading (and corrupting the in-memory migrated settings).

### Save Fail Recovery 

Because the system now prevents reloads of files with unsaved changes I thought it is important to re-check if a file can be written to. I've made it so that if any file fails to save it triggers a save fail recovery which re-attempts to save every file which hasn't been saved successfully. The retries are performed in exponential intervals starting at 2 seconds, capping out at 60 seconds. A total of 15 retries is performed.

### Writable check & Read-Only Modal 

There is a modal in the settings app, which triggers when there are unsaved changes and the settings file is read only. I've made it so that the isReadOnly flag is set only after a save fails due to missing permissions. This way it's only ever set if there are any unsaved changes. This, combined with the removal of the migration system, makes the Process for checking whether `settings.json` is writable obsolete.

#### Modal 
That Read-Only modal features a button for copying all the current settings. Dumping the entire configuration no longer really makes sense in a multi-file setup where settings can be split between multiple files. 

Ideally, it should allow copying changes per-file or showing which file they belong to. However, I decided not to change it for now due to the ongoing settings app re-design, and frankly, my own incompetence when it comes to UI work.

### DIFF
--- a/quickshell/Common/SettingsData.qml
+++ b/quickshell/Common/SettingsData.qml
@@ -89,7 +89,6 @@ Singleton {
     property bool _pluginParseError: false
     property bool _hasLoaded: false
     property bool _isReadOnly: false
-    property bool _hasUnsavedChanges: false
     property bool _selfWrite: false
     property var _loadedSettingsSnapshot: null
     property var pluginSettings: ({})
@@ -1788,7 +1787,6 @@ Singleton {
     function loadSettings() {
         _loading = true;
         _parseError = false;
-        _hasUnsavedChanges = false;
         _pendingMigration = null;
 
         try {
@@ -1908,12 +1906,10 @@ Singleton {
         const wasReadOnly = _isReadOnly;
         _isReadOnly = !writable;
         if (_isReadOnly) {
-            _hasUnsavedChanges = _checkForUnsavedChanges();
             if (!wasReadOnly)
                 log.info("settings.json is now read-only");
         } else {
             _loadedSettingsSnapshot = JSON.stringify(Store.toJson(root));
-            _hasUnsavedChanges = false;
             if (wasReadOnly)
                 log.info("settings.json is now writable");
             if (_pendingMigration)
@@ -3533,7 +3529,6 @@ Singleton {
 
     function _registerSaveFailure() {
         root._isReadOnly = Object.values(settingFiles).some(file => file.isReadOnly)
-        root._hasUnsavedChanges = root._checkForUnsavedChanges();
     }
 
     function _mitigateLoadFailure() {
--- a/quickshell/Common/SettingsData.qml
+++ b/quickshell/Common/SettingsData.qml
@@ -3681,6 +3681,10 @@ Singleton {
         repeat: false
         running: false
         onTriggered: {
+            if (!configDirExists.checked) {
+                restart();
+                return;
+            }
             if (!configDirExists.exists) {
                 return;
             }
@@ -3734,6 +3738,9 @@ Singleton {
         // qmllint disable signal-handler-parameters
         onExited: (code) => {
             exists = code === 0;
+            if (exists) {
+                settingsFilesModelSyncDebounce.restart();
+            }
             checked = true;
             _checkIfAllSettingsFilesFound()
         }
--- a/quickshell/Common/SettingsData.qml
+++ b/quickshell/Common/SettingsData.qml
@@ -3578,7 +3578,6 @@ Singleton {
                     const txt = settingsFileView.text();
                     if (!txt || !txt.trim()) {
                         hasParseFailed = true;
-                        settingsFile.parseError();
                         return;
                     }
                     settingsFile.settings = JSON.parse(txt);
--- a/quickshell/Common/SettingsData.qml
+++ b/quickshell/Common/SettingsData.qml
@@ -1792,7 +1792,7 @@ Singleton {
         _loading = true;
 
         try {
-            let obj = getSettingsObject();
+            let obj = _getSettingsObjectFromFiles();
             let loadedSettings = JSON.stringify(obj);
 
             if (isInitial) {
@@ -3637,7 +3637,7 @@ Singleton {
             _startAfterSettingsFilesFound();
         }
     }
-    function getSettingsObject() {
+    function _getSettingsObjectFromFiles() {
         const settingsObject = {};
         for (const path of _settingsFilesPaths) {
             const settingsFile = _settingsFiles.get(path);
--- a/quickshell/Common/SettingsData.qml
+++ b/quickshell/Common/SettingsData.qml
@@ -3462,6 +3462,7 @@ Singleton {
     function getSettingsObject() {
         const settingsObject = {};
         Object.assign(settingsObject, ...Object.values(loadedSettings));
+        return settingsObject;
     }
     function fileLoaded(fileName, obj) {
         loadedSettings[fileName] = obj;
@@ -3476,7 +3477,7 @@ Singleton {
         }
 
         if (allLoaded()) {
-            Store.parse(root, obj)
+            Store.parse(root, getSettingsObject())
             _loadedSettingsSnapshot = JSON.stringify(Store.toJson(root));
             _hasLoaded = true;
             applyStoredTheme();
--- a/quickshell/Common/SettingsData.qml
+++ b/quickshell/Common/SettingsData.qml
@@ -3476,13 +3476,11 @@ Singleton {
             SessionData.saveSettings();
         }
 
-        if (allLoaded()) {
-            Store.parse(root, getSettingsObject())
-            _loadedSettingsSnapshot = JSON.stringify(Store.toJson(root));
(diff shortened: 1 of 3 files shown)


ignacypekala:
I've observed that running the qml test suite sometimes fails because of `bar-widget-cursor.qml`. 

What's interesting is that it does not always happen, only occasionally when running all the tests with:
```
make test-qml
```

From my testing it doesn't happen when running just that test with:
```
python quickshell/tests/run-surface-fixture.py quickshell/tests/qml/bar-widget-cursor.qml
```

I don't understand what could be causing this. I'm suspecting it might have to do with the fact that settings are loaded and saved asynchronously, which is not the case on master. The problem with that theory, however, is that this test properly awaits for the settings to load with a timer bound to _hasLoaded and it seemingly doesn't rely on the saving whatsoever.

Here's the log from a failing run:
[fail-log.txt](https://github.com/user-attachments/files/32562870/fail-log.txt)
The error message is at line `41`.

I'm totally out of my depth here, as I only have a thin idea of what the test itself does.

ignacypekala:
> I've observed that running the qml test suite sometimes fails because of `bar-widget-cursor.qml`.
> 
> What's interesting is that it does not always happen, only occasionally when running all the tests with:
> 
> ```
> make test-qml
> ```
> 
> From my testing it doesn't happen when running just that test with:
> 
> ```
> python quickshell/tests/run-surface-fixture.py quickshell/tests/qml/bar-widget-cursor.qml
> ```
> 
> I don't understand what could be causing this. I'm suspecting it might have to do with the fact that settings are loaded and saved asynchronously, which is not the case on master. The problem with that theory, however, is that this test properly awaits for the settings to load with a timer bound to _hasLoaded and it seemingly doesn't rely on the saving whatsoever.
> 
> Here's the log from a failing run: [fail-log.txt](https://github.com/user-attachments/files/32562870/fail-log.txt) The error message is at line `41`.
> 
> I'm totally out of my depth here, as I only have a thin idea of what the test itself does.

I've employed AI to figure this out and it succeeded. The issue came down to two things, as anticipated both caused by the asynchronous settings loading. Here's my summary:

1. Before the settings are loaded SettingsData hold the default values from the Spec. With asynchronous file loading the test in question would read those default values and the `running: SettingsData.barConfigs.length === 1` guard of the timer would be even before the settings were applied. This is fixed with the `root.configured` flag set after applying the configuration.

2. The second problem was way more tricky it involves a race condition with the compositor. Because of the asynchronous loading the test timer was able run after the right-hand side of the bar had it's width calculated, but before the window itself was resized to encompass it. This caused the calculations to get skewed, the size of clock's mouse area coordinates to go negative and cover the widget. It is fixed by ensuring the window has the width over a 100 before considering checking for overlaps.

Although the solution was made by an AI I read all the code and understand the whole thing. 

I got the clanker to compose a report on this issue. If you want more details, take a look.
[pr_bar_widget_cursor_explanation.pdf](https://github.com/user-attachments/files/32572197/pr_bar_widget_cursor_explanation.pdf)

PR #3428, opened 2026-09-13
privacy: stop counting screen captures as camera use

avesst:
## Description

The privacy indicator shows the camera as active while OBS captures the screen, even when no camera is in use.

`cameraActive` counts any live `Stream/Input/Video` node as camera use. An app receiving a PipeWire screen cast creates exactly that node, so the screen capture was reported as the camera. These streams are tagged `media.role = "Screen"`, and are now skipped. Streams from a real camera carry `media.role = "Camera"` and are still detected.

## Type of change

## Checklist

- [x] ~~New user-facing strings are wrapped in `I18n.tr()` with translator context, reusing existing terms where possible~~ N/A
- [x] ~~Go changes: ran `make fmt`, added/updated tests, `make test` passes, and `go mod tidy` is clean~~ N/A
- [x] ~~I have opened a corresponding pull request in dlx-docs to document any new behaviors: https://github.com/AvengeMedia/DankLinux-Docs~~ N/A


quickshell/Services/PrivacyService.qml
```diff
@@ -45,7 +45,7 @@ Singleton {
 
         for (let i = 0; i < Pipewire.nodes.values.length; i++) {
             const node = Pipewire.nodes.values[i]
-            if (!node || !node.ready) {
+            if (!node || !node.ready || node.properties?.["media.role"] === "Screen") {
```
bbedward, line 48: I'd prefer a case insensitive match here, just for robustness i guess in case a client sets it with the incorrect vocabularly. But I'm not sure if that's even relevant in practice so good to merge

bbedward (approved):

Issue #1023, opened 2025-12-14
systemd believes that dms exited succesfully when quickshell crashes


real-or-random:
## Compositor

- [x] niri
- [ ] Hyprland
- [ ] dwl (MangoWC)
- [ ] sway
- [ ] Other (specify)

## Distribution

Arch

## dms version

dms v1.0.2

## Description

The exit code of dms is zero (success) if qs crashes. This itself is fine but systemd will believe that dms exited successfully. That's why we currently have `Restart=always` in the service file. Something like `Restart=on-failure` would be cleaner. 

## Steps to Reproduce

1. Enable systemd user service and start dms
2. Make qs crash (e.g., by killing it)
3. Observe exit code of dms (`systemctl --user status dms`)

## Error Messages/Logs

```
Dec 14 12:46:44 dms[2481]:   WARN quickshell.dbus.properties: Error updating property org.mpris.MediaPlayer2.firefox.instance_1_67/org/mpris/MediaPlayer2/org.mpris.MediaPlayer2.Player:Position
Dec 14 12:46:44 dms[2481]:   WARN quickshell.dbus.properties: QDBusError("org.freedesktop.DBus.Error.ServiceUnknown", "The name is not activatable")
Dec 14 12:47:54 dms[2481]:   WARN: pw_stream_connect failed "No such device"
Dec 14 12:47:54 dms[2400]:   INFO  go: [SecretAgent] DeleteSecrets called: path=/org/freedesktop/NetworkManager/Settings, SSID=
Dec 14 12:47:58 systemd-coredump[12358]: Process 2481 (qs) of user 1000 dumped core.
                                         
                                         Stack trace of thread 2481:
                                         #0  0x0000742d2d0d1be4 wl_proxy_get_listener (libwayland-client.so.0 + 0x4be4)
                                         #1  0x0000742d2d0678f2 n/a (libQt6WaylandClient.so.6 + 0xa38f2)
                                         #2  0x0000742d2cf59ac6 n/a (libffi.so.8 + 0x7ac6)
                                         #3  0x0000742d2cf5676b n/a (libffi.so.8 + 0x476b)
                                         #4  0x0000742d2cf5906e ffi_call (libffi.so.8 + 0x706e)
                                         #5  0x0000742d2d0d148d n/a (libwayland-client.so.0 + 0x448d)
                                         #6  0x0000742d2d0d22e9 n/a (libwayland-client.so.0 + 0x52e9)
[...]

PR #3514, opened 2026-09-20
systemd: use explicit shell readiness

korbash:
## Description

When Quickshell crashes and starts recovering, it can release the notification
D-Bus name. With `Type=dbus`, systemd then stops DMS, which can exit successfully
and leave the panel down without triggering `Restart=on-failure`.

Switch to `Type=notify`: DMS sends `READY=1` once the shell, backend connection,
and its notification server are ready. Delegate crash recovery to systemd
without tying the service's lifetime to the notification D-Bus name.

Implementation was AI-assisted.

## Type of change

- [x] Bug fix
- [ ] New feature
- [x] Breaking change: the notify unit requires a matching binary and QML
- [x] Documentation
- [ ] Other

## Related issues

Related to #1023.

## Screenshots / video

Not applicable: no visual changes.

## Checklist


bbedward:
if we switch to the notify pattern, I wonder if it still ensures we get the notification bus name? The main rationale was that mako's service also claims the name and in a startup mako can claim the name before DMS.

If so, then maybe we can also get a solution for the system tray which misses SNI registration if they appear before DMS gets the bus

korbash:
I added `BusName=org.freedesktop.Notifications` back. It still blocks mako’s systemd service with Type=notify I tested it.

docs/systemd.md
```diff
@@ -0,0 +1,47 @@
+# Starting DMS with systemd
```
bbedward: Delete this file, any docs go to the docs repo

korbash:
I’ve updated the PR to check the tray too. Before sending `READY=1`, DMS now checks that:

- `org.freedesktop.Notifications` is owned by its Quickshell process.
- The `org.kde.StatusNotifierWatcher` D-Bus name has an owner, and that watcher responds with `IsStatusNotifierHostRegistered=true`. The watcher can belong to another process.
- DMS’s tray host name is owned by the same Quickshell process, with the owner PID checked.

The tray is also initialized when its widget is hidden.

If mako or another process already owns the notification name, DMS reports the conflict and exits with code `1`, letting systemd handle recovery. If the name is free, DMS waits until its Quickshell process owns it and the remaining checks pass.

Apps that wait for DMS’s `READY=1` before starting should have their tray icons show up correctly. This should help with the startup race discussed in #1073, where waiting for the tray before sending `READY=1` was also suggested.

quickshell/shell.qml
```diff
@@ -19,6 +19,38 @@ ShellRoot {
...
+                Log.warn("Failed to report shell readiness:", response.error);
+                return;
+            }
+            readinessReported = response.result?.success === true;
+        }, 5000);
+    }
+
+    Timer {
+        interval: 1000
+        running: entrypoint.shellReady && !entrypoint.readinessReported
```
bbedward: We can probbly move this to DMSService directly, but besides that we should gate this timer on some max attempts so it never got stuck infinitely retrying in the event the DMS socket doesnt respond.

bbedward (changes requested):
My preferred shape for this is no QML at all, in `core/cmd/dms/shellapp.go` I think we can just send READY when the qs child process spawns

korbash:
Starting Quickshell doesn’t mean the notification server or tray is ready yet. Sending READY=1 immediately would leave the same startup race we’re trying to fix. I think we should wait until Quickshell has initialized the notification server and tray

korbash:
This is similar to what happened to me: an internal Quickshell error took down notifications. If we send `READY=1` right after starting the process, we could report DMS as ready before we know the notification server and tray actually work.

assets/systemd/dms.service
```diff
@@ -5,8 +5,10 @@ After=graphical-session.target
 Requisite=graphical-session.target
 
 [Service]
-Type=dbus
+Type=notify
+NotifyAccess=main
 BusName=org.freedesktop.Notifications
```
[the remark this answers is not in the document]
korbash: DMS now reports the conflict and exits with 78/CONFIG. I’ve added RestartPreventExitStatus=CONFIG, so systemd won’t restart it for this error, even if mako was started manually. This prevents the restart loop without relying on timeouts or start-rate limits.

korbash:
If mako or dunst already owns the D-Bus name, DMS can't start cleanly because we shouldn't kill external processes. Exiting with a clear error is the best option, I think. We can increase the retry count.

bbedward:
> If mako or dunst already owns the D-Bus name, DMS can't start cleanly because we shouldn't kill external processes. Exiting with a clear error is the best option, I think. We can increase the retry count.

It's really better if it isn't a hard failure though, I dont think we want to fail to start the shell just because something else owns the bus. Notifications is more like an add-on than a requirement for our case (unlike mako where its their entire purpose for existing)

